PR #198, opened 2015-04-01
Create helper macros for common functions - NOT READY

infosia:
*NOT READY FOR MERGE*

Define helper macros for common functions/properties which can be used for both header and implementation. It also contains argument check/conversion macro such as `ENSURE_STRING_AT_INDEX`.

- [TIMOB-18403](https://jira.appcelerator.org/browse/TIMOB-18403)
- [TIMOB-18404](https://jira.appcelerator.org/browse/TIMOB-18404)
- [TIMOB-18405](https://jira.appcelerator.org/browse/TIMOB-18405)

PR #199, opened 2015-04-02
Create helper functions/macros for common property additions to classes

infosia:
Define helper macros for common functions/properties which can be used for both header and implementation. It also contains argument check/conversion macro such as `ENSURE_STRING_AT_INDEX`.

- [TIMOB-18403](https://jira.appcelerator.org/browse/TIMOB-18403)
- [TIMOB-18404](https://jira.appcelerator.org/browse/TIMOB-18404)
- [TIMOB-18405](https://jira.appcelerator.org/browse/TIMOB-18405)

*Done*: classes directly under `Titanium` namespace.  *Not ready*: other deeper namespaces such as `Ti.UI.*` and `Ti.Filesystem.*`, but I would like this PR merge for now because this PR becomes already pretty big :sunglasses:

I'll push other APIs in subsequent PR.


Source/TitaniumKit/include/Titanium/Analytics.hpp
```diff
@@ -40,10 +40,10 @@ namespace Titanium
 		static void JSExportInitialize();
 		static JSObject GetStaticObject(const JSContext& js_context) TITANIUM_NOEXCEPT;
 
-		virtual JSValue js_lastEvent() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_featureEvent(const std::vector<JSValue>&, JSObject&) TITANIUM_NOEXCEPT final;
-		virtual JSValue js_navEvent(const std::vector<JSValue>&, JSObject&) TITANIUM_NOEXCEPT final;
-		virtual JSValue js_getLastEvent(const std::vector<JSValue>&, JSObject&) TITANIUM_NOEXCEPT final;
+		TITANIUM_PROPERTY_READONLY_DEF(lastEvent);
```
sgtcoolguy: This loses the "final" modifier here. There's a number of others in this PR (like the functions below in this file). I think we should retain the final modifier for the areas where it's located to designate that we have a cross-platform implementation and don't expect platform-specific overrides.

Source/TitaniumKit/include/Titanium/App.hpp
```diff
@@ -178,47 +179,47 @@ namespace Titanium
...
-		virtual JSValue js_keyboardVisible() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_name() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_proximityDetection() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_proximityState() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_publisher() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_sessionId() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_url() const TITANIUM_NOEXCEPT final;
-		virtual JSValue js_version() const TITANIUM_NOEXCEPT final;
+		TITANIUM_PROPERTY_READONLY_DEF(accessibilityEnabled);
+		TITANIUM_PROPERTY_READONLY_DEF(analytics);
```
sgtcoolguy: We don't need to handle it in this PR, but it'd be nice to have macros that generated the function definitions too for properties (i.e. READONLY would do TITANIUM_FUNCTION_DEF for getProperty, normal PROPERTY_DEF would do both set/get); Given the case change in the function name, I'm not sure if we can easily do this?

sgtcoolguy:
Looks good. Only real concern is with the loss of the final modifier on some methods - I think it's important to retain those.

@infosia, feel free to merge this PR after tat has been addressed.

As a note just for historical context, this PR came from #198

Source/TitaniumKit/include/Titanium/Analytics.hpp
infosia: @sgtcoolguy That's a good point. I should have addressed about this important changes here. Since I found that we don't need `virtual` here (because simply it is not required) , we don't need `final` here too. I also removed `TITANIUM_NOEXCEPT`, because we can't assume our API will not throw C++ exception because every API could throw runtime exception.

PR #204, opened 2015-04-06
[TIMOB-18684] View width/height is ignored when use with Ti.UI.SIZE

infosia:
Fix for [TIMOB-18684](https://jira.appcelerator.org/browse/TIMOB-18684)

To test this, make sure following code shows 75x75 image at (top=0,left=0), and also `NG` and `Corporate` app works as expected.

```javascript
var window = Ti.UI.createWindow({ backgroundColor: 'blue' });

var view = Ti.UI.createImageView({
    image: 'http://api.randomuser.me/portraits/women/0.jpg',
    width: 75,
    height: Ti.UI.SIZE,
    top: 0,
    left: 0
});

window.add(view);
window.open();
```


sgtcoolguy:
@infosia are there layout mocha tests we can stop skipping ow that are fixed by this?

PR #215, opened 2015-04-14
[TIMOB-18828] Fix Titanium.Proxy in parity report

garymathews:
- Fix _Titanium.Proxy_ in parity report

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-18828)


sgtcoolguy:
@garymathews Looks like I was half-right. Not all these proxy methods are implemented yet. Maybe we can just have the doc generator know that Ti.Proxy is actually Titanium::Module in the TitaniumKit codebase? We have fireEvent, addEventListener, removeEventListener, and applyProperties. We don't have apiName, lifecycleContainer, or bubbleParent done.

PR #216, opened 2015-04-14
Clean up code with macros

sgtcoolguy:
This is a follow on to the Sensor/Geolocation PR. It addresses the comments I had, plus starts to use more macros across the code to continue to shrink down the lines of boilerplate code in our proxies.


sgtcoolguy:
@infosia feel free to revert this last commit if it wasn't complete or breaks the build.

PR #219, opened 2015-04-15
Fix: Remove "virtual" from Database modules

infosia:
Fix for [TIMOB-18696](https://jira.appcelerator.org/browse/TIMOB-18696) and [TIMOB-18695](https://jira.appcelerator.org/browse/TIMOB-18695). I noticed that you don't have to make them `virtual` and just use HAL types as it is because Database module is cross-platform and thus nobody extends it. Because original purpose of "removing HAL types" is meant to separate TitaniumKit from platform-specific modules, I think this fix fits the requirement.


infosia:
Closing as taken over by #220

PR #220, opened 2015-04-15
Fix: Database module doesn't properly clean up ResultSets

infosia:
- [TIMOB-18562](https://jira.appcelerator.org/browse/TIMOB-18562) Database module doesn't properly clean up ResultSets
- [TIMOB-18696](https://jira.appcelerator.org/browse/TIMOB-18696) Ti.Database.ResultSet interface should not use HAL types
- [TIMOB-18695](https://jira.appcelerator.org/browse/TIMOB-18695) Ti.Database.DB interface should not use HAL types 

For "removing HAL types", I noticed that you don't have to make methods `virtual` and just use HAL types as it is because Database module is cross-platform and thus nobody extends it. Because original purpose of "removing HAL types" is meant to separate TitaniumKit from platform-specific modules, I think this fix fits the requirement.

### DIFF
--- a/Source/TitaniumKit/src/Database/ResultSet.cpp
+++ b/Source/TitaniumKit/src/Database/ResultSet.cpp
@@ -19,9 +19,7 @@ namespace Titanium
 
 		ResultSet::~ResultSet() {
 			TITANIUM_LOG_DEBUG("ResultSet:: dtor ", this);
-			if (statement__ != nullptr) {
-				close();
-			}
+			close();
 		}
 
 		void ResultSet::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) 
--- a/Source/TitaniumKit/src/Database/ResultSet.cpp
+++ b/Source/TitaniumKit/src/Database/ResultSet.cpp
@@ -17,7 +17,15 @@ namespace Titanium
 			TITANIUM_LOG_DEBUG("ResultSet:: ctor ", this);
 		}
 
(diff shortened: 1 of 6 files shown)


Source/TitaniumKit/src/Database/ResultSet.cpp
```diff
@@ -17,7 +17,15 @@ namespace Titanium
 			TITANIUM_LOG_DEBUG("ResultSet:: ctor ", this);
 		}
 
-		void ResultSet::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) {
+		ResultSet::~ResultSet() {
+			TITANIUM_LOG_DEBUG("ResultSet:: dtor ", this);
```
sgtcoolguy: So one of my concerns was about calling db.execute and not holding onto the returned resultset in JS code. Does this suffice for closing and cleaning those resultsets up? The JS code may hold no reference to the returned ResultSet explicitly, so do we know if it'll ever get GCed properly?

sgtcoolguy: Just for reference, I believe some of the other implementations (iOS maybe) explicitly hold a vector of the resultsets and clean them all up when the db is closed/GCed.

infosia: I see, that sounds reasonable.  db.execute should hold reference to the resultsets and it should get closed once db is closed/GCed.

infosia: @sgtcoolguy Just posted update for it. Will be ready once Jenkins build passed.

PR #221, opened 2015-04-16
DB should release ResultSet when it's closed

infosia:
Related to #220 

Database should receive callback from ResultSet when it's closed and then DB should release shared pointer to the ResultSet. Pretty similar logic that how Ti.Current (iOS) does.

### DIFF
--- a/Source/TitaniumKit/src/UI/TextArea.cpp
+++ b/Source/TitaniumKit/src/UI/TextArea.cpp
@@ -94,7 +94,7 @@ namespace Titanium
 			TITANIUM_ADD_PROPERTY(TextArea, keyboardType);
 			TITANIUM_ADD_PROPERTY(TextArea, returnKeyType);
 			TITANIUM_ADD_PROPERTY(TextArea, textAlign);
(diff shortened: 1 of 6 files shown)


Source/TitaniumKit/src/UI/TextArea.cpp
```diff
@@ -94,7 +94,7 @@ namespace Titanium
 			TITANIUM_ADD_PROPERTY(TextArea, keyboardType);
 			TITANIUM_ADD_PROPERTY(TextArea, returnKeyType);
 			TITANIUM_ADD_PROPERTY(TextArea, textAlign);
-			TITANIUM_ADD_PROPERTY(TextArea, autocapitalization);
+			TITANIUM_ADD_PROPERTY(TextArea, autoCapitalization);
```
sgtcoolguy: The docs state the property is all lowercase here, no mixed/camel case. This change shouldn't be done.

sgtcoolguy:
@infosia good to merge when the TextArea/autocapitalization property change is rolled back.

infosia:
Pushed a fix for TextArea.autocapitalization. Will be ready soon.

PR #222, opened 2015-04-17
[TIMOB-18050] Implement Ti.UI.ActivityIndicator

infosia:
Implement Ti.UI.ActivityIndicator

- [TIMOB-18050](https://jira.appcelerator.org/browse/TIMOB-18050)
- Update NG example to showcase existing components

timob-18050

### DIFF
--- a/Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
+++ b/Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
@@ -66,6 +66,20 @@ namespace Titanium
 			*/
 			TITANIUM_PROPERTY_IMPL_DEF(ACTIVITY_INDICATOR_STYLE, style);
 
+			/*!
+			@property
+			@abstract indicatorColor
+			@discussion Color of the animated indicator.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, indicatorColor);
+
+			/*!
+			@property
+			@abstract indicatorDiameter
+			@discussion Diameter of the indicator. The diameter of the animated indicator, exclusive of any message text.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, indicatorDiameter);
+
 			ActivityIndicator(const JSContext&) TITANIUM_NOEXCEPT;
 
 			virtual ~ActivityIndicator() = default;
@@ -83,6 +97,8 @@ namespace Titanium
 			TITANIUM_PROPERTY_DEF(message);
 			TITANIUM_PROPERTY_DEF(messageid);
 			TITANIUM_PROPERTY_DEF(style);
+			TITANIUM_PROPERTY_DEF(indicatorColor);
+			TITANIUM_PROPERTY_DEF(indicatorDiameter);
 
 			TITANIUM_FUNCTION_DEF(getColor);
 			TITANIUM_FUNCTION_DEF(setColor);
@@ -94,6 +110,10 @@ namespace Titanium
 			TITANIUM_FUNCTION_DEF(setMessageid);
 			TITANIUM_FUNCTION_DEF(getStyle);
 			TITANIUM_FUNCTION_DEF(setStyle);
+			TITANIUM_FUNCTION_DEF(getIndicatorColor);
+			TITANIUM_FUNCTION_DEF(setIndicatorColor);
+			TITANIUM_FUNCTION_DEF(getIndicatorDiameter);
+			TITANIUM_FUNCTION_DEF(setIndicatorDiameter);
 
 		protected:
 #pragma warning(push)
@@ -102,6 +122,8 @@ namespace Titanium
 			Font font__;
 			std::string message__;
 			std::string messageid__;
+			std::string indicatorColor__;
+			std::string indicatorDiameter__;
 			ACTIVITY_INDICATOR_STYLE style__;
 #pragma warning(pop)
 		};
--- a/Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
+++ b/Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
@@ -1,0 +1,111 @@
+/**
+ * TitaniumKit Titanium.UI.ActivityIndicator
+ *
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#ifndef _TITANIUM_ACTIVITYINDICATOR_HPP_
+#define _TITANIUM_ACTIVITYINDICATOR_HPP_
+
+#include "Titanium/Module.hpp"
+#include "Titanium/UI/View.hpp"
+#include "Titanium/UI/Font.hpp"
+#include "Titanium/UI/ActivityIndicatorStyle.hpp"
+
+namespace Titanium
+{
+	namespace UI
+	{
+
+		using namespace HAL;
+
+		/*!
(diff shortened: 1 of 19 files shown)


Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
```diff
@@ -0,0 +1,111 @@
...
+	{
+
+		using namespace HAL;
+
+		/*!
+		  @class
+		  @discussion This is the Titanium ActivityIndicator Module.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/Titanium.UI.ActivityIndicator
+		*/
+		class TITANIUMKIT_EXPORT ActivityIndicator : public View, public JSExport<ActivityIndicator>
```
sgtcoolguy: According to the docs, this doesn't extend from View: http://docs.appcelerator.com/titanium/3.0/#!/api/Titanium.UI.ActivityIndicator

Looks like it really only uses _some_ of the standard properties from View: top/left/bottom/right/height/width

Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
```diff
@@ -0,0 +1,111 @@
...
+			  @discussion Key identifying a string in the locale file to use for the message text.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, messageid);
+
+			/*!
+			  @property
+			  @abstract style
+			  @discussion The style for the activity indicator.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(ACTIVITY_INDICATOR_STYLE, style);
```
sgtcoolguy: I know that iOs/Android don't implement them, but we should probably add indicatorColor and indicatorDiameter to the TitaniumKit definition. I think TitaniumKit should have stubs/definitions for _every_ property regardless of how many or few platforms it exists on. The ones we should consider overriding in Windows would be the set where they actually exist on iOS and Android (as a first pass, eventually we'd implement whatever ones we could).

Source/TitaniumKit/include/Titanium/UI/ActivityIndicatorStyle.hpp
```diff
@@ -0,0 +1,77 @@
...
+			DARK,
+			PLAIN
+		};
+
+		/*!
+		  @class
+		  @discussion This is the Titanium ActivityIndicatorStyle Module.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/Titanium.UI.ActivityIndicatorStyle
+		*/
+		class TITANIUMKIT_EXPORT ActivityIndicatorStyle : public Module, public JSExport<ActivityIndicatorStyle>
```
sgtcoolguy: This feels weird to me that this was given a full proxy/type status and isn't just an enum. I know the docs state it as such, but really there's no events to fire or listen for. I'd argue that this was a bad design decision in the past for this and that it should really be an enum only.

Source/TitaniumKit/include/Titanium/UI/ActivityIndicator.hpp
sgtcoolguy: We may want to discuss in a wider audience these oddities of the API and whether we're ok with possibly straying. I think it makes sense to extend from View since it'd be useful to change the background color, or get the size/rect. But the existing API doesn't treat it as a View subclass. 
And per my note below, I think ActivityIndicatorStyle _doesn't_ need to be a full type and should just be an enum.
Maybe @ingo is the one to raise these up to?

Source/TitaniumKit/src/UI/ActivityIndicator.cpp
```diff
@@ -0,0 +1,130 @@
...
+		TITANIUM_FUNCTION_AS_SETTER(ActivityIndicator, setColor, color)
+
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, getFont, font)
+		TITANIUM_FUNCTION_AS_SETTER(ActivityIndicator, setFont, font)
+
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, getMessage, message)
+		TITANIUM_FUNCTION_AS_SETTER(ActivityIndicator, setMessage, message)
+
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, getMessageid, messageid)
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, setMessageid, messageid)
```
sgtcoolguy: Should be TITANIUM_FUNCTION_AS_SETTER

Source/TitaniumKit/src/UI/ActivityIndicator.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, getStyle, style)
+		TITANIUM_FUNCTION_AS_GETTER(ActivityIndicator, setStyle, style)
```
sgtcoolguy: Should be TITANIUM_FUNCTION_AS_SETTER

Source/TitaniumKit/src/UIModule.cpp
```diff
@@ -981,6 +982,12 @@ namespace Titanium
 		CREATE_TITANIUM_UI(TableViewRow);
 	}
 
+	TITANIUM_FUNCTION(UIModule, createActivityIndicator)
+	{
+		ENSURE_OPTIONAL_OBJECT_AT_INDEX(parameters, 0);
+		CREATE_TITANIUM_UI(ActivityIndicator);
+	}
+
 	// TODO empty implementation so that it won't break default app template. Need to implement later on.
```
sgtcoolguy: Unrelated to this PR, but we should open a ticket for this.

sgtcoolguy: https://jira.appcelerator.org/browse/TIMOB-18839

Source/UI/include/TitaniumWindows/UI/Label.hpp
```diff
@@ -47,6 +47,8 @@ namespace TitaniumWindows
 			virtual void set_font(const Titanium::UI::Font& font) TITANIUM_NOEXCEPT override final;
 			virtual void enableEvent(const std::string& event_name) TITANIUM_NOEXCEPT override final;
 
+			static const std::uint32_t DefaultFontSize = 20;
```
sgtcoolguy: I think there's a DEFINE around somewhere about default font size too...

Source/UI/src/ActivityIndicator.cpp
```diff
@@ -0,0 +1,94 @@
...
+			getViewLayoutDelegate<ActivityIndicatorLayoutDelegate>()->setComponent(panel__);
+		}
+
+		void ActivityIndicator::set_message(const std::string& message) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ActivityIndicator::set_message(message);
+			label__->Text = TitaniumWindows::Utility::ConvertUTF8String(message);
+		}
+
+		void ActivityIndicator::JSExportInitialize() {
```
sgtcoolguy: brace on next line

Source/UI/src/Label.cpp
```diff
@@ -10,8 +10,6 @@
 #include "TitaniumWindows/UI/WindowsViewLayoutDelegate.hpp"
 #include "TitaniumWindows/Utility.hpp"
 
-#define DEFAULT_FONT_SIZE 20
```
sgtcoolguy: Ah, looks like you found the define :+1: 

Ignore my previous comment.

infosia:
Ready for merge.

PR #223, opened 2015-04-21
[TIMOB-18839] Implement Ti.UI.setBackgroundColor

infosia:
- [TIMOB-18839](https://jira.appcelerator.org/browse/TIMOB-18839) Implement Ti.UI.setBackgroundColor
- Add Ti.UI methods
- Implement Ti.UI.`currentWindow` and`currentTab`
- Enable unit tests for Ti.UI.Window
  - some of them are "skipped", need to be fixed later on

### DIFF
--- a/Source/TitaniumKit/src/ApplicationBuilder.cpp
+++ b/Source/TitaniumKit/src/ApplicationBuilder.cpp
@@ -31,6 +31,7 @@ namespace Titanium
 		: js_context__(js_context),
 		  global_object__(js_context__.get_global_object()),
 		  ti__(js_context__.CreateObject(JSExport<Titanium::TiModule>::Class())),
+		  ui__(js_context__.CreateObject(JSExport<Titanium::UIModule>::Class())),
 		  api__(js_context__.CreateObject(JSExport<Titanium::API>::Class())),
 		  view__(js_context__.CreateObject(JSExport<Titanium::UI::View>::Class())),
 		  geolocation__(js_context__.CreateObject(JSExport<Titanium::GeolocationModule>::Class())),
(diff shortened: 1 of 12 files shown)


Source/TitaniumKit/src/ApplicationBuilder.cpp
```diff
@@ -140,7 +139,7 @@ namespace Titanium
 
 			  // Create the global alert function
 			  alert = function (_msg) {
-				  Ti.UI.createAlertDialog({
+				  Ti.ui__.createAlertDialog({
```
sgtcoolguy: I don't think this is right. Should still be Ti.UI.createAlertDialog

Source/TitaniumKit/src/UI/Window.cpp
```diff
@@ -38,12 +47,16 @@ namespace Titanium
 
 		void Window::close(const std::shared_ptr<CloseWindowParams>& params) TITANIUM_NOEXCEPT
 		{
-			TITANIUM_LOG_WARN("Window::close: Unimplemented");
+			GET_UI();
+			const auto ui_ptr = UI.GetPrivate<Titanium::UIModule>();
+			ui_ptr->set_currentWindow(nullptr);
```
sgtcoolguy: So I think this is technically incorrect and points towards an issue I've long been avoiding: We need something similar to https://github.com/appcelerator/titanium_mobile_windows.bak/blob/master/Source/TitaniumPedro/Modules/UI/TiPageManager.h

We don't keep track of the stack of windows/pages and I think because we always use a generic page type when we open a new one, I don't think we get the benefits of the framework keeping track of the stack of pages. So when we close a window, if it's the last one then the current window is null. But that should only really be when the app exits. It should make the current window the one that's next in the stack.

Source/TitaniumKit/src/UIModule.cpp
```diff
@@ -982,17 +996,70 @@ namespace Titanium
 		CREATE_TITANIUM_UI(TableViewRow);
 	}
 
+	TITANIUM_PROPERTY_GETTER(UIModule, backgroundImage)
```
sgtcoolguy: This looks ripe for another macro for such simple cases:
TITANIUM_PROPERTY_GETTER_TYPED(UIModule, String, backgroundImage);

I don't know how many cases we have of straight up pass through conversions around, but I'm guessing there's a good number for std::string->String on properties (and maybe bool->Boolean). Maybe less cases for uint32_t/int32_t/double->Number.

Source/TitaniumKit/src/UIModule.cpp
```diff
@@ -982,17 +996,70 @@ namespace Titanium
...
+	{
+		return get_context().CreateString(get_backgroundImage());
+	}
+
+	TITANIUM_PROPERTY_SETTER(UIModule, backgroundImage)
```
sgtcoolguy: Same thing here. Straight pass-throughs like this look like another case where we could just use a macro to do it all.

Source/TitaniumKit/src/UIModule.cpp
```diff
@@ -982,17 +996,70 @@ namespace Titanium
...
+	TITANIUM_PROPERTY_GETTER(UIModule, currentWindow)
+	{
+		const auto window = get_currentWindow();
+		if (window != nullptr) {
+			return window->get_object();
+		}
+		return get_context().CreateNull();
+	}
+
+	TITANIUM_FUNCTION_AS_GETTER(UIModule, getBackgroundImage, backgroundImage)
```
sgtcoolguy: It's killing me that C++ macros can't do something simple like allow us to uppercase the first letter of an argument so we can avoid the middle arguments to these macros. Oh well.

PR #227, opened 2015-04-24
Add navigation UI for Windows

infosia:
Part of [TIMOB-18734](https://jira.appcelerator.org/browse/TIMOB-18734), adding navigation interface for Windows. This also introduces some system icons.

- `Ti.UI.Windows.CommandBar`
- `Ti.UI.Windows.AppBarButton`
- `Ti.UI.Windows.AppBarToggleButton`
- `Ti.UI.Windows.AppBarSeparator`
- `Ti.UI.Windows.SystemIcon`

Note that as far as I can see `AppBarSeparator` shows no UI on Windows Phone 8.1 simulator.

```javascript
    backgroundColor: 'blue'
});

// Windows command bar
var commandBar = Ti.UI.Windows.createCommandBar();

// create buttons for the command bar
var backButton = Ti.UI.Windows.createAppBarButton({ icon: Ti.UI.Windows.SystemIcon.BACK, touchEnabled: false });
var nextButton = Ti.UI.Windows.createAppBarButton({ icon: Ti.UI.Windows.SystemIcon.FORWARD });
var separator  = Ti.UI.Windows.createAppBarSeparator();
var acceptButton = Ti.UI.Windows.createAppBarToggleButton({ icon: Ti.UI.Windows.SystemIcon.ACCEPT });

backButton.addEventListener('click', function () {
    win.close();
});

acceptButton.addEventListener('click', function () {
    backButton.touchEnabled = acceptButton.checked;
});

var newwin = Ti.UI.createWindow({
    backgroundColor: 'green'
});
var closeButton = Ti.UI.createButton({
    backgroundColor: 'black',
    title: 'Close this window'
});
closeButton.addEventListener('click', function () {
    newwin.close();
});
newwin.add(closeButton);

nextButton.addEventListener('click', function () {
    // command bar should be closed when newwin is opened, because command bar is not attached to newwin.
    newwin.open();
});

commandBar.items = [backButton, nextButton, separator, acceptButton];

win.add(commandBar);
win.open();
```

should be like this:

commandbar


Examples/NMocha/src/Assets/ti.ui.windows.commandbar.test.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+var should = require('./should');
+
+describe("Titanium.UI.Windows.CommandBar", function () {
+    function isWin() {
+       return (Ti.Platform.osname == 'windowsphone' || Ti.Platform.osname == 'windowsstore');
+    }
+
+    it("constructor", function (finish) {
+        if (!isWin()) return;
```
sgtcoolguy: actually needs to call finish() or the test will hang forever.

Source/UI/include/TitaniumWindows/UIModule.hpp
```diff
@@ -37,7 +39,8 @@ namespace TitaniumWindows
 #endif
 
 		static void JSExportInitialize();
-
+	private:
+		JSObject windowsStore__;
```
sgtcoolguy: Might want to rename to something like windowsModule__ or windowsNamespace__?

Source/UI/src/UIModule.cpp
```diff
@@ -15,13 +16,22 @@ namespace TitaniumWindows
...
+		: Titanium::UIModule(js_context)
+		, windowsStore__(js_context.CreateObject(JSExport<TitaniumWindows::WindowsModule>::Class()))
+	{
 	}
 
 	UIModule::~UIModule() TITANIUM_NOEXCEPT
 	{
 	}
 
+	// Initialize UI.WindowsPhone objects
```
sgtcoolguy: // Initialize Windows-specific UI types

Source/UI/src/UIModule.cpp
```diff
@@ -15,13 +16,22 @@ namespace TitaniumWindows
...
+	void UIModule::postInitialize(JSObject& js_object) 
+	{
+		Titanium::UIModule::postInitialize(js_object);
+		js_object.SetProperty("Windows", windowsStore__);
```
sgtcoolguy: windowsModule__ or just windows__ looks like best fit name?

Source/UI/src/Windows/AppBarButton.cpp
```diff
@@ -0,0 +1,93 @@
...
+			TITANIUM_PROPERTY_GETTER(AppBarButton, icon) 
+			{
+				return icon__;
+			}
+
+			TITANIUM_PROPERTY_SETTER(AppBarButton, icon)
+			{
+				icon__ = argument;
+
+				if (argument.IsNumber()) {
```
sgtcoolguy: Not sure how much we care because I don't think we expect anyone to extend these Windows-specific types, but obviously this is all HAL types/bridge functions and we're not mapping to standard types.

I think here it's ok because we're not expecting subclasses or anything so why jump through more method calls/conversion hoops?

Source/UI/src/Windows/CommandBar.cpp
```diff
@@ -0,0 +1,80 @@
...
+				JSExport<CommandBar>::SetParent(JSExport<Titanium::Module>::Class());
+				TITANIUM_ADD_PROPERTY(CommandBar, items);
+			}
+
+			TITANIUM_PROPERTY_GETTER(CommandBar, items)
+			{
+				return items__;
+			}
+
+			TITANIUM_PROPERTY_SETTER(CommandBar, items)
```
sgtcoolguy: We should enable use of the secondary commands property too. Not sure if we want another property or we want to overload items, so that if it is an array containing two arrays, the first inner array is primary and second is secondary?

May be more logical to have a second property?

Source/UI/src/Windows/SystemIcon.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+	namespace UI
+	{
+		namespace WindowsXaml
+		{
+			using namespace Windows::UI::Xaml;
+
+			SystemIcon::SystemIcon(const JSContext& js_context) TITANIUM_NOEXCEPT
+				: Titanium::Module(js_context)
+				, cancel__(js_context.CreateNumber(static_cast<std::uint32_t>(Controls::Symbol::Cancel)))
```
sgtcoolguy: We should expose _all_ the symbols. I know there are a lot...

https://msdn.microsoft.com/en-US/library/windows/apps/windows.ui.xaml.controls.symbol.aspx?f=255&MSPPError=-2147217396

sgtcoolguy:
I'm going to merge for now and I'll try and follow up with a commit or two to address some of my own comments...

PR #230, opened 2015-04-29
[TIMOB-18734] Fix: No way to navigate back to directory view

infosia:
Fix for [TIMOB-18734](https://jira.appcelerator.org/browse/TIMOB-18734)

- Add `windows:back` event which handles hardware back button
- Navigate back to directory window on back button event (Corporate app)

To check how this works on Corporate app, navigate to profile view and then press hardware back button and see if it navigates back to directory view.

### DIFF
--- a/Examples/Corporate/src/Assets/alloy/controllers/profile.js
+++ b/Examples/Corporate/src/Assets/alloy/controllers/profile.js
@@ -72,6 +72,9 @@ function Controller() {
         title: "Profile",
         id: "profile"
     });
+    $.__views.profile.addEventListener('windows:back', function () {
+        $.__views.profile.close();
+    });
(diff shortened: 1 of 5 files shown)


Examples/Corporate/src/Assets/alloy/controllers/profile.js
```diff
@@ -72,6 +72,9 @@ function Controller() {
         title: "Profile",
         id: "profile"
     });
+    $.__views.profile.addEventListener('windows:back', function () {
```
sgtcoolguy: This works fine for this particular example, but I'd think we should have default behavior of basically going back one in the Window stack (unless the user explicitly hangs their own back button handler like this one). So basically by default, do a window.close() unless the user has added their own event listener with a callback.

PR #232, opened 2015-04-30
[TIMOB-18855] Fix: new Date() returns wrong value

infosia:
It was originally because JavaScriptCore returns wrong value for `new Date()` so we just need to upgrade the library.


infosia:
Noticed that I accidentally closed it. reopening..

sgtcoolguy:
test this please

Sorry Kota, looks like I temporarily broke the builds when I killed an out of date build running... comment above will re-try build.

sgtcoolguy:
test this please

PR #235, opened 2015-04-30
[TIMOB-18758] Implement Ti.App events

infosia:
Implement basic Ti.App events

- `pause`, `paused`
- `resume`, `resumed`
- `keyboardframechanged`

### DIFF
--- a/Source/Titanium/src/TitaniumWindows.cpp
+++ b/Source/Titanium/src/TitaniumWindows.cpp
@@ -28,6 +28,14 @@
 #include <Windows.h>
 #include <collection.h>
 
+#define GET_TITANIUM_APP(VARNAME) \
+  JSValue Titanium_property = js_context__.get_global_object().GetProperty("Titanium"); \
+  TITANIUM_ASSERT(Titanium_property.IsObject()); \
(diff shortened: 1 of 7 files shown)


sgtcoolguy:
test this please

Sorry Kota, looks like I temporarily broke the builds when I killed an out of date build running... comment above will re-try build.

sgtcoolguy:
test this please

Source/Titanium/src/TitaniumWindows.cpp
```diff
@@ -78,6 +87,7 @@ namespace TitaniumWindows
 		                                                            .build());
 
 		Suspending += ref new Windows::UI::Xaml::SuspendingEventHandler(this, &Application::OnSuspending);
+		Resuming += ref new Windows::Foundation::EventHandler<::Platform::Object ^>(this, &Application::OnResuming);
```
sgtcoolguy: Totally unrelated to this, but I see there's an UnhandledException event on Application. Do we already hook something there to basically do our red screen of death? Seems like a good place to do a catch-all and do red screen of death / log.

PR #239, opened 2015-05-06
OptionDialog, ProgressBar & ScrollableView TitaniumKit

infosia:
- Ti.UI.OptionDialog (TitaniumKit & Windows implementation)
- Ti.UI.ProgressBar (TitaniumKit & Windows implementation)
- Ti.UI.ScrollableView (TitaniumKit)
- Small fix for AlertDialog etc

Sample code for ProgressBar:

```javascript
    backgroundColor: 'blue'
});

var bar = Ti.UI.createProgressBar({
    width: '70%',
    min: 0,
    max: 10,
    value: 0,
    color: '#fff',
    message: 'Downloading 0 of 10'
});
win.add(bar);

var id = setInterval(inc, 1000);

function inc() {
    if (bar.value >= bar.max) {
        clearTimeout(id);
        return;
    }
    bar.value = bar.value + 1;
    bar.message = 'Downloading ' + bar.value + ' of ' + bar.max
}

win.open();
```

Sample code for OptionDialog and AlertDialog

```javascript
    title: 'Click',
    backgroundColor:'blue'
});

var opts = {
    cancel: 1,
    options: ['Confirm', 'Cancel'],
    selectedIndex: 0,
    destructive: 1,
    title: 'Show Alert?'
};

    var dialog = Ti.UI.createOptionDialog(opts);
    dialog.addEventListener('click', function (_e) {
        if (!_e.cancel) {
            var alert_dialog = Ti.UI.createAlertDialog({
                cancel: 1,
                buttonNames: ['Confirm'],
                message: 'Would you like to close?',
                title: 'Alert Dialog'
            });
            alert_dialog.show();
        }
    });
    dialog.show();
});

win.add(button);
win.open();
```

### DIFF
--- a/Source/TitaniumKit/include/Titanium/UI/ScrollableView.hpp
+++ b/Source/TitaniumKit/include/Titanium/UI/ScrollableView.hpp
@@ -1,0 +1,260 @@
+/**
+ * TitaniumKit Titanium.UI.ScrollableView
+ *
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#ifndef _TITANIUM_UI_SCROLLABLEVIEW_HPP_
+#define _TITANIUM_UI_SCROLLABLEVIEW_HPP_
+
+#include "Titanium/Module.hpp"
+#include "Titanium/UI/Dimension.hpp"
+#include "Titanium/UI/View.hpp"
+
+namespace Titanium
+{
+	namespace UI
+	{
+
+		using namespace HAL;
+
+		/*!
+		  @class
+		  @discussion This is the Titanium ScrollableView Module.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/Titanium.UI.ScrollableView
+		*/
+		class TITANIUMKIT_EXPORT ScrollableView : public View, public JSExport<ScrollableView>
+		{
+
+		public:
+
+			/*!
+			  @property
+			  @abstract cacheSize
+			  @discussion Number of pages to cache (pre-render).
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, cacheSize);
+
+			/*!
+			  @property
+			  @abstract currentPage
+			  @discussion Index of the active page.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, currentPage);
+
+			/*!
+			  @property
+			  @abstract disableBounce
+			  @discussion Determines whether page bouncing effect is disabled.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, disableBounce);
+
+			/*!
+			  @property
+			  @abstract overScrollMode
+			  @discussion Determines the behavior when the user overscolls the view.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(JSValue, overScrollMode);
+
+			/*!
+			  @property
+			  @abstract pagingControlColor
+			  @discussion Color of the paging control, as a color name or hex triplet.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, pagingControlColor);
+
+			/*!
+			  @property
+			  @abstract pagingControlHeight
+			  @discussion Height of the paging control, in pixels.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, pagingControlHeight);
+
+			/*!
+			  @property
+			  @abstract showPagingControl
+			  @discussion Determines whether the paging control is visible.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, showPagingControl);
+
+			/*!
+			  @property
+			  @abstract pagingControlTimeout
+			  @discussion Number of milliseconds to wait before hiding the paging control.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, pagingControlTimeout);
+
+			/*!
+			  @property
+			  @abstract pagingControlAlpha
+			  @discussion Alpha value of the paging control.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(double, pagingControlAlpha);
+
+			/*!
+			  @property
+			  @abstract pagingControlOnTop
+			  @discussion Determines whether the paging control is displayed at the top or bottom of the view.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, pagingControlOnTop);
+
+			/*!
+			  @property
+			  @abstract overlayEnabled
+			  @discussion Determines whether the paging control is added as an overlay to the view.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, overlayEnabled);
+
+			/*!
+			  @property
+			  @abstract scrollingEnabled
+			  @discussion Determines whether scrolling is enabled for the view.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, scrollingEnabled);
+
+			/*!
+			  @property
+			  @abstract views
+			  @discussion Sets the pages within this Scrollable View.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::vector<std::shared_ptr<View>>, views);
+
+			/*!
+			  @property
+			  @abstract clipViews
+			  @discussion Determines whether the previous and next pages are clipped, so that they are not visible adjacent to the current page.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, clipViews);
+
+			/*!
+			  @property
+			  @abstract hitRect
+			  @discussion Sets the region where this view responds to gestures.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(Dimension, hitRect);
+
+			/*!
+			  @method
+			  @abstract addView
+			  @discussion Adds a new page to this Scrollable View.
+			*/
+			virtual void addView(const std::shared_ptr<View>& view) TITANIUM_NOEXCEPT;
+
+			/*!
+			  @method
+			  @abstract moveNext
+			  @discussion Sets the current page to the next consecutive page in <Titanium.UI.ScrollableView.views>.
+			*/
+			virtual void moveNext() TITANIUM_NOEXCEPT;
+
+			/*!
+			  @method
+			  @abstract movePrevious
+			  @discussion Sets the current page to the previous consecutive page in <Titanium.UI.ScrollableView.views>.
+			*/
+			virtual void movePrevious() TITANIUM_NOEXCEPT;
+
+			/*!
+			  @method
+			  @abstract removeView
+			  @discussion Removes an existing page from this Scrollable View.
+			*/
+			virtual void removeView(const std::shared_ptr<View>& view) TITANIUM_NOEXCEPT;
+
+			/*!
+			  @method
+			  @abstract scrollToView
+			  @discussion Scrolls to the specified page in <Titanium.UI.ScrollableView.views>.
+			*/
+			virtual void scrollToView(const std::shared_ptr<View>& view) TITANIUM_NOEXCEPT;
+
+			ScrollableView(const JSContext&) TITANIUM_NOEXCEPT;
+
+			virtual ~ScrollableView() = default;
+			ScrollableView(const ScrollableView&) = default;
+			ScrollableView& operator=(const ScrollableView&) = default;
+#ifdef TITANIUM_MOVE_CTOR_AND_ASSIGN_DEFAULT_ENABLE
+			ScrollableView(ScrollableView&&)                 = default;
+			ScrollableView& operator=(ScrollableView&&)      = default;
+#endif
+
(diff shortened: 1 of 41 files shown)


Source/TitaniumKit/include/Titanium/UI/ScrollableView.hpp
```diff
@@ -0,0 +1,260 @@
...
+			  @discussion Determines whether page bouncing effect is disabled.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, disableBounce);
+
+			/*!
+			  @property
+			  @abstract overScrollMode
+			  @discussion Determines the behavior when the user overscolls the view.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(JSValue, overScrollMode);
```
sgtcoolguy: Looks like this is Android-only and takes a Number. We should be consistent in how we treat this vs ProgressBar.style. Maybe have it take an uint32_t, but basically be unimplemented at this level, and the Android subclass would handle verifying it's a valid enum value?

Source/TitaniumKit/include/Titanium/UI/ScrollableView.hpp
```diff
@@ -0,0 +1,260 @@
...
+			  @discussion Determines whether the paging control is visible.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, showPagingControl);
+
+			/*!
+			  @property
+			  @abstract pagingControlTimeout
+			  @discussion Number of milliseconds to wait before hiding the paging control.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, pagingControlTimeout);
```
sgtcoolguy: I'd use std::chrono::milliseconds here like we do for things like setTimeout.

Source/TitaniumKit/include/Titanium/UI/ScrollableView.hpp
```diff
@@ -0,0 +1,260 @@
...
+			  @discussion Determines whether the previous and next pages are clipped, so that they are not visible adjacent to the current page.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, clipViews);
+
+			/*!
+			  @property
+			  @abstract hitRect
+			  @discussion Sets the region where this view responds to gestures.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(Dimension, hitRect);
```
sgtcoolguy: I try to alphabetize the properties/methods so they line up with the docs more easily, and to be able to more quickly scan the files for a particular property/method impl.

Source/TitaniumKit/include/Titanium/UIModule.hpp
```diff
@@ -10,6 +10,9 @@
 #define _TITANIUM_UI_HPP_
 
 #include "Titanium/Module.hpp"
+#include "Titanium/UI/ScrollableView.hpp"
```
sgtcoolguy: again, I'm nuts, but I'm trying to keep these alphabetized so we can easily match up against file/doc listings.

Source/TitaniumKit/src/UI/OptionDialog.cpp
```diff
@@ -0,0 +1,283 @@
...
+			const auto view = get_androidView();
+			if (view != nullptr) {
+				return view->get_object();
+			}
+			return get_context().CreateNull();
+		}
+
+		TITANIUM_PROPERTY_SETTER(OptionDialog, androidView)
+		{
+			TITANIUM_ASSERT(argument.IsObject());
```
sgtcoolguy: Should we be using TITANIUM_ASSERT_AND_THROW?

Source/TitaniumKit/src/UI/OptionDialog.cpp
```diff
@@ -0,0 +1,283 @@
...
+			std::vector<JSValue> options;
+			for (auto option : get_options()) {
+				options.push_back(get_context().CreateString(option));
+			}
+			return get_context().CreateArray(options);
+		}
+
+		TITANIUM_PROPERTY_SETTER(OptionDialog, options)
+		{
+			TITANIUM_ASSERT(argument.IsObject());
```
sgtcoolguy: Looks like we have a couple cases of converting between std::vector<std::string> and an JSArray of JSStrings. Maybe we should have a macro for these conversions?

Source/TitaniumKit/src/UI/OptionDialog.cpp
```diff
@@ -0,0 +1,283 @@
...
+			TITANIUM_ASSERT(argument.IsString());
+			set_titleid(static_cast<std::string>(argument));
+			return true;
+		}
+
+		TITANIUM_FUNCTION(OptionDialog, show)
+		{
+			ENSURE_OPTIONAL_OBJECT_AT_INDEX(params, 0);
+
+			show(params.GetPrivate<OptionDialogShowParams>());
```
sgtcoolguy: Don't we need to convert the object into the struct? This is an abstract type that users will never construct on their own.

Source/TitaniumKit/src/UI/OptionDialog.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+			return get_context().CreateUndefined();
+		}
+
+		TITANIUM_FUNCTION(OptionDialog, hide)
+		{
+			ENSURE_OPTIONAL_OBJECT_AT_INDEX(params, 0);
+
+			hide(params.GetPrivate<OptionDialogHideParams>());
```
sgtcoolguy: Don't we need to convert the object into the struct? This is an abstract type that users will never construct on their own.

Source/TitaniumKit/src/UI/ProgressBar.cpp
```diff
@@ -0,0 +1,189 @@
...
+namespace Titanium
+{
+	namespace UI
+	{
+
+
+		ProgressBar::ProgressBar(const JSContext& js_context) TITANIUM_NOEXCEPT
+			: View(js_context),
+			color__(""),
+			max__(0),
```
sgtcoolguy: Looks like iOS uses default max of 1.0, min of 0.0, value of 0.0.

I think Android does the same, but it does some conversions to ints.

Source/TitaniumKit/src/UI/ProgressBar.cpp
```diff
@@ -0,0 +1,189 @@
...
+			TITANIUM_ADD_FUNCTION(ProgressBar, setMessage);
+			TITANIUM_ADD_FUNCTION(ProgressBar, getMin);
+			TITANIUM_ADD_FUNCTION(ProgressBar, setMin);
+			TITANIUM_ADD_FUNCTION(ProgressBar, getStyle);
+			TITANIUM_ADD_FUNCTION(ProgressBar, setStyle);
+			TITANIUM_ADD_FUNCTION(ProgressBar, getValue);
+			TITANIUM_ADD_FUNCTION(ProgressBar, setValue);
+		}
+
+		TITANIUM_PROPERTY_GETTER(ProgressBar, color)
```
sgtcoolguy: Also, looks like we could generate a macro for defining both getter and setter for common types like std:string, i.e.:
TITANIUM_PROPERTY_READWRITE_STRING(ProgressBar, color);

Source/TitaniumKit/src/UI/ProgressBar.cpp
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		TITANIUM_PROPERTY_SETTER(ProgressBar, font)
+		{
+			TITANIUM_ASSERT(argument.IsObject());
+			set_font(js_to_Font(static_cast<JSObject>(argument)));
+			return true;
+		}
+
+		TITANIUM_PROPERTY_GETTER(ProgressBar, max)
```
sgtcoolguy: macro for double getter/setter?

Source/TitaniumKit/src/UI/ScrollableView.cpp
```diff
@@ -0,0 +1,402 @@
...
+	namespace UI
+	{
+
+
+		ScrollableView::ScrollableView(const JSContext& js_context) TITANIUM_NOEXCEPT
+			: View(js_context),
+			cacheSize__(0),
+			currentPage__(0),
+			disableBounce__(false),
+			pagingControlColor__("Black"),
```
sgtcoolguy: "black"

Source/TitaniumKit/src/UI/ScrollableView.cpp
```diff
@@ -0,0 +1,402 @@
...
+		}
+
+		TITANIUM_PROPERTY_GETTER(ScrollableView, cacheSize)
+		{
+			return get_context().CreateNumber(get_cacheSize());
+		}
+
+		TITANIUM_PROPERTY_SETTER(ScrollableView, cacheSize)
+		{
+			TITANIUM_ASSERT(argument.IsNumber());
```
sgtcoolguy: macro for uint32_t property getter/setters?

Source/TitaniumKit/src/UI/ScrollableView.cpp
```diff
@@ -0,0 +1,402 @@
...
+			set_currentPage(static_cast<std::uint32_t>(argument));
+			return true;
+		}
+
+		TITANIUM_PROPERTY_GETTER(ScrollableView, disableBounce)
+		{
+			return get_context().CreateBoolean(get_disableBounce());
+		}
+
+		TITANIUM_PROPERTY_SETTER(ScrollableView, disableBounce)
```
sgtcoolguy: macro for bool property getter/setters?

Source/UI/src/OptionDialog.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <ppltasks.h>
+
+namespace TitaniumWindows
+{
+	namespace UI
+	{
+		using namespace Windows::UI::Popups;
+		using namespace Windows::Foundation;
+
+		OptionDialog::OptionDialog(const JSContext& js_context) TITANIUM_NOEXCEPT
```
sgtcoolguy: OptionDialog and AlertDialog are so very similar. Can we define one in terms of the other? Maybe we can get away with just having AlertDialog be an OptionDialog under the hood (i.e. just define it in JS and have it really instantiate an OptionDialog), so we don't need two implementations?

Source/UI/src/ProgressBar.cpp
```diff
@@ -0,0 +1,91 @@
...
+			panel__->HorizontalAlignment = HorizontalAlignment::Left;
+			label__->HorizontalAlignment = HorizontalAlignment::Center;
+
+			panel__->Children->Append(label__);
+			panel__->Children->Append(bar__);
+
+			Titanium::UI::ProgressBar::setLayoutDelegate<WindowsViewLayoutDelegate>();
+			layoutDelegate__->set_defaultHeight(Titanium::UI::LAYOUT::FILL);
+			layoutDelegate__->set_defaultWidth(Titanium::UI::LAYOUT::FILL);
+			getViewLayoutDelegate<WindowsViewLayoutDelegate>()->setComponent(panel__);
```
sgtcoolguy: I don't think we offer any way to change the IsIndeterminate flag for the ProgressBar, but may want to be extra safe and do as we did with the ActivityIndicator and forcibly set it to false on hide, and then back to it's original value on show - as it has the same performance concerns about using resources by animating even when not visible if IsIndeterminate is true.

Source/UI/src/ProgressBar.cpp
```diff
@@ -0,0 +1,91 @@
...
+		void ProgressBar::set_value(const double& value) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ProgressBar::set_value(value);
+			bar__->Value = value;
+		}
+
+		void ProgressBar::set_color(const std::string& colorName) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ProgressBar::set_color(colorName);
+			bar__->Foreground = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
```
sgtcoolguy: I know we haven't touched on this for any of the proxies we've already done, and maybe we don't care until much later - but should we consider a Color type at the TitaniumKit level so we can enforce the string is a valid value up there before we try converting to the platform-specific type?

It's probably too much hassle with no gain right now to worry about it.

Source/UI/src/OptionDialog.cpp
infosia: Yes, on Windows they are very similar. Regarding "merging" them, because we define AlertDialog and OptionDialog in TitaniumKit they must be subclass of `Titanium::UI::AlertDialog` and `Titanium::UI::OptionDialog`, which makes it difficult to merge their implementations...

Source/UI/src/ProgressBar.cpp
infosia: Nice catch yes. I'll create ticket for it.

infosia:
OptionDialog and ProgressBar should be ready. ScrollableView for Windows is not ready yet, will be pushed in subsequent PR.

Source/UI/src/OptionDialog.cpp
sgtcoolguy: I was thinking perhaps we could define one (say AlertDialog) in JS code, and it could basically wrap an OptionDialog and forward on method calls or something? I'm not sure, maybe that's not much better...

PR #244, opened 2015-05-14
Stub out properties/methods/constants of Titanium.Media

infosia:
~~NOT READY FOR MERGE~~

- [TIMOB-18761] Implement Ti.Network change event
- [TIMOB-18864] Stub Ti.Media module with constants, stubs for methods
  - Ti.Media
  - Ti.Media.CameraOptionsType
  - Ti.Media.Item
  - Ti.Media.Sound
  - Ti.Media.MediaQueryInfoType
  - Ti.Media.MediaQueryType
  - Ti.Media.MusicLibraryResponseType
  - Ti.Media.MusicLibraryOptionsType
  - Ti.Media.PhotoGalleryOptionsType
  - Ti.Media.PreviewImageOptions
  - Ti.Media.RouteDescription
  - Ti.Media.AudioPlayer
  - Ti.Media.AudioRecorder
  - Ti.Media.MusicPlayer
  - Ti.Media.VideoPlayer
  - Ti.Media.CameraMediaItemType
  - Ti.ErrorResponse


Source/Network/src/Network.cpp
```diff
@@ -9,20 +9,33 @@
 
 namespace TitaniumWindows
 {
+	using namespace Windows::Networking::Connectivity;
+
 	NetworkModule::NetworkModule(const JSContext& js_context) TITANIUM_NOEXCEPT
 		: Titanium::NetworkModule(js_context)
 	{
 		TITANIUM_LOG_DEBUG("TitaniumWindows::NetworkModule::ctor Initialize");
+		change_event__ = NetworkInformation::NetworkStatusChanged += ref new NetworkStatusChangedEventHandler([this](Platform::Object^ sender) {
```
sgtcoolguy: I haven't really looked into this, but do we think this has some performance cost if we have no change listeners? It may be better to hook this in enable/disableEvent and not track with a bool...

Source/TitaniumKit/include/Titanium/Media/AudioPlayer.hpp
```diff
@@ -0,0 +1,283 @@
...
+			  @discussion Size of the buffer used for streaming, in bytes.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, bufferSize);
+
+			/*!
+			  @property
+			  @abstract time
+			  @discussion Current playback position of the audio.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::chrono::milliseconds, time);
```
sgtcoolguy: I wonder if there should be a difference between progress and time. It looks like a bug in the API to me and they both appear to be the same thing (just sued on different platforms?) cc @ingo @rblalock

Source/TitaniumKit/include/Titanium/Media/MusicLibraryOptionsType.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+		/*!
+		  @struct
+		  @discussion An Object describing the current audio route.
+		  This is an abstract type. Any object meeting this description can be used where this type is used.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/MusicLibraryOptionsType
+		*/
+		struct MusicLibraryOptionsType
+		{
+			JSValue cancel;
```
sgtcoolguy: Can/Should we use some other type to describe callbacks? Maybe use JSFunction, or JSObject? Maybe we can use the slots/signals stuff to better describe the callback (i..e what args it passes along)?

Source/TitaniumKit/include/Titanium/Media/PhotoGalleryOptionsType.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Titanium
+{
+	namespace Media
+	{
+		using namespace HAL;
+
+		/*!
+		  @struct
+		  @discussion An Object describing the current audio route.
```
sgtcoolguy: Looks like this struct is a (mistaken) copy of a different one? It's missing a number of properties from the link given.

Source/TitaniumKit/include/Titanium/Media/PreviewImageOptions.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+		/*!
+		  @struct
+		  @discussion An Object describing the current audio route.
+		  This is an abstract type. Any object meeting this description can be used where this type is used.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/PreviewImageOptions
+		*/
+		struct PreviewImageOptions
+		{
+			JSValue error;
```
sgtcoolguy: missing image property (a Blob)

Source/TitaniumKit/include/Titanium/Media/RouteDescription.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+		/*!
+		  @struct
+		  @discussion An Object describing the current audio route.
+		  This is an abstract type. Any object meeting this description can be used where this type is used.
+		  See http://docs.appcelerator.com/titanium/latest/#!/api/RouteDescription
+		*/
+		struct RouteDescription
+		{
+			JSValue callback;
```
sgtcoolguy: this should have 'inputs' and 'outputs' properties, which are each an array of AUDIO_SESSION_PORT enum values

Source/TitaniumKit/include/Titanium/Media/Sound.hpp
```diff
@@ -0,0 +1,229 @@
...
+			  @discussion Determines whether the audio should continue playing even when its activity is paused.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, allowBackground);
+
+			/*!
+			  @property
+			  @abstract duration
+			  @discussion Duration of the audio resource.
+			*/
+			TITANIUM_PROPERTY_IMPL_READONLY_DEF(std::chrono::milliseconds, duration);
```
sgtcoolguy: Docs says this is ms for Android, seconds for iOS. Not sure what to do here. Maybe use a more generic  std::chrono::duration so we can report the number and the measurement type?

sgtcoolguy: @ingo @rblalock looks like another case of inconsistent APIs. We report seconds on IOS, milliseconds on Android. Also note that the time property used to do the same, but as of 3.2+ iOS time is reported in milliseconds.

Source/TitaniumKit/include/Titanium/MediaModule.hpp
```diff
@@ -0,0 +1,587 @@
...
+		  @discussion Takes a screen shot of the visible UI on the device.
+		*/
+		virtual void takeScreenshot(JSValue callback) TITANIUM_NOEXCEPT;
+
+		/*!
+		  @method
+		  @abstract vibrate
+		  @discussion Makes the device vibrate.
+		*/
+		virtual void vibrate(std::vector<std::uint32_t> pattern) TITANIUM_NOEXCEPT;
```
sgtcoolguy: maybe make this a std::vector<std::chrono::milliseconds>? the numbers ar emeant to be durations in ms for the vibration durations and delays.

Source/TitaniumKit/include/Titanium/Media/MusicLibraryOptionsType.hpp
infosia: Yes, that's a good point. The reason why I use `JSValue` to describe callback is that it can be set to `null` or `undefined`. Also, because it is a natural JavaScript object, you can not map C++ function quite naturally because JS function object itself may have some custom properties & functions. But in spite of these limitations I feel that mapping C++ slots/signals with JSObject may be interesting idea worth trying. 

So what I'm trying to figure out now is that if we could provide natural mapping between JSObject and std::function like below...I didn't want to use `boost::signals2` on HAL because I don't want HAL depends on `boost`.

```c++
bool test(int i, std::string str) {
  std::cout << "It worked!" << std::endl;
  return true;
}
JSFunction  js_function = js_context.CreateFunction<bool(int,std::string)>(test);
globalObject.SetProperty("Test", js_function);

/*
 * From JavaScript:
 * Test(101, "Yay");
 */
```

Source/TitaniumKit/include/Titanium/Media/Sound.hpp
infosia: Looks like `std::chrono::duration` always works with template? I would just use `std::uint32_t` here.

infosia:
Should be ready now.

sgtcoolguy:
test this please

sgtcoolguy:
test this please

sgtcoolguy:
test this please

PR #246, opened 2015-05-20
[TIMOB-18928] Reduce number of VS module projects

infosia:
Before adding `Ti.Media` module (TIMOB-18880), we need to reduce number of VS projects due to the reference/linkage issue on VS described at [TIMOB-18774](https://jira.appcelerator.org/browse/TIMOB-18774). I ended up moving `Ti.App`, `Ti.API`, `Ti.Platform` and `Ti.Utils` into `Ti` project so that I can add new VS project for `Ti.Media`.


infosia:
test this please

PR #249, opened 2015-05-22
[TIMOB-18880] Implement Ti.Media methods

infosia:
Implement part of `Ti.Media` methods. It's not yet completed but because we'll focus on packaging overall in next sprint I'd like to push this as current state. Note that because there are some differences between Windows Store and Windows Phone API especially for media functions, we need implementations for both platforms.

- Ti.Media.showCamera and takePicture
- Ti.Media.vibrate
- Ti.Media.volume event
- Ti.Media.start/stopMicrophoneMonitor
- Ti.Media.AudioRecorder (partially)

### DIFF
--- a/Source/Media/src/Media.cpp
+++ b/Source/Media/src/Media.cpp
@@ -1,0 +1,308 @@
+/**
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#include "TitaniumWindows/Media.hpp"
+#include "Titanium/Media/CameraOptionsType.hpp"
+#include "TitaniumWindows/Utility.hpp"
+#include <ppltasks.h>
+
+namespace TitaniumWindows
+{
+	using namespace Windows::Foundation;
+	using namespace Windows::Storage;
+	using namespace Windows::Media;
+	using namespace Windows::Media::Capture;
+	using namespace Windows::Media::MediaProperties;
+
+	void MediaModule::beep() TITANIUM_NOEXCEPT
+	{
+		TITANIUM_LOG_WARN("MediaModule::beep: Unimplemented");
+	}
+
+	void MediaModule::hideCamera() TITANIUM_NOEXCEPT
+	{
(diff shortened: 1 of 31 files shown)


Source/Media/src/Media.cpp
```diff
@@ -0,0 +1,308 @@
...
+	std::vector<std::shared_ptr<Titanium::Media::Item>> MediaModule::queryMusicLibrary(const Titanium::Media::MediaQueryType& query) TITANIUM_NOEXCEPT
+	{
+		TITANIUM_LOG_WARN("MediaModule::queryMusicLibrary: Unimplemented");
+		return std::vector<std::shared_ptr<Titanium::Media::Item>>();
+	}
+
+	void MediaModule::fireMediaVolumeEvent(const SoundLevel& level) {
+		// There's only three level on Windows: Muted, Low and Full.
+		// TODO: Find reasonable number for volume. Right now it's mapped to Muted=0, Low=0.5 and Full=1.
+		auto obj = get_context().CreateObject();
```
sgtcoolguy: assign context to temp var and re-use it to avoid calling get_context() multiple times?

Source/Media/src/Media.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+	void MediaModule::vibrate(std::vector<std::chrono::milliseconds> pattern) TITANIUM_NOEXCEPT
+	{
+#if WINAPI_FAMILY == WINAPI_FAMILY_PHONE_APP
+		using namespace Windows::Phone::Devices::Notification;
+		const auto device = VibrationDevice::GetDefault();
+		Windows::Foundation::TimeSpan duration;
+		for (const auto msec : pattern) {
+			duration.Duration = msec.count() * 10000;
+			device->Vibrate(duration); // does this a blocking function so we can push durations like this?
```
sgtcoolguy: We need to push the vibration durations, but also the "pauses" between. Looks like we need to schedule a series of them using some async libs, basically we take the first value as the initial delay, then next as the vibration duration, then wait for next value, then vibrate for next, etc.

sgtcoolguy:
I think we'll need to keep the original JIRA ticket open and push it to later sprints, but I'll merge what we have so far for now.

PR #253, opened 2015-05-28
[TIMOB-18937][TIMOB-18938][TIMOB-18939][TIMOB-18942] Add module support to the CLI - READY

garymathews:
##### READY FOR REVIEW
- Created project structure for native modules
- Created templates for module project generation
- Implemented a CLI hook to generate the Visual Studio projects using CMake
- Updated build script to copy over necessary header files (HAL and TitaniumKit) into the SDK ```windows/lib/*/include``` directory for successful module compilation
- Added ```WindowsStore``` ```ARM``` to build script

##### DEPENDS ON
[TIMOB-18978](https://jira.appcelerator.org/browse/TIMOB-18978) - https://github.com/appcelerator/titanium_mobile/pull/6892

##### NOTES
- I have updated the ```build.js``` script to use ```fs-extra``` as I was having problems with ```wrench```
- @infosia Do you still need ```classname``` or can you use ```moduleid```?

##### JIRA TICKETS
**EPIC** [TIMOB-17911](https://jira.appcelerator.org/browse/TIMOB-18937)
- [TIMOB-18937](https://jira.appcelerator.org/browse/TIMOB-18937)
- [TIMOB-18938](https://jira.appcelerator.org/browse/TIMOB-18938)
- [TIMOB-18939](https://jira.appcelerator.org/browse/TIMOB-18939)
- [TIMOB-18942](https://jira.appcelerator.org/browse/TIMOB-18942)

### DIFF
--- a/templates/module/default/template/windows/CMakeLists.txt.ejs
+++ b/templates/module/default/template/windows/CMakeLists.txt.ejs
@@ -1,0 +1,120 @@
+# Titanium Windows Native Module - <%- moduleName %>
+#
+# Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+# Licensed under the terms of the Apache Public License.
+# Please see the LICENSE included with this distribution for details.
+
+cmake_minimum_required(VERSION 3.0.0)
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "WindowsPhone")
+  set(PLATFORM wp)
+  add_definitions("-DPHONE")
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "WindowsStore")
+  set(PLATFORM store)
+else()
+  message(FATAL_ERROR "This app supports Store / Phone only.")
+endif()
+
+project(<%- moduleName %>)
+
+set(<%- moduleName %>_VERSION 0.1.0)
+
+set(WINDOWS_SOURCE_DIR "<%- tisdkPath.replace(/\\/g,'/') %>/windows")
+
+SET(CMAKE_FIND_LIBRARY_PREFIXES "")
+SET(CMAKE_FIND_LIBRARY_SUFFIXES ".lib" ".dll")
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+set(CMAKE_INCLUDE_CURRENT_DIR_IN_INTERFACE ON)
+
+option(<%- moduleName %>_DISABLE_TESTS "Disable compiling the tests" OFF)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+get_filename_component(APPCELERATOR_CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake ABSOLUTE)
+list(INSERT CMAKE_MODULE_PATH 0 ${APPCELERATOR_CMAKE_MODULE_PATH})
+
+set(Boost_USE_STATIC_LIBS ON )
+set(Boost_USE_MULTITHREADED ON )
+set(Boost_USE_STATIC_RUNTIME OFF)
+find_package(Boost 1.55.0 REQUIRED)
+
+find_package(HAL REQUIRED)
+find_package(TitaniumKit REQUIRED)
+find_package(JavaScriptCore REQUIRED)
+
+enable_testing()
+
+set(SOURCE_<%- moduleName %>
+  include/<%- moduleName %>.hpp
+  src/<%- moduleName %>.cpp
+  )
+
+source_group(<%- moduleName %> FILES ${SOURCE_<%- moduleName %>})
+
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+
+add_library(<%- moduleName %> SHARED
+  ${SOURCE_<%- moduleName %>}
+  )
+
+include(GenerateExportHeader)
+generate_export_header(<%- moduleName %>)
+target_compile_definitions(<%- moduleName %> PRIVATE <%- moduleName %>_EXPORTS)
+
+target_include_directories(<%- moduleName %> PUBLIC
+  ${PROJECT_SOURCE_DIR}/include
+  $<TARGET_PROPERTY:HAL,INTERFACE_INCLUDE_DIRECTORIES>
+  $<TARGET_PROPERTY:TitaniumKit,INTERFACE_INCLUDE_DIRECTORIES>
+  ${JavaScriptCore_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS}
+  )
+
+target_link_libraries(<%- moduleName %>
+  ${Boost_LIBRARIES}
+  HAL
+  TitaniumKit
+  )
+
+set_target_properties(<%- moduleName %> PROPERTIES VS_WINRT_COMPONENT TRUE)
+
+set_property(TARGET <%- moduleName %> APPEND_STRING PROPERTY LINK_FLAGS_DEBUG "/OPT:NOREF /OPT:NOICF")
+
+if (NOT <%- moduleName %>_DISABLE_TESTS)
+  add_subdirectory(test)
+endif()
+
+set_property(TARGET <%- moduleName %> PROPERTY VERSION ${<%- moduleName %>_VERSION})
+set_property(TARGET <%- moduleName %> PROPERTY SOVERSION 0)
+set_property(TARGET <%- moduleName %> PROPERTY INTERFACE_<%- moduleName %>_MAJOR_VERSION 0)
+set_property(TARGET <%- moduleName %> APPEND PROPERTY
+  COMPATIBLE_INTERFACE_STRING <%- moduleName %>_MAJOR_VERSION
+  )
+
+install(TARGETS <%- moduleName %> EXPORT <%- moduleName %>_Targets
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin
+  INCLUDES DESTINATION include
+  )
+
+export(EXPORT <%- moduleName %>_Targets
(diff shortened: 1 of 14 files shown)


infosia:
@garymathews How does the final module structure look like? I'm expecting it expands like below just like other Ti modules does.

- YOUR_MODULE
  - include
    - YOUR_MODULE.hpp
  - phone
    - ARM
        - YOUR_MODULE.dll
        - YOUR_MODULE.winmd
        - YOUR_MODULE.lib
        - etc
    - x86
        - YOUR_MODULE.dll
        - YOUR_MODULE.winmd
        - YOUR_MODULE.lib
        - etc
  - store
    - x86
        - YOUR_MODULE.dll
        - YOUR_MODULE.winmd
        - YOUR_MODULE.lib
        - etc

infosia:
test this please

infosia:
@garymathews Would you fix merge conflict please?

How do I create/build the module through Ti CLI? I tried `ti build` according to README but it says `timodule.xml not found`.

```
$ ti create -t module --id com.example.test1 -p windows -n com.example.test1
$  ti build --platform windows --build-type production --dir com.example.test1
[ERROR] Invalid project directory "." because tiapp.xml or timodule.xml not found
```

Also, I found that when I use module id with package name like `com.example.test1`, it generates invalid class name like below.

```
class TITANIUMWINDOWS_COM.EXAMPLE.TEST1_EXPORT com.example.test1 : public JSExport<com.example.test1>
```

I would suggest we want to use with capital case namespace like `Com::Example::Test1` in this case. Also, can you add new entry *`classname`* in `manifest` file to describe the C++ class name? I'm planning to use it from Titanium CLI build command to handle require hook. It's gonna be like below:

```
#
# this is your module manifest and used by Titanium
# during compilation, packaging, distribution, etc.
#
version: 1.0.0
...

# these should not be edited
name: com.example.test1
moduleid: com.example.test1
guid: 0f3c4ec0-d56a-4e31-a287-7a5d77756440
platform: windows
minsdk: 4.1.0.v20150603181231
classname: Com::Example::Test1
```

infosia:
Also, how should cmake project name look like when you use the module name like `com.example.test1`? I saw it creates cmake project name like `TitaniumWindows_com.example.test1`, but I think we don't want to use `TitaniumWindows` here, I would say use camel case like `ComExampleTest1`?

```
project(TitaniumWindows_com.example.test1)
↓
project(ComExampleTest1)
```

This recalls one more request, could you add new entry *`projectname`* in `manifest` file to describe cmake project name like below?

```
classname: Com::Example::Test1
projectname: ComExampleTest1
```

Thanks,

templates/module/default/template/windows/include/TitaniumWindows/{{ModuleName}}.hpp.ejs
```diff
@@ -0,0 +1,39 @@
...
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#ifndef _TITANIUMWINDOWS_<%- moduleName.toUpperCase() %>_HPP_
+#define _TITANIUMWINDOWS_<%- moduleName.toUpperCase() %>_HPP_
+
+#include "TitaniumWindows_<%- moduleName %>_EXPORT.h"
+
+#include "Titanium/Titanium.hpp"
+#include "HAL/HAL.hpp"
```
infosia: I don't think we need to include `Titanium/Titanium.hpp` for the module because it does not use any Titanium API typs. I would just include `Titanium/detail/TiBase.hpp` just like how [Module.hpp](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/include/Titanium/Module.hpp#L12) does.

```
#include "Titanium/detail/TiBase.hpp"
```

templates/module/default/template/windows/include/TitaniumWindows/{{ModuleName}}.hpp.ejs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace TitaniumWindows
```
infosia: I think we want to use unique namespace here based on module id. For example `com.example.test1` would have `Test1` class with `Com::Example` namespace.

templates/module/default/template/windows/manifest.ejs
```diff
@@ -0,0 +1,18 @@
...
+author: <%- author %>
+license: Specify your license
+copyright: Copyright (c) <%- (new Date).getFullYear() %> by <%- publisher %>
+
+# these should not be edited
+name: <%- moduleName %>
+moduleid: <%- moduleId %>
+guid: <%- guid %>
+platform: <%- platform %>
+minsdk: <%- tisdkVersion %>
```
infosia: @garymathews  Can you add `classname` and `projectname` entry as I noted in https://github.com/appcelerator/titanium_mobile_windows/pull/253#issuecomment-108683744?

templates/module/default/template/windows/build.js
```diff
@@ -0,0 +1,128 @@
...
+	// empty any existing module directory
+	fs.emptyDirSync(moduleDir);
+
+	// copy necesary folders and files
+	fs.copySync(path.join(__dirname, '..', 'documentation'), path.join(moduleDir, 'documenation'));
+	fs.copySync(path.join(__dirname, '..', 'example'), path.join(moduleDir, 'example'));
+	fs.copySync(path.join(__dirname, '..', 'assets'), path.join(moduleDir, 'assets'));
+	fs.copySync(path.join(__dirname, '..', 'LICENSE'), path.join(moduleDir, 'LICENSE'));
+	fs.copySync(path.join(__dirname, 'manifest'), path.join(moduleDir, 'manifest'));
+	fs.copySync(path.join(__dirname, 'include'), path.join(moduleDir, 'include'));
```
infosia: I think we want to place `timodule.xml` in `moduleDir` because `ti build` command complains about it. (https://github.com/appcelerator/titanium_mobile_windows/pull/253#issuecomment-108683744)

sgtcoolguy:
@garymathews @infosia Can't we just use the moduleId to determine the class and project name? I assume that's what Gary is doing anyways. I'd assume we'd just transform the moduleId to derive the classname and project name.

infosia:
Yes that's a good point. I just wanted to give more flexibility to it, as I'm not sure if developer happens to need different class/project name.

templates/module/default/template/windows/CMakeLists.txt.ejs
```diff
@@ -0,0 +1,119 @@
...
+if(${CMAKE_SYSTEM_NAME} STREQUAL "WindowsPhone")
+  set(PLATFORM wp)
+  add_definitions("-DPHONE")
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "WindowsStore")
+  set(PLATFORM store)
+else()
+  message(FATAL_ERROR "This app supports Store / Phone only.")
+endif()
+
+project(TitaniumWindows_<%- moduleName %>)
```
sgtcoolguy: I think Kota is right, we don't want the third-party modules to use a TitaniumWindows_ prefix for their module /project names.

templates/module/default/template/windows/build_and_test.sh.ejs
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+# Titanium Windows Native Module - <%- moduleName %>
+#
+# Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+# Licensed under the terms of the Apache Public License.
+# Please see the LICENSE included with this distribution for details.
+
+if ! test -d "${GTEST_ROOT}"; then
```
sgtcoolguy: This is only needed if they're writing tests for their module. I'm not sure we should require it.

templates/module/default/template/windows/cmake/FindHAL.cmake
```diff
@@ -0,0 +1,39 @@
+# FindHAL
```
sgtcoolguy: These Find scripts are just copies of the ones we already have in https://github.com/appcelerator/titanium_mobile_windows/tree/master/templates/build/cmake right? Can't we just copy those, rather than have another copy of the file in the repo? I'm worried we'll have to maintain both.

templates/module/default/template/windows/src/{{ModuleName}}.cpp.ejs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace TitaniumWindows
+{
+	<%- moduleName %>::<%- moduleName %>(const JSContext& js_context, const std::vector<JSValue>& arguments) TITANIUM_NOEXCEPT
+	{
+		TITANIUM_LOG_DEBUG("<%- moduleName %>::ctor Initialize");
+	}
+
+	void <%- moduleName %>::postInitialize(JSObject& js_object) {
+		HAL_LOG_DEBUG("<%- moduleName %>:: postInitialize ", this);
```
sgtcoolguy: I don't think we want HAL debug log macros here.

templates/module/default/template/windows/src/{{ModuleName}}.cpp.ejs
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	void <%- moduleName %>::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments)
+	{
+		HAL_LOG_DEBUG("<%- moduleName %>:: postCallAsConstructor ", this);
+	}
+
+	void <%- moduleName %>::JSExportInitialize() {
+		JSExport<<%- moduleName %>>::SetClassVersion(1);
```
sgtcoolguy: Should we add comments to guide the user on what they should be doing to start implementing?

templates/module/default/template/windows/include/TitaniumWindows/{{ModuleName}}.hpp.ejs
sgtcoolguy: Agreed, and that will also in turn include HAL/HAL.hpp, so you could then remove that.

We need to require TiBase anyways to get all the nice macros for defining properties/methods.

PR #254, opened 2015-06-01
Integrate native module loading - NOT READY

infosia:
*NOT READY FOR MERGE*

Part of [TIMOB-18940](https://jira.appcelerator.org/browse/TIMOB-18940).

Important changes here:

- `Global` module now handles all native module loading (including `Windows` types)
- `Global` module will be built dynamically during Ti `build`.
- NativeModuleLoader will be generated dynamically during Ti `build`.
- Ti `build` command should copy Global module DLL and other native modules into the app project


sgtcoolguy:
@infosia Most of this looks good. The issue is having to rebuild the full Global module as part of ti build. I don't think we can realistically do that, as we'll need to have the user have boost installed and set up on their system to be able to compile it.

I think we need a mechanism whereby ti build builds the user's app _and_ a separate module that has the "native" code/wrappers/hooks embedded that we compile, and it needs to avoid usage of boost or any other such third-party dependencies that won't already be on their system with VS 2013. The hook could be registered in the same app callback as the TitaniumWindows::Application gets instantiated, as I think so long as it's a separate module we'd be fine loading there.

infosia:
> I don't think we can realistically do that, as we'll need to have the user have boost installed and set up on their system to be able to compile it.

@sgtcoolguy ok, makes sense.

infosia:
Closing as we have better solution by #255

PR #255, opened 2015-06-01
Windows Native Wrappers - NOT READY

sgtcoolguy:
A work in progress on generating wrappers as part of build process, hooking them up to the application/framework.

PR #256, opened 2015-06-02
 Integrate module loading through RequireHook - NOT READY

infosia:
Taking over #255 

- Integrate module loading through RequireHook
- Save reference to RequireHook to prevent from GC
- Fix include/link error around TitaniumWindows_Native

TODO:
- Integrate native module loading based on `tiapp.xml`


infosia:
Close PR, as we started working together based on #255

PR #258, opened 2015-06-03
[TIMOB-18087] Packaging

dawsontoth:
Still a WIP but it's working with some hard coded values. Let the sanity
check begin...

To test this at the moment, you'll need to add this to your tiapp.xml:

    <property name="ti.windows.publishername" type="string">CN=Dawson Toth</property>

And when prompted during the build, use the password "**testing**" for your certs. A future iteration will prompt you for the password.

I've hacked together the dist-phonestore command:

    ti build -p windows -T dist-phonestore

You'll also need the PR on windowslib -- https://github.com/appcelerator/windowslib/pull/16 -- in your node modules.

https://jira.appcelerator.org/browse/TIMOB-18087

### DIFF
--- a/cli/hooks/wp-package.js
+++ b/cli/hooks/wp-package.js
@@ -2,7 +2,7 @@
  * Signs an app and copies the result to the specified output directory.
  *
  * @copyright
- * Copyright (c) 2014 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
  *
  * @license
  * Licensed under the terms of the Apache Public License
@@ -23,33 +23,29 @@ exports.init = function (logger, config, cli) {
 	cli.on('build.post.compile', {
 		priority: 10000,
 		post: function (builder, finished) {
-			if (builder.buildOnly || builder.target !== 'dist-phonestore') return finished();
+			if (builder.buildOnly || builder.target !== 'dist-phonestore') {
+				return finished();
+			}
 
-			/*
-			TODO: sign the app. all you need to do is:
+			var wpARMDir = builder.cmakeTargetDir,
+				releaseDir = path.join(wpARMDir, 'Release'),
+				bundle = path.join(releaseDir, fs.readdirSync(releaseDir).filter(function (f) {
+					return f.indexOf('_Bundle') >= 0;
+				})[0]),
+				appx = path.join(bundle, fs.readdirSync(bundle).filter(function (f) {
+					return f.indexOf('.appx') >= 0 && f.indexOf('_scale') === -1;
+				})[0]),
+				outputDir = builder.outputDir,
+				dest = path.join(outputDir, path.basename(appx));
 
-			logger.info(__('Signing app'));
+			if (outputDir && outputDir != path.dirname(appx)) {
+				fs.existsSync(outputDir) || wrench.mkdirSyncRecursive(outputDir);
+				fs.existsSync(dest) && fs.unlinkSync(dest);
+				appc.fs.copyFileSync(appx, dest, {logger: logger.debug});
+			}
 
-			appc.subprocess.run(builder.windowsInfo.windowsphone[builder.wpsdk].xapSignTool, [
-			], function () {
-				var signedApp = path.join(builder.buildTargetProductDir, 'someapp.xap'),
-					outputDir = builder.outputDir;
-
-				if (outputDir && outputDir != path.dirname(signedApp)) {
-					fs.existsSync(outputDir) || wrench.mkdirSyncRecursive(outputDir);
-					dest = path.join(outputDir, path.basename(dest));
-					fs.existsSync(dest) && fs.unlinkSync(dest);
-					appc.fs.copyFileSync(builder.apkFile, dest, { logger: logger.debug });
-				}
-
-				logger.info(__('Packaging complete'));
-				logger.info(__('Package location: %s', dest.cyan));
-
-				finished();
-			});
-			*/
-
-			finished(); // temp
+			logger.info(__('Packaging complete'));
+			logger.info(__('Package location: %s', dest.cyan));
 		}
 	});
 
--- a/cli/hooks/ws-package.js
+++ b/cli/hooks/ws-package.js
@@ -1,0 +1,173 @@
+/**
+ * Signs an app and copies the result to the specified output directory.
+ *
+ * @copyright
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ *
+ * @license
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+const
+	appc = require('node-appc'),
+	fs = require('fs'),
+	path = require('path'),
+	wrench = require('wrench'),
+	moment = require('moment'),
+	async = require('async'),
+	__ = appc.i18n(__dirname).__;
+
+exports.cliVersion = '>=3.2';
+
+exports.init = function (logger, config, cli) {
+
+	cli.on('build.pre.compile', {
+		priority: 10000,
+		post: function (builder, finished) {
+			if (builder.buildOnly || builder.target !== 'dist-winstore') {
+				return finished();
+			}
+
+			var projectDir = builder.projectDir,
+				publisherName = builder.tiapp.properties['ti.windows.publishername'],
+				wSDK = builder.windowsInfo.windows[builder.wpsdk],
+				wsCert = builder.wsCert;
+
+			if (!publisherName || !publisherName.value) {
+				return next(new Error(__(
+					'ti.windows.publishername is a required tiapp.xml string property for publishing!' +
+					'\nFor example:' +
+					'\n<property name="ti.windows.publishername" type="string">CN=Appcelerator, Inc.</property>'
+				)));
+			}
+			publisherName = publisherName.value;
+
+			if (wsCert) {
+				builder.certificatePath = wsCert;
+				calculateThumbprint(wsCert, function (err, thumbprint) {
+					builder.certificateThumbprint = thumbprint;
+					finished();
+				});
+
+				return;
+			}
+
+			var expirationDate = moment().add(1, 'year').format('MM/DD/YYYY'),
+				pvk = path.join(projectDir, 'generated.pvk'),
+				cer = path.join(projectDir, 'generated.cer'),
+				pfx = path.join(projectDir, 'generated.pfx');
+
+			async.series([
+				function checkCertificates(next) {
+					if (fs.existsSync(pvk) && fs.existsSync(cer)) {
+						logger.info(__('Using existing generated pvk and cer'));
+						return next();
+					}
+
+					logger.info('');
+					logger.info(__('Creating a certificate'));
+					logger.info(__('Please follow the prompts'));
+					var args = [
+						'/n', publisherName, '/r', '/h', '0', '/eku', '1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13',
+						'/e', expirationDate, '/sv', pvk, cer
+					];
+					appc.subprocess.run(wSDK.makeCert.x86, args, function (code, out, err) {
+						return !err ? next() : next(err || new Error('Certificate generation failed'));
+					});
+				},
+				function checkPFX(next) {
+					logger.info(__('Creating a PFX'));
+					var args = [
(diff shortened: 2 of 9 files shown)


cli/hooks/wp-package.js
```diff
@@ -23,33 +24,102 @@ exports.init = function (logger, config, cli) {
...
+			publisherName && (publisherName = publisherName.value);
 
-				if (outputDir && outputDir != path.dirname(signedApp)) {
-					fs.existsSync(outputDir) || wrench.mkdirSyncRecursive(outputDir);
-					dest = path.join(outputDir, path.basename(dest));
-					fs.existsSync(dest) && fs.unlinkSync(dest);
-					appc.fs.copyFileSync(builder.apkFile, dest, { logger: logger.debug });
-				}
+			// TODO: Should we let them specify different names for the signing resources?
+			var pvk = path.join(projectDir, 'generated.pvk'),
```
sgtcoolguy: I think we should do a search in their project and if there's a pfx file with any name, we assume that's what they want to use and they already have a cert and private key, and the cert is installed. That's effectively what Windows does, it looks for any pfx file in the appx bundle during build.

cli/hooks/wp-package.js
```diff
@@ -23,33 +24,102 @@ exports.init = function (logger, config, cli) {
...
+						return next();
+					}
+
+					logger.info(__('Creating a certificate'));
+					logger.info(__('Please follow the prompts'));
+					var args = [
+						'/n', publisherName, '/r', '/h', '0', '/eku', '1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13',
+						'/e', expirationDate, '/sv', pvk, cer
+					];
+					appc.subprocess.run(wSDK.makeCert.x86, args, function (code, out, err) {
```
sgtcoolguy: Do we just make the cert, or do we also install it into the system for the user? From what I recall we need to install it for windows store apps before we can run them, and when you install it, it basically moves the file out, so the original cert file is gone.

cli/hooks/wp-package.js
```diff
@@ -23,33 +24,102 @@ exports.init = function (logger, config, cli) {
 	cli.on('build.post.compile', {
 		priority: 10000,
 		post: function (builder, finished) {
-			if (builder.buildOnly || builder.target !== 'dist-phonestore') return finished();
+			if (builder.buildOnly || builder.target !== 'dist-phonestore') {
```
sgtcoolguy: My understanding is that phone apps don't ever get signed by the user/developer, and that MS signs them when you submit them to the store via the dev center.

There's no mention of certs or signing anywhere for the phone apps, that I could find:
https://msdn.microsoft.com/library/windows/apps/jj206724.aspx
https://msdn.microsoft.com/en-us/library/windows/apps/br230835.aspx

cli/hooks/ws-package.js
```diff
@@ -0,0 +1,54 @@
...
+exports.cliVersion = '>=3.2';
+
+exports.init = function (logger, config, cli) {
+
+	cli.on('build.post.compile', {
+		priority: 10000,
+		post: function (builder, finished) {
+			if (builder.buildOnly || builder.target !== 'dist-windowsstore') return finished();
+
+			logger.info(__('Signing app'));
```
sgtcoolguy: Now for Windows store apps, there _does_ appear to be a requirement to sign them to even test them out locally on your machine. But again, I don't really see any signing stuff for submission to the store.

So technically all the signing code probably should only be done in the ws-run hook. As far as "packaging" goes for phone and store apps, it looks like you just do it all through MS' dev center website.

dawsontoth: Well I made this more complicated than it needed to be... working on revisions.

dawsontoth:
@sgtcoolguy I'm a bit stumped. If you don't specify a cert, I have it generate one for you. If you don't specify a password, everything works well. But if you do, then it becomes apparent that `certutil -addstore trusteduser your.pfx` isn't adding it to where visual studio wants it. You'll see a build error like the below.

`105: Cannot import the key file 'C:\Users\Dawson\Code\winning\generated.pfx'. The key file may be password protected. To correct this, try to import the certificate manually into the current user's personal certificate store.`

If I open the generated project in Visual Studio, then do `Package.appxmanifest > Packaging > Choose Certificate... > Configure Certificate... > Select from File...` it will prompt for the password and add it successfully, after which builds will work...

sgtcoolguy:
@dawsontoth I believe that Windows will actually generate some powershell script to handle acquiring the developer license, install the cert, etc - and it may give details on how to install a cert with a password by looking through it, I'm not sure. (https://msdn.microsoft.com/en-us/library/windows/apps/dn832613.aspx )

From what I can tell from Googling around you can import a cert, or you can import a pfx with a password using certutil. So in this case, we could use that command to import the cert from the pfx with the pfx password: 
```
certutil -p PFXPassword -importPFX your.pfx
```
From https://social.technet.microsoft.com/Forums/systemcenter/en-US/aca77b14-b598-4145-8658-a2e07140382b/how-to-use-command-prompt-to-import-pfx-certificate-to-workgroup-computer?forum=configmgribcm

I also found a random powershell script someone had written to try and solve the problem of installing a cert file with a password: http://poshcode.org/1937

dawsontoth:
Well, that didn't work either... but I did discover that if I open the pfx file(instead of the cert) and follow the default prompts to install it to the system, it works. So I added some scripting to launch the proper UI and tell the user to follow the prompts. If they already have a PFX, then we won't do this step. Or, if they don't specify a password when creating their PFX, then we'll skip it too.

I think it's all working now.

dawsontoth:
Rebased and squashed.

infosia:
Merged latest master and pushed an attempt to fix issue at test.js.

sgtcoolguy:
Some notes:
- We need to also do the cert stuff for ws-local, not just distribution. When I tried `ti build -p windows -T ws-local -G 00000000-0000-1000-8000-000000000000` it failed during MSBuild portion because teh CLI never did any of the cert./key stuff:
```
[INFO]  C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v12.0\AppxPackage\Microsoft.AppXPackage.Targets(2247,9): error APPX0101: A signing key is required in order to package this project. Please specify a PackageCertificateKeyFile or PackageCertificateThumbprint value in the project file. [C:\Users\Christopher\AppData\Local\Temp\mocha\ws.x86\Mocha.vcxproj]
```
- When trying to do a `ti build -p windows -T dist-winstore -G 00000000-0000-1000-8000-000000000000` it prompted me for pfx location, output location, brought up the cert import stuff, but once that native cert/key import was done, the CLi never moved on. It just sat at:
```
[INFO]  Creating a certificate
[INFO]  Please follow the prompts
[INFO]  Creating a PFX
[INFO]
[INFO]  Importing PFX
[INFO]  Please follow the prompts
[INFO]  (The default options will suffice)
^CTerminate batch job (Y/N)? Y
```
- Right after that failure, I retried. Now when it prompted about the pfx, it picked up the existing generated.pfx as default. I kept that default, and then the build again hung, now at:
```
[INFO]  Deploy type: production
[INFO]  Debugging disabled
[INFO]  Profiler disabled
[INFO]  Forcing rebuild: C:\Users\CHRIST~1\AppData\Local\Temp\mocha\build-manifest.json does not exist
```

After looking into the code and adding some logging, it appears that it has an empty path to the cert that it passes along to calculateThumbprint in ws-package.js.

sgtcoolguy:
@dawsontoth Another issue - running MakeCert is failing for me. After spending all morning tracking it down, it's because the makecert.exe you're finding use different arg styles. They use -e/-r, etc rather than switches (/e, /r).

The makecert.exe in the Windows Kits seem ok, but we can't specify the full certificate subject name AND do all the other options we need - bascially it gives you some small basic set of options, or some extended options which allow you to control the common name, but not the full cert subject name.

```
C:\Users\Christopher\titanium_mobile_windows\Tools\Scripts\build\mocha>C:\makec
rt.exe -!
Usage: MakeCert [ basic|extended options] [outputCertificateFile]
Extended Options
 -tbs <file>         Certificate or CRL file to be signed
 -sc  <file>         Subject's certificate file
 -sv  <pvkFile>      Subject's PVK file; To be created if not present
 -ic  <file>         Issuer's certificate file
 -ik  <keyName>      Issuer's key container name
 -iv  <pvkFile>      Issuer's PVK file
 -is  <store>        Issuer's certificate store name.
 -ir  <location>     Issuer's certificate store location
                        <CurrentUser|LocalMachine>.  Default to 'CurrentUser'
 -in  <name>         Issuer's certificate common name.(eg: Fred Dews)
 -a   <algorithm>    The signature's digest algorithm.
                        <md5|sha1|sha256|sha384|sha512>.  Default to 'sha1'
 -ip  <provider>     Issuer's CryptoAPI provider's name
 -iy  <type>         Issuer's CryptoAPI provider's type
 -sp  <provider>     Subject's CryptoAPI provider's name
 -sy  <type>         Subject's CryptoAPI provider's type
 -iky <keytype>      Issuer key type
                        <signature|exchange|<integer>>.
 -sky <keytype>      Subject key type
                        <signature|exchange|<integer>>.
 -l   <link>         Link to the policy information (such as a URL)
 -cy  <certType>     Certificate types
                        <end|authority>
 -b   <mm/dd/yyyy>   Start of the validity period; default to now.
 -m   <number>       The number of months for the cert validity period
 -e   <mm/dd/yyyy>   End of validity period; defaults to 2039
 -h   <number>       Max height of the tree below this cert
 -len <number>       Generated Key Length (Bits)
                        Default to '2048' for 'RSA' and '512' for 'DSS'
 -r                  Create a self signed certificate
 -nscp               Include Netscape client auth extension
 -crl                Generate a CRL instead of a certificate
 -eku <oid[<,oid>]>  Comma separated enhanced key usage OIDs
 -?                  Return a list of basic options
 -!                  Return a list of extended options

C:\Users\Christopher\titanium_mobile_windows\Tools\Scripts\build\mocha>C:\makec
rt.exe -n "CN=Appcelerator, Inc." -r -h 0 -eku 1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.31
.10.3.13 -e 01/01/2020 -sv generated.pvk generated.cer
Error: CryptCertStrToNameW failed => 0x80092023 (-2146885597)
Failed

C:\Users\Christopher\titanium_mobile_windows\Tools\Scripts\build\mocha>
```

I believe the docs you saw and the version of MakeCert invoked by the powershell script in windowslib are a different variant than the ones you're picking up.

It'd be good if we had all this logic for generating a cert, creating a pfx from pvk and cert were contained to windowslib itself. We already have a cert.js file for creating a cert and the pfx file. If we need more options - like setting a password, or breaking apart the cert generation from the pfx generation - then it should be folded into changes to windowslib and we should just invoke what we need from that library.

It looks to me like we need to break apart cert generation from pfx generation there, and maybe offer more options for cert generation so we can specify a password, and so we don't have to import to the truststore immediately as an admin during the generation.

sgtcoolguy:
Looks like the cryptic failure I'm seeing when using the -flags over the /switches is related to the usage of a comma in the common name (in the publisher name). We'll need to sanitize that value as well, it appears.
```
C:\Users\Christopher\titanium_mobile_windows\Tools\Scripts\build\mocha>C:\makece
rt.exe -n "CN=Appcelerator, Inc." -r -h 0 -eku 1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311
.10.3.13 -e 01/01/2020 -sv generated.pvk generated.cer
Error: CryptCertStrToNameW failed => 0x80092023 (-2146885597)
Failed
```

```
C:\Users\Christopher\titanium_mobile_windows\Tools\Scripts\build\mocha>C:\makece
rt.exe -n "CN=Appcelerator Inc." -r -h 0 -eku 1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.
10.3.13 -e 01/01/2020 -sv generated.pvk generated.cer
Succeeded
```

sgtcoolguy:
@dawsontoth  I also forgot to mention I rebased this branch on master, fixed a merge conflict and pushed a commit, so you'll need to delete your branch locally and re-create from the remote branch.

sgtcoolguy:
@dawsontoth I've got things working now locally. I'll merge this over once it builds. I need to make some changes to windowslib too...

sgtcoolguy:
Requires fixes contained in appcelerator/windowslib#18

sgtcoolguy:
@dawsontoth the -flag versus /switch thing was a red herring: you can use either variety. The issue I was running into was a bad subjectName (comma in company/common name) plus a series of other issues that were causing it not to run for me.

PR #263, opened 2015-06-09
Implement module with namespace

infosia:
Add support for namespace for modules such as `com.example.test1`.

```bash
$  ti create -t module -p windows --id com.example.test2 --name com.example.test2
```


sgtcoolguy:
test this please.

PR #277, opened 2015-06-11
[TIMOB-18999] Use actual PPI when populating layout properties

garymathews:
- Use actual PPI instead of ```1```
- This allows ```cm, in and dp``` to be calculated correctly

###### EXAMPLE
```Javascript
var win = Titanium.UI.createWindow({
    backgroundColor: '#fff'
});
 
var view = Titanium.UI.createView({
    backgroundColor: 'red',
    width: '1in',
    height: '10dp'
});
win.add(view);
 
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-18999)

### DIFF
--- a/Examples/NMocha/src/Assets/ti.ui.layout.test.js
+++ b/Examples/NMocha/src/Assets/ti.ui.layout.test.js
@@ -755,8 +755,7 @@ describe("Titanium.UI.Layout", function () {
     });
     
     // Functional Test #1001 #1002 #1003 #1004 #1005 #1006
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
@garymathews Any chance we could write up some simple mocha tests for this? I assume we'll always have the same emulator running by default so can rely on it's values remaining constant for now?

garymathews:
There was already mocha tests for _unitMeasurements_, I've enabled them.

PR #278, opened 2015-06-12
Provide ability to set capabilities on appx.manifest in tiapp.xml

infosia:
[TIMOB-18995](https://jira.appcelerator.org/browse/TIMOB-18995)

Provide ability to set capabilities on appx.manifest in tiapp.xml. Not only `Capabilities` but also `Properties`, `﻿Prerequisites` and `Resources` are supported.

```xml
<windows>
	<manifest>
		<Capabilities>
		</Capabilities>
	</manifest>
</windows>
```

By providing `target` attribute, you can setup platform-specific properties. Entries with no target means common properties across `phone` and `store`.

Updated: See example at https://github.com/appcelerator/titanium_mobile_windows/pull/278#issuecomment-111519660

In this case this ends up with `location` and `internetClient` capabilities set for `phone` app.

```xml
<Capabilities>
	<Capability Name="location" />
</Capabilities>
```

And `optical` and `internetClient` capabilities set for `store` app.

```xml
<Capabilities>
	<Capability Name="optical" />
</Capabilities>
```

### DIFF
--- a/cli/commands/_build.js
+++ b/cli/commands/_build.js
@@ -1786,6 +1786,8 @@ WindowsBuilder.prototype.readTiAppManifest = function readTiAppManifest() {
  */
 WindowsBuilder.prototype.generateAppxManifest = function generateAppxManifest(next) {
 
+	this.tiapp.windows = this.tiapp.windows || {};
+
 	this.readTiAppManifest();
 
 	if (!this.tiapp.windows.manifests) {
--- a/cli/commands/_build.js
+++ b/cli/commands/_build.js
@@ -1771,13 +1771,17 @@ WindowsBuilder.prototype.readTiAppManifest = function readTiAppManifest() {
  *
  *  Example:
  *	<windows>
- *		<manifest>
- *			<Capabilities target="phone">
+ *		<manifest target="phone">
+ *			<Capabilities>
  *				<Capability Name="location" />
  *			</Capabilities>
- *			<Capabilities target="store">
+ *		</manifest>
+ *		<manifest target="store">
+ *			<Capabilities>
  *				<Capability Name="optical" />
  *			</Capabilities>
+ *		</manifest>
+ *		<manifest>
  *			<Capabilities>
  *				<Capability Name="internetClient" />
  *			</Capabilities>
--- a/cli/commands/_build.js
+++ b/cli/commands/_build.js
@@ -1791,16 +1791,17 @@ WindowsBuilder.prototype.readTiAppManifest = function readTiAppManifest() {
 WindowsBuilder.prototype.generateAppxManifest = function generateAppxManifest(next) {
 
 	this.tiapp.windows = this.tiapp.windows || {};
-
 	this.readTiAppManifest();
-
-	if (!this.tiapp.windows.manifests) {
-		next();
-		return;
-	}
 
 	var xprops = {phone:{}, store:{}},
 		domParser = new DOMParser();
+
+	if (!this.tiapp.windows.manifests) {
+		this.generateAppxManifestForPlatform("store", xprops.store);
+		this.generateAppxManifestForPlatform("phone", xprops.phone);
+		next();
+		return;
+	}
 
 	// Construct manifest properties
 	for (var i = 0; i < this.tiapp.windows.manifests.length; i++) {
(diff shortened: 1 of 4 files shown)


infosia:
deployment target should be set at `manifest` attribute. Updated:

```xml
<windows>
	<manifest target="phone">
		<Capabilities>
			<Capability Name="location" />
		</Capabilities>
	</manifest>
	<manifest target="store">
		<Capabilities>
			<Capability Name="optical" />
		</Capabilities>
	</manifest>
	<manifest>
		<Capabilities>
		</Capabilities>
	</manifest>
</windows>
```

sgtcoolguy:
@infosia Some capabilities are declared with <Capability> tags and some are with <DeviceCapability>, I assume this works for both? It looks like it should...

infosia:
@sgtcoolguy Yes it supports anything as long as it has XML form, which will be added as it is.

```
  <Capabilities>
     <!-- YOU CAN ADD ANYTHING HERE AS LONG AS IT HAS XML FORM -->
  </Capabilities>
```

cli/commands/_build.js
```diff
@@ -1709,6 +1710,136 @@ WindowsBuilder.prototype.generateCmakeList = function generateCmakeList(next) {
...
+ * Write appxmanifest.in according to manifest properties
+ */
+WindowsBuilder.prototype.generateAppxManifestForPlatform = function generateAppxManifestForPlatform(target, properties) {
+	var template = fs.readFileSync(path.resolve(this.platformPath, 
+		'templates', 'build', 'Package.'+target+'.appxmanifest.in.ejs'), 'utf8'),
+		dest = path.join(this.buildDir, 'Package.'+target+'.appxmanifest.in');
+
+	// Supported properties
+	properties.Properties    = properties.Properties    || [];
+	properties.Capabilities  = properties.Capabilities  || [];
```
sgtcoolguy: I think we may want to retain a default of assuming "internetClient" as sole capability if they specify none? Otherwise some basic stuff just won't work, like our analytics. There actually may be some others we'd want to always set on by default.

I don't know the right thing to do here, obviously we'd like the user to be able to specify not using internet in case they had a solely local app - but for 99% of the apps I'd think we'd just want to assume some of the "basic" capabilities our APIs touch - and I'm not sure having to update the tiapp.xml after project creation/import to do so is the right way to go.

Any ideas @ingo ?

sgtcoolguy: We could update the default project template we use, or add some windows project create hook to add it by default - but then we'd also likely want Studio/CLI to do the same on import of old projects too?

dawsontoth:
Could we add it in a build step, @sgtcoolguy? When we generate the project from the tiapp.xml et al., we could make sure it has that capability by default.

sgtcoolguy:
@dawsontoth Yeah, that's what I was suggesting in my original comment, but then the user would have no way to explicitly say "No I _don't want to use an internet connection at all". Not sure how likely that is...

dawsontoth:
Hmmm...

```
if (capabilities section === undefined) {
    use internet capabilities and whatever other defaults we deem appropriate
}
else {
    use what they have specified (even if it's an empty tag, thus excluding internet capabiltiies)
}
```

dawsontoth:
Plus, we warn them at build time if they've excluded `internetClient` (probably on accident) that analytics and network requests won't work.

infosia:
@sgtcoolguy @dawsontoth sounds good, I'll implement default capability like below:

```
if (capabilities section === undefined) {
    use internet capabilities and whatever other defaults we deem appropriate
} else {
    use what they have specified (even if it's an empty tag, thus excluding internet capabiltiies)
}
```

When user doesn't specify any `Capabilities` in `tiapp.xml`, it assumes `internetClient`.

```xml
<windows>
    <manifest>
        <!-- "internetClient" capability will be added by default -->
    </manifest>
</windows>
```

When user want  to exclude `internetClient`, place empty `Capabilities`.

```xml
<windows>
    <manifest>
        <Capabilities>
        </Capabilities>
    </manifest>
</windows>
```

sgtcoolguy:
👍🏻

dawsontoth:
**CR** and **FR** passed.

PR #280, opened 2015-06-14
[TIMOB-19012] Fix: TabGroup shows two Window at once

infosia:
Fix for [TIMOB-19012](https://jira.appcelerator.org/browse/TIMOB-19012).


dawsontoth:
OK verified this! Looks great. *CR* and *FR* passed!

PR #283, opened 2015-06-15
Fix: assertion should not crash the app but show errrors

infosia:
[TIMOB-19019](https://jira.appcelerator.org/browse/TIMOB-19019)

```javascript

  title: 'PUSH'
});
    // On Debug build this should cause assertion error,
    // because it is guarded by TITANIUM_ASSERT(argument.IsBoolean())
    // On Release build this will go through, as runtime tries to convert it to boolean forcibly
    Ti.Platform.batteryMonitoring = 'this is not a boolean';
});
win.add(button);
win.open();
```

### DIFF
--- a/Source/TitaniumKit/include/Titanium/Module.hpp
+++ b/Source/TitaniumKit/include/Titanium/Module.hpp
@@ -100,6 +100,8 @@ namespace Titanium
 		virtual void fireEvent(const std::string& name) TITANIUM_NOEXCEPT final;
 		virtual void fireEvent(const std::string& name, const JSObject& event) TITANIUM_NOEXCEPT final;
 
+		virtual void showRedScreenOfDeath(const std::string& message) TITANIUM_NOEXCEPT final;
+
 		Module(const JSContext&) TITANIUM_NOEXCEPT;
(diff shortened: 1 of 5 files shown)


dawsontoth:
@infosia I'm not seeing any error screens. I'm pretty sure I built it right (I did the same stuff as I have done on the other PRs). Before pulling the changes in to my local workspace, I tried running the above code. It doesn't crash the app for me. If I place log statements before and after, I can see both. So maybe I'm running in release mode somehow?

Any guidance you can provide on testing this would be helpful.

I'm running with `ti build -p windows -C 8-1-1 -G 00000000-0000-1000-8000-000000000000` by the way.

garymathews:
Yes, you're in release mode. I tested it in Debug and I receive a red screen error message.

dawsontoth:
I'm ok merging this if you are, Gary. CR looks good to me, and your FR looks good to you. Any comments?

infosia:
Yes Gary is right. Would you merge this if this looks good to you @garymathews?

dawsontoth:
Separate question, then -- should my `ti build` above be running in release, or would debug be more appropriate? (If you ran an iOS app in the simulator via `ti build -p ios` and hit an error, it would show you the red screen of death.)

PR #284, opened 2015-06-15
Fix TIMOB-19022 Windows: Geolocation never returning data for getCurrentPosition

sgtcoolguy:
We never set up our Windows implementation to "override" the TitaniumKit Geolocation module.

### DIFF
--- a/Source/Titanium/src/TitaniumWindows.cpp
+++ b/Source/Titanium/src/TitaniumWindows.cpp
@@ -24,6 +24,7 @@
 #include "TitaniumWindows/File.hpp"
 #include "TitaniumWindows/HTTPClient.hpp"
 #include "TitaniumWindows/Media.hpp"
+#include "TitaniumWindows/Geolocation.hpp"
 
 #include <Windows.h>
 #include <collection.h>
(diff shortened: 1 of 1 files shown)


Source/Titanium/src/TitaniumWindows.cpp
```diff
@@ -58,6 +59,7 @@ namespace TitaniumWindows
 		                                                            .AppObject(js_context__.CreateObject(JSExport<TitaniumWindows::AppModule>::Class()))
 		                                                            .PlatformObject(js_context__.CreateObject(JSExport<TitaniumWindows::Platform>::Class()))
 		                                                            .GestureObject(js_context__.CreateObject(JSExport<TitaniumWindows::Gesture>::Class()))
+																	.GeolocationObject(js_context__.CreateObject(JSExport<TitaniumWindows::Geolocation>::Class()))
```
dawsontoth: Spaces vs tabs? Which should we use?

sgtcoolguy: I think for these lines here we're doing spaces, but typically we do tabs for indent levels?

dawsontoth:
Looks good. **CR** and **FR** passed. Merging...

PR #285, opened 2015-06-15
Fix TIMOB-19023 Windows: HTTPClient GET is failing

sgtcoolguy:
The initial error was because we didn't "skip" setting up a timeout timer when timeout was 0 (or less). Since timeout defaults to 0, we always cancelled the request immediately.

I also took the time to clean up the code so we didn't have three separate "send" implementations. Now they all just gather up the content object they need to send and call the shared send implementation.


dawsontoth:
*CR* looks good. *FR* in process.

dawsontoth:
**CR** and **FR** passed. Merging...

PR #288, opened 2015-06-16
Fix: Geolocation module should instantiate Geolocator on-demand

infosia:
Quick fix for [TIMOB-19028](https://jira.appcelerator.org/browse/TIMOB-19028). Developer should take care of the timing when first use of Geolocation, which should be done on UI thread according to the [guideline](https://msdn.microsoft.com/en-us/library/windows/apps/hh465148.aspx).

```javascript
Ti.Geolocation.addEventListener('location', ...); // should be done on the UI thread!
```

### DIFF
--- a/Source/Sensors/src/Geolocation.cpp
+++ b/Source/Sensors/src/Geolocation.cpp
@@ -24,10 +24,6 @@ namespace TitaniumWindows
 	Geolocation::Geolocation(const JSContext& js_context) TITANIUM_NOEXCEPT
 		: Titanium::GeolocationModule(js_context)
 	{
-		geolocator_ = ref new Geolocator();
-		geolocator_->MovementThreshold = 1;
-		geolocator_->ReportInterval = 0;
-		geolocator_->DesiredAccuracy = PositionAccuracy::High;
 	}
 
 	void Geolocation::JSExportInitialize() {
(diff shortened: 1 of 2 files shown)


Source/Sensors/src/Geolocation.cpp
```diff
@@ -118,12 +128,19 @@ namespace TitaniumWindows
...
 	{
+		ensureLoadGeolocator();
+
 		geolocator_->MovementThreshold = distance;
 		Titanium::GeolocationModule::set_distanceFilter(distance);
 	}
 
 	bool Geolocation::get_locationServicesEnabled() const TITANIUM_NOEXCEPT
 	{
+		// we can't load GeoLocator here because this is const function. Just return false when it's not ready.
```
dawsontoth: I think people quite often do code like

```
if (Ti.Geolocation.locationServicesEnabled) {
    Ti.Geolocation.addEventListener('location', ...);
}
```
If the geolocator is null, can we still check to see if it can be enabled without setting the ref?

sgtcoolguy: I think we'd have to lazily instantiate the geolocator for the locationServicesEnabled check too - _if_ it's actually required.

The property is actually supposed to tell us if the app is registered for location services, and our current impl is wrong: https://jira.appcelerator.org/browse/TIMOB-19021

It comes down to knowing if the app has the "location" DeviceCapability set, and ideally also checking some other settings on the device. Here's the docs say:

> The user sets the privacy of their location data with the location privacy settings in the Settings app. Your app can access the user's location only when:
> Location for this device... is turned on (not applicable to Windows 10 Mobile Insider Preview)
> The location services setting, Location, is turned on
Under Choose apps that can use your location, your app is set to on

dawsontoth: I suppose we can take care of this in TIMOB-19021, then!

PR #289, opened 2015-06-16
[TIMOB-19029] Report errors on JS parse error

infosia:
Fix for [TIMOB-19029](https://jira.appcelerator.org/browse/TIMOB-19029).

### DIFF
--- a/cli/commands/_build.js
+++ b/cli/commands/_build.js
@@ -1186,6 +1186,36 @@ WindowsBuilder.prototype.copyResources = function copyResources(next) {
 		}
 	}
 
+	/*
+	 * Check JS syntax and report errors. Mostly copied from jsanalyze
+	 */
+	function reportJSErrors(filename, contents, ex) {
+		var errmsg = [ __('Failed to parse %s', filename) ];
+		if (ex.line) {
+			errmsg.push(__('%s [line %s, column %s]', ex.message, ex.line, ex.col));
+		} else {
+			errmsg.push(ex.message);
+		}
+		try {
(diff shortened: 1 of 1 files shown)


cli/commands/_build.js
```diff
@@ -1417,8 +1447,15 @@ WindowsBuilder.prototype.copyResources = function copyResources(next) {
...
 
 				// Look for native requires here
-				// FIXME Avoid parsing the AST twice! We do it below using jsanalyze for non html referenced JS files!
-				var toplevel = UglifyJS.parse(fs.readFileSync(from).toString());
+				var fromContent = fs.readFileSync(from, {encoding:'utf8'});
+				try {
+					// FIXME Avoid parsing the AST twice! We do it below using jsanalyze for non html referenced JS files!
+					var toplevel = UglifyJS.parse(fromContent);
+				} catch  (E) {
+					t_.logger.error(reportJSErrors(from, fromContent, E));
```
dawsontoth: Should this be fatal (and kill the build)? Or, should it kill the build at some point? When I run this, I see the error message, but then the emulator starts and the CLI just sits there until I terminate it.

sgtcoolguy: I don't _think_ it should be fatal. The parsing here is done to sniff for native requires. If we can't parse the JS for some reason that's not a great sign, but I think we can try and move on and assume no native APIs were used in the file with bad syntax.

dawsontoth: OK then.

dawsontoth:
**CR** and **FR** passed.

PR #293, opened 2015-06-17
Package all headers including TitaniumWindows

infosia:
Followup for [TIMOB-18992](https://jira.appcelerator.org/browse/TIMOB-18992).

Package all headers including `TitaniumWindows` for Titanium SDK build, because it may be used by module developers. For example `ti.map` may need access to constants in `TitaniumWindows::UI` as well as `Titanium::UI::View`.


sgtcoolguy:
I'm fine with this. I recall a long time back we had concerns about including headers before we open-sourced the repo. Might want to confirm with @ingo that we're OK with shipping them...

dawsontoth:
**CR** and **FR** good.

ingo:
I am fine with shipping headers.

PR #296, opened 2015-06-17
Parallelize Builds

dawsontoth:
Build for the various targets and architectures in parallel, and then copy over the various files in parallel. Builds take about 30 minutes for me using this, and they _appear_ to work as normal. (I say _appear_
 because I'm just poking around with an app to test.)

### DIFF
--- a/Tools/Scripts/build/build.js
+++ b/Tools/Scripts/build/build.js
@@ -11,23 +11,130 @@ var path = require('path'),
 	exec = require('child_process').exec,
 	spawn = require('child_process').spawn,
 	cmakeLocation = path.join(__dirname, '..', '..', '..', 'cli', 'vendor', 'cmake', 'bin', 'cmake.exe'),
-	MSBuildLocation = "C:/Program Files (x86)/MSBuild/12.0/Bin/MSBuild.exe";
+	MSBuildLocation = 'C:/Program Files (x86)/MSBuild/12.0/Bin/MSBuild.exe';
+
+// Now do the builds
+var rootDir = path.join(__dirname, '..', '..', '..'),
+	titaniumWindowsSrc = path.join(rootDir, 'Source', 'Titanium'),
+	buildRoot = path.join(rootDir, 'build'),
+	distRoot = path.join(rootDir, 'dist', 'windows'),
+	distLib = path.join(distRoot, 'lib'),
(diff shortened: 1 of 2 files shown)


dawsontoth:
Still testing to measure how long building them in series takes on my machine. I know it takes a while... but haven't quantified the speedup yet.

dawsontoth:
Also added a --quiet flag if you don't care about the output from cmake or msbuild.

dawsontoth:
Yeah, this reduced my build time from 50 minutes to 30 minutes.

sgtcoolguy:
@dawsontoth I had previously used the /m flag for msbuild to parallelize the actual compilation and ran into exactly what happened here - it saturated the build agent's CPU causing the java process that communicates with Jenkins master to get starved and the agent to be marked offline mid-build.

That's why the PR build failed...

dawsontoth:
aw :(

dawsontoth:
hold on a sec...

dawsontoth:
There, how about that Chris? So the build server can do its own thing, and when building locally, we can use all the processors.

Tools/Scripts/build/build.js
```diff
@@ -34,71 +162,42 @@ function runCMake(sourceDir, buildDir, buildType, platform, arch, callback) {
...
-	});
-
-	prc.on('close', function (code) {
-		if (code != 0) {
-			callback("Failed to run MSBuild");
-		} else {
-			callback();
-		}
-	});
+	var args = ['/p:Configuration=' + buildType];
```
garymathews: Could add ```/m``` here? https://msdn.microsoft.com/en-us/library/bb651793.aspx

sgtcoolguy:
I think the jenkins build agent is hosed right now. We'll need to reboot it before builds will work again, I'm betting

dawsontoth:
sorry if i did that :)

sgtcoolguy:
@dawsontoth Any reason you killed this PR? I'm ok with the idea to help speed up local builds. The last build failed due to an issue copying all the HAL headers over, much like the Windows doc PR. I'm not sure why that's suddenly happening...

dawsontoth:
Pulled in the fix for CI, hopefully it'll pass now.

dawsontoth:
test this please

sgtcoolguy:
test this please.

dawsontoth:
will have to rebase again

PR #300, opened 2015-06-18
Fix to actually set request headers before making the request.

sgtcoolguy:
We used to set HTTP request headers for GET only, then when I refactored for the fix recently even that got removed. This is an attempt to bring it back for all request methods.


dawsontoth:
I think I'm seeing the request `onload` fire twice. Not sure if it's pre-existing, or just with this PR...
```javascript
	backgroundColor: '#fff'
});
win.open();

	onload: function () {
		console.log('load event fired');
	},
	onerror: function () {
		console.log('errored');
	}
});
client.open('GET', 'http://www.google.com/');
client.send();
```

sgtcoolguy:
test this please.

PR #308, opened 2015-06-22
Fix TIMOB-19044 Windows: Accessing a sliders value always returns 0

sgtcoolguy:
The problem is that we need to override the getter for value in the Windows subclass to always return the value from the widget itself, rather than just parrot back the internal value field (which is only changed by the property setter, not from actually manipulating the widget).

https://jira.appcelerator.org/browse/TIMOB-19044


dawsontoth:
**CR** looks good. Reproduced the problem, now I'm mostly through waiting for my build to generate the updated SDK...

PR #309, opened 2015-06-22
ti.windows.publishername tweaks

sgtcoolguy:
We attempt to use the value if it's there in tiapp, otherwise we generate one with"CN=" + tiapp.publisher (and log a warning). If there's no publisher (display name) then we show an error message about needing the value and exit with error code 1.

https://jira.appcelerator.org/browse/TIMOB-19009
https://jira.appcelerator.org/browse/TIMOB-19025

### DIFF
--- a/Tools/Scripts/build/test.js
+++ b/Tools/Scripts/build/test.js
@@ -100,25 +100,6 @@ function generateWindowsProject(next) {
 	});
 }
 
-// Add required properties
-function addTiAppProperties(next) {
-	var tiapp_xml = path.join(__dirname, 'mocha', 'tiapp.xml'),
-		publishername = '<property name="ti.windows.publishername" type="string">CN=Dawson Toth</property>';
-
-	// Not so smart but this should work...
-	var content = [];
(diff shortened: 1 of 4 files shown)


Tools/Scripts/build/test.js
```diff
@@ -100,25 +100,6 @@ function generateWindowsProject(next) {
...
-	var tiapp_xml = path.join(__dirname, 'mocha', 'tiapp.xml'),
-		publishername = '<property name="ti.windows.publishername" type="string">CN=Dawson Toth</property>';
-
-	// Not so smart but this should work...
-	var content = [];
-	fs.readFileSync(tiapp_xml).toString().split(/\r?\n/).forEach(function(line){
-		content.push(line);
-		if (line.indexOf('<guid>') >= 0) {
-			content.push('\t'+publishername);
-			content.push('\t<windows><manifest><Capabilities><Capability Name=\"internetClient\" /></Capabilities></manifest></windows>');
```
infosia: I think we still need this to enable `internetClient` capability.

sgtcoolguy: We default to including that capability if none are set, so it just plain works now.

PR #310, opened 2015-06-22
Possible fix for TIMOB-19039 - Windows: Titanium.version does not return correct values

sgtcoolguy:
Stuff sdkVersion into _app_info_.json and read it at runtime to lazily init Ti.version


sgtcoolguy:
Merged manually.

PR #311, opened 2015-06-22
Global.String for Windows

infosia:
- Implement `Global.String` functions for Windows
  - `formatDate`, `formatTime`, `formatCurrency` and `formatDecimal`

This PR introduces new `Titaniium.String` namespace which is basically holds Module instance for `Global.String`. Global.String needs this because `String` is built-in class which can not contain `HAL` type, and on the other hand HAL needs to extract pointer of Global.String instance from JavaScript object in any way in order to make HAL binding work.


Source/Global/src/GlobalString.cpp
```diff
@@ -26,26 +27,43 @@ namespace TitaniumWindows
...
 	}
 
-	std::string GlobalString::formatDate(JSObject date, std::string format) TITANIUM_NOEXCEPT
+	std::string GlobalString::formatDate(JSObject dateObject, std::string format) TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_ERROR("GlobalString::formatDate: Unimplemented");
-		return static_cast<std::string>(static_cast<JSValue>(date));
+		using namespace Windows::Globalization::DateTimeFormatting;
+		DateTimeFormatter^ formatter;
+		if (format == "short") {
```
sgtcoolguy: do we need to handle medium? Do we validate the value is "short", "medium", or "long" up in TitaniumKit?

Source/Global/src/GlobalString.cpp
```diff
@@ -26,26 +27,43 @@ namespace TitaniumWindows
...
 	}
 
-	std::string GlobalString::formatTime(JSObject date, std::string format) TITANIUM_NOEXCEPT
+	std::string GlobalString::formatTime(JSObject dateObject, std::string format) TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_ERROR("GlobalString::formatTime: Unimplemented");
-		return static_cast<std::string>(static_cast<JSValue>(date));
+		using namespace Windows::Globalization::DateTimeFormatting;
+		DateTimeFormatter^ formatter;
+		if (format == "short") {
```
sgtcoolguy: same here - can we handle "medium", do we validate values up in TK?

Source/TitaniumKit/src/GlobalString.cpp
```diff
@@ -83,7 +95,8 @@ namespace Titanium
 		ENSURE_OBJECT_AT_INDEX(date, 0);
 		ENSURE_OPTIONAL_STRING_AT_INDEX(format, 1, "");
```
sgtcoolguy: Looks like default should be "short": https://github.com/appcelerator/titanium_mobile/blob/0ff8c6818ca7a1b1cd0da18c11c60bab00ccb423/iphone/Classes/KrollContext.m#L489

We should also verify the value is one of the valid ones: "short", "medium", "long" (and undocumented, but "full" seems to be an option too).

Source/TitaniumKit/src/GlobalString.cpp
```diff
@@ -100,7 +114,8 @@ namespace Titanium
...
```
sgtcoolguy: default is "short". We should validate value is one of "short", "medium", "long.

Source/Utility/include/TitaniumWindows/Utility.hpp
```diff
@@ -19,6 +19,7 @@
 #include <sstream>
 #include <iomanip>
 #include <cstdint>
+#include "HAL/HAL.hpp"
```
sgtcoolguy: Related to build times: We may want to move non-templated method defintions over to Utility.cpp and move includes over there for things like HAL/HAL.hpp - it should help improve our build times. I think for example in this case we could forward declare HAL::JSValue in the hpp when we declare the GetDateTime method, and only include in the cpp. Then anything that includes Utility.hpp should compile faster (since it won't need to include all of HAL's headers/etc).

Source/Global/src/GlobalString.cpp
infosia: Looks like Windows DateTime formatter only supports short and long template. [DateTimeFormatter class](https://msdn.microsoft.com/en-us/library/windows.globalization.datetimeformatting.datetimeformatter.aspx)

Source/Utility/include/TitaniumWindows/Utility.hpp
infosia: Fixed entire `TitaniumWindows_Utility` so that it property export its functions via DLL. It should be making compile faster and more efficient.

Source/Utility/include/TitaniumWindows/Utility.hpp
```diff
@@ -44,104 +51,52 @@ namespace TitaniumWindows
 		//
 		// Convert std::string into Platform::String^
 		//
-		static ::Platform::String^ ConvertString(const std::string& str)
-		{
-			return ref new ::Platform::String(std::wstring(str.begin(), str.end()).c_str());
-		}
+		TITANIUMWINDOWS_UTILITY_EXPORT::Platform::String^ ConvertString(const std::string& str);
```
sgtcoolguy: Looks like this isn't exporting the method properly now? I noticed some of these you have no space between the EXPORT and the return type, like this here, but not sure if that matters.

Source/Utility/include/TitaniumWindows/Utility.hpp
```diff
@@ -44,104 +51,52 @@ namespace TitaniumWindows
...
 
 		//
 		// Convert Platform::String^ into std::string
 		//
-		static std::string ConvertString(::Platform::String^ str)
-		{
-			return std::string(str->Begin(), str->End());
-		}
+		TITANIUMWINDOWS_UTILITY_EXPORT std::string ConvertString(::Platform::String^ str);
```
sgtcoolguy: RequireHook is failing to be able to link to this method now. Not sure why...

sgtcoolguy:
@infosia looks like there's a merge conflict too? Good to merge once conflict is fixed and builds pass...

PR #313, opened 2015-06-22
[TIMOB-19054] Ti.Filesystem.File.write() 'append' parameter should be optional

garymathews:
- Fix ```Ti.Filesystem.File.write()``` parameter validator

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19054)


sgtcoolguy:
Looks fine to me. Ideally we'd have a couple unit tests here to verify that not specifying the second param was equivalent to a value of false.

PR #316, opened 2015-06-23
HTTPClient GET on an incorrect URL throws onload instead of onerror

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19042

### DIFF
--- a/Examples/NMocha/src/Assets/ti.network.httpclient.test.js
+++ b/Examples/NMocha/src/Assets/ti.network.httpclient.test.js
@@ -45,6 +45,24 @@ describe("Titanium.Network.HTTPClient", function () {
         xhr.send();
     });
 
+    it("TIMOB-19042", function (finish) {
+        this.timeout(6e4);
+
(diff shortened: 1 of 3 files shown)


infosia:
test this please

yuchi:
A 404 is a perfectly good response, it must **not** fire `onerror`.

sgtcoolguy:
As per our other platforms - we should fire onerror if the code is 400-599. The bug here is that we should do so only if an onerror handler is registered, otherwise we fire onload. I need to fix this PR to check that.

- Chris

On Jun 24, 2015, at 8:00 AM, Pier Paolo Ramon <notifications@github.com<mailto:notifications@github.com>> wrote:


A 404 is a perfectly good response, it must not fire onerror.

-

infosia:
Looks good, once all tests passed. I believe build failure on Travis is already fixed by #320.

sgtcoolguy:
Merged manually.

yuchi:
I'm afraid the HTTPClient diverges too much from XmlHTTPRequests.

PR #317, opened 2015-06-23
[TIMOB-19048] Word wrap on labels does not work if width and height are not set

garymathews:
- Set default ```width``` and ```height``` to ```Ti.UI.FILL```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19048)


sgtcoolguy:
@garymathews This change has caused the Ti.Layout.fourPins unit test to fail.

sgtcoolguy:
Testing this locally right now... (I killed the Windows PR build which is why it shows failure)

sgtcoolguy:
Manually merged.

PR #321, opened 2015-06-24
Encryption - Ready to Go

dawsontoth:
ready to go


Source/Global/src/GlobalObject.cpp
```diff
@@ -90,18 +107,66 @@ namespace TitaniumWindows
 		Platform::String^ content;
 		bool hasError = false;
 		concurrency::event event;
-		concurrency::task<Windows::Storage::StorageFile^>(Windows::Storage::StorageFile::GetFileFromPathAsync(module_path)).then([&content, &hasError, &event](concurrency::task<Windows::Storage::StorageFile^> task) {
-			try {
-				concurrency::task<Platform::String^>(Windows::Storage::FileIO::ReadTextAsync(task.get(), Windows::Storage::Streams::UnicodeEncoding::Utf8)).then([&content, &hasError, &event](concurrency::task<Platform::String^> task) {
-					content = task.get();
+
+		if (seed__.compare("") == 0) {
```
sgtcoolguy: I think you're lazily initializing the value here when it's empty? I'd do:
seed__.empty()

Source/Titanium/include/TitaniumWindows/TitaniumWindows.hpp
```diff
@@ -24,6 +24,7 @@ namespace TitaniumWindows
 	{
 	public:
 		Application();
+		Application(::Platform::String^ seed);
```
sgtcoolguy: Instead of introducing a new constructor, I'd modify the one above to be:
Application(std::string seed = "");

Basically make it an optional parameter with a default of being empty. Then you wouldn't need to convert from Platform::String in TitaniumWindows.cpp constructor either.

Source/Global/src/GlobalObject.cpp
```diff
@@ -90,18 +107,66 @@ namespace TitaniumWindows
 		Platform::String^ content;
 		bool hasError = false;
 		concurrency::event event;
-		concurrency::task<Windows::Storage::StorageFile^>(Windows::Storage::StorageFile::GetFileFromPathAsync(module_path)).then([&content, &hasError, &event](concurrency::task<Windows::Storage::StorageFile^> task) {
-			try {
-				concurrency::task<Platform::String^>(Windows::Storage::FileIO::ReadTextAsync(task.get(), Windows::Storage::Streams::UnicodeEncoding::Utf8)).then([&content, &hasError, &event](concurrency::task<Platform::String^> task) {
-					content = task.get();
+
+		if (seed__.empty()) {
+			concurrency::task<StorageFile^>(StorageFile::GetFileFromPathAsync(module_path)).then([&content, &hasError, &event](concurrency::task<StorageFile^> task) {
```
sgtcoolguy: So it's not very obvious when looking at the code and docs, but basically async tasks like this can be written in two ways and they're handled slightly differently.

When you use the form that expects a callback arg of type task, that means you need to use get() to get the wrapped value, and that call may throw an exception if the operation failed. That's what you're using here.

If you're chaining a number of tasks using then, you can avoid unwrapping the task at each step and just handled the exceptions in the last one (if you don't need/care about handling at each step along the way).

``` c++
concurrency::task<StorageFile^>(StorageFile::GetFileFromPathAsync(module_path))
.then([&content, &hasError, &event](StorageFile^ file) {
  return FileIO::ReadTextAsync(file, Windows::Storage::Streams::UnicodeEncoding::Utf8);
}, concurrency::task_continuation_context::use_arbitrary())
.then([&content, &hasError, &event](concurrency::task<Platform::String^> task) {
  try {
    content = task.get();
  catch (Platform::COMException^ ex) {
    hasError = true;
  }
   event.set();
}, concurrency::task_continuation_context::use_arbitrary());
```

Additionally, in the example above what I'm showing is that if the return value inside a task is another async operation, it can be chained to the next task (the return FileIO part).

sgtcoolguy: I just hacked that example together, it may be horribly broken, but hopefully illustrates what I was saying.

https://msdn.microsoft.com/en-us/library/windows/apps/hh780559.aspx

Source/Global/src/GlobalObject.cpp
```diff
@@ -90,18 +109,71 @@ namespace TitaniumWindows
 		Platform::String^ content;
 		bool hasError = false;
 		concurrency::event event;
-		concurrency::task<Windows::Storage::StorageFile^>(Windows::Storage::StorageFile::GetFileFromPathAsync(module_path)).then([&content, &hasError, &event](concurrency::task<Windows::Storage::StorageFile^> task) {
-			try {
-				concurrency::task<Platform::String^>(Windows::Storage::FileIO::ReadTextAsync(task.get(), Windows::Storage::Streams::UnicodeEncoding::Utf8)).then([&content, &hasError, &event](concurrency::task<Platform::String^> task) {
-					content = task.get();
+
+		if (seed__.empty()) {
+			concurrency::task<StorageFile^>(StorageFile::GetFileFromPathAsync(module_path))
```
dawsontoth: Chris, is this what you meant when you described the task changes? I'm getting a compile error now:

C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\ppltasks.h(433): error C2338: incorrect parameter type for the callable object in 'then'; consider _ExpectedParameterType or task<_ExpectedParameterType> (see below) (C:\Users\Dawson\Code\titanium_mobile_windows\Source\Global\src\GlobalObject.cpp) [C:\Users\Dawson\Code\titanium_mobile_windows\build\phone\x86\Global\TitaniumWindows_Global.vcxproj]

Source/Titanium/src/Blob.cpp
```diff
@@ -29,8 +33,17 @@ namespace TitaniumWindows
 	{
 		width_ = 0;
 		height_ = 0;
-		data_ = TitaniumWindows::Utility::GetContentFromFile(file);
+
 		path_ = TitaniumWindows::Utility::ConvertString(file->Path);
+		if (boost::ends_with(path_, ".js") || boost::ends_with(path_, ".json")){
+			// TODO: We should refactor to get the IBuffer back directly, rather than converting the contents multiple times.
+			std::string contents = GlobalObject::readRequiredModule(path_);
```
sgtcoolguy: You need the instance of GlobalObject. I don't think this method is static.

dawsontoth: I extracted a static version of it.

PR #322, opened 2015-06-24
Attempt to clean up code for faster builds

sgtcoolguy:
- Massive cleanup of headers. 
- Break up TiBase into two halves TiBase for declaration macros, TiImpl for definition macros.
- Use forward declarations as much as possible

PR #327, opened 2015-06-25
Fix NG TitaniumWindows_Native includes

garymathews:
- Fix Missing ```#include```'s causing compilation errors

### DIFF
--- a/Source/TitaniumKit/include/Titanium/detail/TiBase.hpp
+++ b/Source/TitaniumKit/include/Titanium/detail/TiBase.hpp
@@ -51,6 +51,7 @@
 #endif
 
 #include "TitaniumKit_EXPORT.h"
+#include "Titanium/detail/TiImpl.hpp"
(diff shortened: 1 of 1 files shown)


Source/TitaniumKit/include/Titanium/detail/TiBase.hpp
```diff
@@ -51,6 +51,8 @@
 #endif
 
 #include "TitaniumKit_EXPORT.h"
+#include "Titanium/detail/TiLogger.hpp"
```
sgtcoolguy: TiImpl includes TiLogger, so technically this one isn't needed.

infosia:
I think that one of the point of #322 is to make sure `TiBase.hpp` doesn't contain macro implementation, to make the build faster. So I think what we need to update to make `NG` work now is files under `Examples/NG/Native`...I'm working on it and push PR for it soon.

infosia:
Closing, overridden by #329

PR #341, opened 2015-06-30
[TIMOB-19119] Fix LayoutEngine 'left, right' and 'top, bottom' scaling

garymathews:
- Fix ```left, right``` and ```top, bottom``` scaling for sized components such as ```TextField```

###### EXAMPLE
```Javascript
    backgroundColor: 'blue'
});
var viewA = Ti.UI.createTextField({
    hintText: 'SOME TEXT',
    top: '5dp',
    left: '10dp',
    
    // removing this line should allow the TextField to scale according to its contents
    // retaining this line should scale the TextField to the same size as the red view below
    right: '10dp',
    
    height: '40dp'
});
var viewB = Ti.UI.createView({
    backgroundColor: 'red',
    top: '60dp',
    left: '10dp',
    right: '10dp',
    height: '40dp'
});
win.add(viewA);
win.add(viewB);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19119)

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -686,8 +686,6 @@ namespace TitaniumWindows
 				rect.width = component->ActualWidth;
 			}
 
-			oldRect__ = Titanium::LayoutEngine::RectMake(rect.x, rect.y, rect.width, rect.height);
-
 			if (is_panel__) {
 				for (auto child : panel->Children) {
 					child->Visibility = Visibility::Collapsed;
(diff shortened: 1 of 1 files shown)


garymathews:
test this please

garymathews:
test this please

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -794,7 +804,7 @@ namespace TitaniumWindows
 				prop.value = "UI.FILL";
 			}
 
-			auto ppi = Windows::Graphics::Display::DisplayInformation::GetForCurrentView()->LogicalDpi;
+			auto ppi = Windows::Graphics::Display::DisplayInformation::GetForCurrentView()->LogicalDpi/2;
```
dawsontoth: I'm not sure this calculation is right. On my 8.1 emulator, everything looks half height now.

dawsontoth: (The other changes look good, though!)

sgtcoolguy:
@garymathews @dawsontoth can someone take a screenshot of pre- and post- this PR to see visually the difference? Maybe we can compare versus an equivalent Android/iOS app, and one of the builtin Windows Phone apps to see if this makes us look much more similar size-wise to the other platforms, or ends up making us look too different from Windows app "standards"?

dawsontoth:
Well, maybe I'm wrong about this. I'm only seeing it when we specify units with `dp`. I am on a MBP Retina and Windows is scaling down the windows. In the screenshots, the other example I ran from another PR didn't change when I ran with these changes. But you can see the changes when I ran the example above. Also note that the font size of the text field doesn't change.

withdpifix
without
withoutdpifix

sgtcoolguy:
I tried using the corporate directory and that looked the same. Looking at the code, it seems to only matter for units specified in in/dp/dip/cm.

I'm diving into this now to confirm what the right thing to do is here...

dawsontoth:
Yeah, I've been rolling it around in my mind, too.

sgtcoolguy:
I think I've got it, let me open a new PR...

garymathews:
Updated my PR

dawsontoth:
:+1:

PR #344, opened 2015-07-01
TIMOB-19127 Windows: Ti.Platform.createUUID() should not have curly brackets

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19127

This adds more testing of the createUUID value in our unit tests to validate that the UUID is 36 characters long, doesn't start with open brace or end with close brace and tests against a regexp for format.

I also remove the hacks around the issue in analytics.


dawsontoth:
CR :+1:

dawsontoth:
FR :+1:

PR #345, opened 2015-07-01
Simple tool to convert JS files into C++ header files all at once

infosia:
Very simple tool to convert JS files into C++ header files all at once. It is useful when you debug `*.js` files in TitaniumKit, to avoid running `cmake` every time you update the JS file.

How to Use:

```bash
$ # update JS file, for example Source/TitaniumKit/src/analytics.js
$ vi Source/TitaniumKit/src/analytics.js
$ node Tools/Scripts/generate_cpp_headerfile.js
$ # Source/TitaniumKit/include/Titanium/analytics_js.hpp is updated!
```

### DIFF
--- a/Tools/Scripts/generate_cpp_headerfile.js
+++ b/Tools/Scripts/generate_cpp_headerfile.js
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) 2014 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+/*
+ * Simple utility to convert JS files into C++ header files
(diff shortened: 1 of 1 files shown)


Tools/Scripts/generate_cpp_headerfile.js
```diff
@@ -0,0 +1,74 @@
...
+        if (position > 0) {
+            output.push(",");
+        }
+        output.push(data.readInt8(i));
+    }
+    output.push(",0"); // NULL termination
+    return output.join('').trim();
+}
+
+for (var i = 2; i < files_to_process.length; i++) {
```
garymathews: Is there a reason why we skip the first two?

infosia: Oops typo...during copy & paste. Fixed.

PR #347, opened 2015-07-01
Fix TIMOB-19128 Windows: Ti.Filesystem.File.createDirectory will fail if parent directory does not exist

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19128

```javascript
 
var btn = Ti.UI.createButton({
	title: 'Click'
});
 
btn.addEventListener('click', function (e) {
 
	var dir1 = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'dir1');
	dir1.createDirectory();
 
	console.info('Working fine');
 
	var dir2 = Ti.Filesystem.getFile(Ti.Filesystem.applicationDataDirectory, 'sub', 'dir2');
	dir2.createDirectory();
 
	console.info('Will have crashed on Windows');
});
 
win.add(btn);
win.open();
```

### DIFF
--- a/Examples/NMocha/src/Assets/ti.filesystem.test.js
+++ b/Examples/NMocha/src/Assets/ti.filesystem.test.js
@@ -198,8 +198,7 @@ describe('Titanium.Filesystem', function () {
         should(Ti.Filesystem.getFile).not.be.undefined;
         should(Ti.Filesystem.getFile).be.a.Function;
         var file = Ti.Filesystem.getFile('app.js');
-        should(file).not.be.null;
-        should(file).not.be.undefined;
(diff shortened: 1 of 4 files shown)


sgtcoolguy:
@dawsontoth looks like I broke the unit tests with this. I'll have to take a look tomorrow.

sgtcoolguy:
OK, fixed up the tests, added one to explicitly test the use case described in the ticket.

There was/is some funkiness with `should(xyz).not.be.null;` causing stack overflows, so I changed where we did:
```javascript
should(file).not.be.null;
should(file).not.be.undefined;
```

into
```javascript
should(file).be.ok; // not null, undefined
```

dawsontoth:
**CR** & **FR** :+1:

PR #348, opened 2015-07-01
TIMOB-19104 Windows: If generating ti.windows.publishername from tiapp publisher value, escape invalid characters

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19104

A good example is to set your tiapp publisher value to "Appcelerator, Inc." (like some of our samples do) and make sure the build doesn't break.

### DIFF
--- a/cli/commands/_build/validate.js
+++ b/cli/commands/_build/validate.js
@@ -19,6 +19,13 @@ function mixin(WindowsBuilder) {
 }
 
 /**
+ * Escapes a value in a relative distinguished name. Used to generate the distinguished name for the windows cert string from tiapp publisher value.
+ **/
+function escapeRDN(value) {
+	return value.trim().replace(/\\/g, '\\\\').replace(/([\/,#+<>;"=])/g, '\\$1');
+}
+
+/**
  * Validates the Windows build-specific arguments, tiapp.xml settings, and environment.
  *
  * @param {Object} logger - The logger instance.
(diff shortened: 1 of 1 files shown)


cli/commands/_build/validate.js
```diff
@@ -19,6 +19,25 @@ function mixin(WindowsBuilder) {
 }
 
 /**
+ * Escapes a value in a relative distinguished name. Used to generate the distinguished name for the windows cert string from tiapp publisher value.
+ **/
+function escapeRDN(value) {
+	var escapeChar = "\\",
+		chars = ["/", ",", "#", "+", "<", ">", ";", '"', "="],
+		// cheat and just trim leading and trailing whitespace, rather than escaping
+		escaped = value.trim().replace(escapeChar, escapeChar + escapeChar);
```
dawsontoth: This will only replace the first instance of the character.

    "what\\what\\".trim().replace("\\", "\\\\") === "what\\what\"

But maybe that's ok with what we're escaping? I'm not sure...

Better, maybe:

    return value.replace(/\\/g, '\\\\').replace(/([\/,#+<>;"=])/g, '\\$1');

sgtcoolguy: I was going to use regexp but got lazy. This is much better. Thanks! New commit coming soon...

dawsontoth:
:+1:

dawsontoth:
I should have rested this more...

```bash
[WARN]  CMake Error at CMakeLists.txt:67 (set):
  Syntax error in cmake code at

    C:/Users/Dawson/AppData/Local/Temp/exapp/CMakeLists.txt:67

  when parsing string

    CN=Appcelerator\, Inc.

  Invalid escape sequence \,
```

PR #349, opened 2015-07-02
[TIMOB-19119] Fix layout ppi value

sgtcoolguy:
This supercedes #341 

I fiddled with checking all the values of actual pixels, versus what the APIs report, the logical dpi, raw dpi, etc. Eventually I figured out how many pixels equaled an inch (based on using the stated size of the emulator as being the diagonal size) and found the right values that would scale us properly when using inches as our units and made some buttons to ensure they looked the right size visually given the expected actual width in inches of the device.

By my calculations, the 6" emulator should be ~2.958" wide.
The 4.7" emulator should be ~2.304" wide.

So then I just generated a simple app and set some views up using inches as measurements and verified those widths panned out. They (more or less) do now.

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -15,6 +15,7 @@
 #include <algorithm>
 #include <cctype>
 #include <collection.h>
+#include <Windows.h>
 
 #include "TitaniumWindows/Utility.hpp"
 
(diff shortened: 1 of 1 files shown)


garymathews:
@sgtcoolguy Looks good! I'll remove my changes from my PR.

dawsontoth:
13:18:56   C:\Jenkins\workspace\titanium_mobile_windows_prs\Source\UI\src\WindowsViewLayoutDelegate.cpp(829): error C2039: 'RawPixelsPerViewPixel' : is not a member of 'Windows::Graphics::Display::DisplayInformation' [C:\Jenkins\workspace\titanium_mobile_windows_prs\build\store\x86\Map\UI\TitaniumWindows_UI.vcxproj]
13:18:56 C:\Jenkins\workspace\titanium_mobile_windows_prs\Source\UI\src\WindowsViewLayoutDelegate.cpp(831): error C3536: 'rppp': cannot be used before it is initialized

sgtcoolguy:
Yeah, looks like the RawPixelsPerPixel isn't available on Store apps. For those we'll have to fall back on LogicalDPI?

dawsontoth:
:+1:

garymathews:
Changes for _FIXME_

```C++
auto dpi = info->RawDpiX;
switch (name) {
	case Titanium::LayoutEngine::ValueName::Bottom:
	case Titanium::LayoutEngine::ValueName::CenterY:
	case Titanium::LayoutEngine::ValueName::Height:
	case Titanium::LayoutEngine::ValueName::MinHeight:
	case Titanium::LayoutEngine::ValueName::Top:
		dpi = info->RawDpiY;
}
auto rppp = info->RawPixelsPerViewPixel;
auto ppi = dpi / rppp;
```

PR #352, opened 2015-07-03
[TIMOB-19152] Allow encrypted file types to load if they are not encrypted

garymathews:
- Allows unencrypted file types to load by checking if an ```.iv``` is present
- ```seed__``` is now a ```::Platform::String^``` to prevent unnecessary string conversions
- Changed variable names to match our naming convention

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19152)


sgtcoolguy:
@dawsontoth Were you able to see if this also fixed https://jira.appcelerator.org/browse/TIMOB-19135 and https://jira.appcelerator.org/browse/TIMOB-19157 ?

PR #355, opened 2015-07-06
Fix TIMOB-19144 Windows: Ti.version is blank

sgtcoolguy:
This is a very ugly hack. Basically we throw the sdk version from the CLI build into the _app_info_.json file along with some of the other tiapp values. Then When we load those up in the Ti.App module, we load the sdkVersion value too. And when the user asks for Ti.version, we basically just secretly do Ti.App._sdkVersion to grab that value.

https://jira.appcelerator.org/browse/TIMOB-19144


dawsontoth:
**CR** and **FR** good.

PR #356, opened 2015-07-06
Fix TIMOB-19112 Windows: Splash Screen not showing

sgtcoolguy:
Changed the copy order. First we copy over the template files, then copy the app files over top. That way app's branding/assets "win".

https://jira.appcelerator.org/browse/TIMOB-19112


dawsontoth:
cr looks good

dawsontoth:
fr good too

PR #361, opened 2015-07-07
[TIMOB-19186] Optimize getAppInfo()

garymathews:
- If ```_app_info_.json``` does not exist, don't keep attempting to load it

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19186)

### DIFF
--- a/Source/TitaniumKit/include/Titanium/App.hpp
+++ b/Source/TitaniumKit/include/Titanium/App.hpp
@@ -228,6 +228,7 @@ namespace Titanium
 	protected:
 #pragma warning(push)
 #pragma warning(disable : 4251)
+		JSValue app_info__;
 		bool accessibilityEnabled__;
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
@garymathews I could be wrong, but this doesn't seem to address the issue where we load the file multiple times, once per property instead of just loading the file once and grabbing the underlying values from the json. I could swear that has always been the case, regardless of the use of static keyword on the json var.

garymathews:
I will look into this to see if that's happening.

garymathews:
Ok, I've addressed the issue. Very strange issue with ```static``` variables inside the function. It was only the _second_ call that changed the variables value.

sgtcoolguy:
@garymathews OK to merge once build passes

infosia:
Looks good, merging.

PR #365, opened 2015-07-08
Fix: Ti.UI.TextField.value = undefined will display value as text

infosia:
Fix for [TIMOB-19124](https://jira.appcelerator.org/browse/TIMOB-19124).

```javascript
 
win.add(Ti.UI.createTextField({
	top: 100,
	height: 50,
        width: Ti.UI.FILL,
        value: undefined
}));
 
win.open();
```


dawsontoth:
**CR** and **FR** good for TextField. Cloned ticket for same thing on Labels. Also verified that TextArea behaves correctly, like TextField will now.

dawsontoth:
Will merge momentarily.

PR #366, opened 2015-07-08
[TIMOB-19104] Remove Invalid Chars from Publisher Name

dawsontoth:
Took the regex from Microsoft's specifications for publisher names.
Also, move the trimming so that it will apply to all passed in publisher
names.

It should be tested to work with various deploy types, using a publisher name that contains character's Microsoft doesn't like, such as commas and spaces.

```bash
ti build -p windows -G 00000000-0000-1000-8000-000000000000 -C 8-1-1
ti build -p windows -G 00000000-0000-1000-8000-000000000000 --target dist-winstore
```

https://jira.appcelerator.org/browse/TIMOB-19104

### DIFF
--- a/cli/hooks/ws-package.js
+++ b/cli/hooks/ws-package.js
@@ -44,8 +44,7 @@ exports.init = function (logger, config, cli) {
 				return;
 			}
 
-			var expirationDate = moment().add(1, 'year').format('MM/DD/YYYY'),
-				pvk = path.join(projectDir, 'generated.pvk'),
+			var pvk = path.join(projectDir, 'generated.pvk'),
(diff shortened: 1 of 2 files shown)


cli/hooks/ws-package.js
```diff
@@ -44,8 +44,7 @@ exports.init = function (logger, config, cli) {
 				return;
 			}
 
-			var expirationDate = moment().add(1, 'year').format('MM/DD/YYYY'),
```
dawsontoth: Note that `expirationDate` was an unused variable, holdover from an earlier approach to certificate generation.

PR #369, opened 2015-07-10
Fix: RSS Reader and Corporate app

infosia:
Fix for [TIMOB-19193](https://jira.appcelerator.org/browse/TIMOB-19193) and [TIMOB-19195](https://jira.appcelerator.org/browse/TIMOB-19195)

- Fix: RSS Reader fails to show RSS feed items in ListView
- Fix: Corporate app crashes at directory search
- Add: RSS Reader example

To test RSS Reader: generate `RSSReader` project and then run `RSSReader.sln`, at least you should see a ListView like following without errors.

```
$ titanium_mobile_windows/Tools/Scripts/generate_vs_project.sh RSSReader
```

windows_8_1

To test Corporate app:generate `Corporate` project and then run `Corporate.sln`, you should be able to search directory as you type in text area. NOTE: it may be still slow to reload the views as you type but at least it should not crash.

```
$ titanium_mobile_windows/Tools/Scripts/generate_vs_project.sh Corporate
```


dawsontoth:
Should the RSSReader support clicking on items without erroring? I get a runtime error:

```
Rruntime Error during itemclick event: undefined is not an object (evaluating 'model.get')
```

dawsontoth:
Also, how come we're absent from the corporate directory? Suspicious... :eyes:

PR #371, opened 2015-07-13
Fix so visible property changes are actually done

sgtcoolguy:
We had overridden hide() and show() for Windows, but really could have kept them as set_visible(true/false) calls up in TitaniumKit. We never overrode set/get_visible though, so assigning to the visible property wasn't honored.

Fixed that and added a unit test for it.


sgtcoolguy:
test this please.

dawsontoth:
*cr** and **fr** passed

PR #375, opened 2015-07-15
FIx; Wrong use of std::unordered_map::emplace

infosia:
Fix for [TIMOB-19216](https://jira.appcelerator.org/browse/TIMOB-19216)

`std::unordered_map::emplace` should not be used to replace elements because it works only when key is unique. Also added assertion to make sure if it's unique when you use `emplace`.


dawsontoth:
*cr* and *fr* passed

PR #379, opened 2015-07-16
Fix: setTimeout delay param should be optional

infosia:
Fix for [TIMOB-19212](https://jira.appcelerator.org/browse/TIMOB-19212).

- Fix: setTimeout delay param should be optional
- Fix wrong line break and remove obsolete comments

```javascript
setTimeout(function () {
    Ti.API.info('yay!');
});
```

### DIFF
--- a/Source/TitaniumKit/src/GlobalObject.cpp
+++ b/Source/TitaniumKit/src/GlobalObject.cpp
@@ -474,9 +474,6 @@ namespace Titanium
 		return std::make_shared<UnimplementedTimer>(callback, interval);
 	}
 
-	// TODO: The following functions can automatically be generated from
-	// the YAML API docs.
-
 	void GlobalObject::JSExportInitialize()
 	{
 		TITANIUM_LOG_DEBUG("GlobalObject::JSExportInitialize");
@@ -498,8 +495,6 @@ namespace Titanium
 
 	TITANIUM_FUNCTION(GlobalObject, require)
 	{
-		// TODO: Validate these precondition checks (which could be
-		// automaticaly generated) with the team.
 		ENSURE_STRING_AT_INDEX(moduleId, 0);
 
 		const auto global_object = this_object.get_context().get_global_object();
@@ -511,10 +506,8 @@ namespace Titanium
 
 	TITANIUM_FUNCTION(GlobalObject, setTimeout)
 	{
-		// TODO: Validate these precondition checks (which could be
-		// automaticaly generated) with the team.
 		ENSURE_OBJECT_AT_INDEX(function, 0);
-		ENSURE_UINT_AT_INDEX(delay, 1);
+		ENSURE_OPTIONAL_UINT_AT_INDEX(delay, 1, 0);
 		
 		const auto chrono_delay = std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(delay));
 
@@ -527,8 +520,6 @@ namespace Titanium
 
 	TITANIUM_FUNCTION(GlobalObject, clearTimeout)
 	{
-		// TODO: Validate these precondition checks (which could be
-		// automaticaly generated) with the team.
 		ENSURE_UINT_AT_INDEX(timerId, 0);
 
 		const auto global_object = this_object.get_context().get_global_object();
@@ -542,9 +533,6 @@ namespace Titanium
 
 	TITANIUM_FUNCTION(GlobalObject, setInterval)
 	{
-		// TODO: Validate these precondition checks (which could be
-		// automaticaly generated) with the team.
-		
 		ENSURE_OBJECT_AT_INDEX(function, 0);
 		ENSURE_UINT_AT_INDEX(delay, 1);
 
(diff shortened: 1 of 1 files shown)


yuchi:
Also `clearTimeout` ([here](https://github.com/infosia/titanium_mobile_windows/blob/8748bfd50afb689cd7e9cea26ac5684b038e6d43/Source/TitaniumKit/src/GlobalObject.cpp#L523)) shouldn’t throw on non-numeric values.

```js
var timeoutId;

function delayed() {
  clearTimeout(timeoutId);
  setTimeout(implementation, 1e3);
}

function implementation() {
  console.log('hi');
}

delayed(); // timeoutId is still undefined
delayed();
delayed();
```

yuchi:
Didn’t tested it, but I expect it to throw, given [this line](https://github.com/appcelerator/titanium_mobile_windows/blob/2a2e1846c389ad0f062136c8ff5ac085f6907eda/Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp#L131).

sgtcoolguy:
@infosia @yuchi brings up a good point. Now, I think our asserts actually don't do anything in release builds for SDK users, but fire exceptions in Debug mode for SDK developers. So I think in this case it'd just keep going and try to cast the arg to a number, and at that point it may fail to do so?

(See https://github.com/appcelerator/titanium_mobile_windows/blob/2a2e1846c389ad0f062136c8ff5ac085f6907eda/Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp#L12 where we basically make the asserts no-ops)

This is sort of related to https://jira.appcelerator.org/browse/TIMOB-19124 - Namely we want to be more forgiving in terms of how we handle some unexpected input. In that case it was treating undefined/null as empty string for the label. In this case, it's a matter of what we do when the arg is not a number like we expect. Here we assert and continue, perhaps the right thing to do is to assert and if the assertion fails, it logs some error to console and then basically just bails and doesn't try to do anything else? It's almost like we want the behavior to differ based on the deploy type rather than a compiler define, since we likely want the red screen of death/stricter enforcement for developers when they're testing locally, but the more liberal attempt to continue in production builds of their apps.

yuchi:
@sgtcoolguy If we go deeper on the issue we see that the way we handle the casting is potentially different from the spec. Mandatory notes: a) *potentially different* because it’s not explicitly bound to ES spec terms and b) `{set,clear}{Timeout,Interval,Immediate}` are not part of EcmaScript but of HTML.

So we have 2 different issues at hand (if I got it right):

- timers behave in the wrong way (small note: on Android too!),
- casting is home-brewed.

For the first part let’s follow the [specification](https://html.spec.whatwg.org/multipage/webappapis.html#timers).

For the casting I’d suggest to follow the [`ToInteger`](http://www.ecma-international.org/ecma-262/6.0/#sec-tointeger) spec (and you can see it used [here on `String.prototype.slice`](http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.slice)) and friends if this is not already the case.

sgtcoolguy:
@yuchi I'm not sure I follow. The casting is between JS arguments and native C++ types. We use a static_cast to an int32_t which ends up here: https://github.com/appcelerator/HAL/blob/master/src/detail/JSUtil.cpp#L288

I guess you're saying you'd prefer we followed the ECMAScript spec for ToInteger rather than ToInt32 for the timeout value? the spec you posted for setTimeout declares the numbers as "long", which is equivalent in value range to int32: https://msdn.microsoft.com/en-us/library/s3f49ktz.aspx

infosia:
@sgtcoolguy @yuchi For the casting number value I think we can keep using [to_int32_t](https://github.com/appcelerator/HAL/blob/master/src/detail/JSUtil.cpp#L288). For the topic how our assertion should work, I started thinking that we should not throw JS exception but log these as warning message on the console, so that it won't block the app but notify developers only on the Debug mode. I think current assertion (JS exception) is too strict in the real world.

yuchi:
What I think we should do is:

- ensure that all casting behaves like other implementations, such as the browser, and I *expect* them to follow the `To*` abstract methods' spec
- not throw in the casting
- have the same semantics (I currently cannot compare the spec to the casting macros, I'm on mobile)

yuchi:
Ok, I had a deeper look.

- the `double -> int32` conversion is explicitly following the spec at `ToInt32`. Great.
- we’re not following the spec on the initial casting. More precisely I see no reference to [`ToNumber`](http://www.ecma-international.org/ecma-262/6.0/#sec-tonumber). If this is true, `clearTimeout(..., "222")` throws.

infosia:
Pushed #384, confirmed that the implicit conversion like `clearTimeout("222")` is working even on Debug mode.

```javascript
function delayed(e) {
    if (e.index == 0) {
        setTimeout(function () {
            createAlert(delayed);
        }, "2000");
    }
}

function createAlert(callback) {
    var alertDialog = Ti.UI.createAlertDialog({
        message: 'tests',
        buttonNames: ['create alert', 'close']
    });
    alertDialog.addEventListener('click', callback);
    alertDialog.show();
}

createAlert(delayed);
```

PR #380, opened 2015-07-16
Fix: Word wrap on labels does not work if width and height are not set

infosia:
Fix for [TIMOB-19048](https://jira.appcelerator.org/browse/TIMOB-19048).

```javascript

var lbl = Ti.UI.createLabel({
    text: '*** Resources Dir: Test presence of file that exists => blahTest presence of file that does not exist => bleh*** appData Dir:Test presence of file that exists => meh Test presence of file that does not exist => merh',
    color: 'black'
});

win.add(lbl);
win.open();
```

windows_8_1

### DIFF
--- a/Source/UI/src/Label.cpp
+++ b/Source/UI/src/Label.cpp
@@ -24,11 +24,14 @@ namespace TitaniumWindows
 		{
 			WindowsViewLayoutDelegate::setComponent(label);
 
-			// TIMOB-19048: max width is set to screen width by default
-			defaultMaxWidth__ = label->MaxWidth;
+			// TIMOB-19048: max size is set to screen size by default
+			defaultMaxWidth__  = label->MaxWidth;
+			defaultMaxHeight__ = label->MaxHeight;
 			const auto current = Windows::UI::Xaml::Window::Current;
 			if (current) {
-				label->MaxWidth = current->Bounds.Width;
+				auto b = current->Bounds;
+				label->MaxWidth  = current->Bounds.Width;
+				label->MaxHeight = current->Bounds.Height;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Source/UI/src/Label.cpp
```diff
@@ -15,6 +14,32 @@ namespace TitaniumWindows
...
+		WindowsLabelLayoutDelegate::WindowsLabelLayoutDelegate() TITANIUM_NOEXCEPT
+			: WindowsViewLayoutDelegate()
+		{
+		}
+
+		void WindowsLabelLayoutDelegate::setComponent(Windows::UI::Xaml::FrameworkElement^ label) 
+		{
+			WindowsViewLayoutDelegate::setComponent(label);
+
+			// TIMOB-19048: max width is set to screen width by default
```
sgtcoolguy: So setting this to the width of the current window seems Ok for the case when the label is basically just a child of the windows (or some view that fills the window). I assume the true max width should likely be the width of the parent component. If the label is part of some view that is expected to be half the size of the window, we don't want the label to extend out past it, do we?

infosia: you're right. mmm..

infosia:
```javascript

var view = Ti.UI.createView({
    width: '80%',
    height: '80%',
    backgroundColor: 'gray'
});
var lbl = Ti.UI.createLabel({
    text: '*** Resources Dir: Test presence of file that exists => blahTest presence of file that does not exist => bleh*** appData Dir:Test presence of file that exists => meh Test presence of file that does not exist => merh',
    color: 'black'
});

view.add(lbl);
win.add(view);
win.open();
```

windows_8_1

Source/UI/src/Label.cpp
infosia: should be fixed now.

PR #381, opened 2015-07-16
[TIMOB-19224] Fix 'dp' calculation

garymathews:
- Fixed ```dp``` calculation
- ```1dp is equal to 1 physical pixel on a 160 DPI screen```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19224)

### DIFF
--- a/Source/LayoutEngine/src/ParseProperty.cpp
+++ b/Source/LayoutEngine/src/ParseProperty.cpp
@@ -66,7 +66,7 @@ namespace Titanium
 				} else if (units == "px") {
 					return parsedValue;
 				} else if (units == "dp" || units == "dip") {
-					return parsedValue * (ppi / 96);  // 96 - http://en.wikipedia.org/wiki/Dots_per_inch
+					return parsedValue * ppi / 160;
 				}
 			}
 			return 0;


Source/LayoutEngine/src/ParseProperty.cpp
```diff
@@ -66,7 +66,7 @@ namespace Titanium
 				} else if (units == "px") {
 					return parsedValue;
 				} else if (units == "dp" || units == "dip") {
-					return parsedValue * (ppi / 96);  // 96 - http://en.wikipedia.org/wiki/Dots_per_inch
+					return parsedValue * 160 / ppi;
```
dawsontoth: This is supposed to turn a unit in dots-per-inch to pixels, correct?

I believe the correct calculation is `parsedValue * ppi / 160`.

For example:

320dp on a 160ppi screen should return 320px (which your equation handles correctly).
160dp on an 320ppi screen should return 320px as well (which your equation doesn't handle correctly -- it'd return 80px). Since it's double the density of the reference ppi of 160, it should return twice the number of pixels. This is to allow the units to be density independent.

dawsontoth: Otherwise extrapolated as "1dp is equal to 1 physical pixel on a 160 DPI screen" therefore 1dp is equal to 2 physical pixels on a 320 DPI screen. It's easy to visualize this with a non-retina iPhone, with the same size as a retina iPhone, where the resolutions are just doubled. Saying something should take up 160dp (half the screen) would be 160 actual px on the non-retina, and 320 actual px on the retina, but visually look like 50% on both.

garymathews:
Yep, I got it the wrong way round. Unfortunately the device I tested on has a DPI of 165 so everything looked correct. I'll correct it.

garymathews:
You're right, Windows does use ```1/96"``` as 1 _view pixel_ and Android uses ```1/160"``` as 1 _density pixel_. But we want to keep parity with the other platforms. Using ```1/96"``` means we have to change the ```dp``` scalars so they look correct on Windows and use different values for other platforms. Which is why we should use ```1/160"``` since that wouldn't require any changes and the measurements would look the same on all platforms.

garymathews:
Comparison of Geocoder on different platforms.

###### Windows (with this dp fix)
windows
###### iOS
ios
###### Android
android

PR #382, opened 2015-07-16
[TIMOB-19176] Allow analytics to receive a response before suspending

garymathews:
- Delay app suspension to allow analytics to receive a response from the server

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19176)

### DIFF
--- a/Source/Titanium/src/TitaniumWindows.cpp
+++ b/Source/Titanium/src/TitaniumWindows.cpp
@@ -251,10 +251,24 @@ namespace TitaniumWindows
 	{
 		auto deferral = e->SuspendingOperation->GetDeferral();
 
+		js_context__.JSEvaluateScript("Ti.Analytics._receivedResponse = false");
+
 		// Since we only have "suspending" event, we fires both pause and paused event.
 		GET_TITANIUM_APP(App);
 		App->fireEvent("pause");
 		App->fireEvent("paused");
+
+		// wait for Ti.Analytics to receive a response
+		auto retry_count = 0;
+		while (static_cast<bool>(js_context__.JSEvaluateScript("!Ti.Analytics._receivedResponse"))) {
+			Sleep(200);
+			retry_count++;
+
+			// wait for a maximum of 3000ms
+			if (retry_count > 15) {
+				break;
+			}
+		}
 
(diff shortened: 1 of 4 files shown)


Source/Titanium/src/TitaniumWindows.cpp
```diff
@@ -256,6 +256,9 @@ namespace TitaniumWindows
 		App->fireEvent("pause");
 		App->fireEvent("paused");
 
+		// TIMOB-19176: Allow time for response from analytics server
+		Sleep(3000);
```
infosia: @garymathews  Do we want to wait 3 sec every time before suspend? Could we check latest response state and then suspend? Maybe something like...?

```c++
auto wait_for_response = true;
auto retry_count = 0;
while(wait_for_response) {
    wait_for_response = static_cast<bool>(JSEvaluateScript("Ti.Analytics.lastEvent != 'ti.background'"));
    Sleep(100);
    retry_count++;
    if (retry_count > 5) break;
}
```

By the way, looks like lastEvent is returning JS object, but according to the API doc it should have returned a String.

garymathews:
@infosia Updated PR

Source/Titanium/src/TitaniumWindows.cpp
```diff
@@ -251,10 +251,24 @@ namespace TitaniumWindows
 	{
 		auto deferral = e->SuspendingOperation->GetDeferral();
 
+		js_context__.JSEvaluateScript("Ti.Analytics._receivedResponse = false");
```
infosia: Is `receivedResponse` meant to use only for suspend event? I would do this inside `Analytics.prototype.postEvents` so that it can be used on every events...:

```javascript
Analytics.prototype.postEvents = function postEvents() {
	if (Ti.Network.online) {
          this.receivedResponse = false;
...
        }
...
}
```

garymathews:
I would of done that, but I have to wait for the ```pause``` event to get fired, ```receivedResponse``` could still be ```true``` before when it gets to line 263. I guess it wouldn't do any harm __also__ setting it to false in ```postEvents```.

garymathews:
@infosia Updated PR

PR #389, opened 2015-07-24
Fix: Logging does not work when using device

infosia:
Fix for [TIMOB-19077](https://jira.appcelerator.org/browse/TIMOB-19077).


infosia:
I should note that in order to make logging work on your device, you should enable wi-fi because `windowslib.LogRelay` needs HTTP network access to your computer.

PR #394, opened 2015-07-28
Implement Ti.Media.VideoPlayer, Sound and AudioPlayer

infosia:
[TIMOB-17913](https://jira.appcelerator.org/browse/TIMOB-17913) and [TIMOB-17918](https://jira.appcelerator.org/browse/TIMOB-17918)

- Ti.Media.Sound
- Ti.Media.VideoPlayer
- Ti.Media.AudioPlayer
- Add sample media resource (cricket.wav)

VideoPlayer:
```javascript
var window = Ti.UI.createWindow({});
var player = Ti.Media.createVideoPlayer({ url: "cricket.wav" });

player.addEventListener('change', function (e) {
    Ti.API.info('[video change]');
});

player.addEventListener('complete', function (e) {
    Ti.API.info('[video complete]');
});

player.addEventListener('error', function (e) {
    Ti.API.info('[video error]' + e.message);
});

window.add(player);
window.open();
```

Sound
```javascript
var window = Ti.UI.createWindow({});
var player = Ti.Media.createSound({ url: "cricket.wav" });
var playButton = Ti.UI.createButton({title:'Play'});

playButton.addEventListener('click', function () {
    player.play();
});

player.addEventListener('change', function (e) {
    Ti.API.info('[sound change]');
});

player.addEventListener('complete', function (e) {
    Ti.API.info('[sound complete]');
});

player.addEventListener('error', function (e) {
    Ti.API.info('[sound error]' + e.message);
});

window.add(playButton);
window.open();
```

Note: Using `Windows::UI::Xaml::Controls::MediaElement` behind `Ti.Media.Sound` and `Ti.Media.AudioPlayer` was bit tricky because `MediaElement` is Xaml UI component but `Sound` and `AudioPlayer` are not a Titanium View proxy. I ended up adding "hidden" MediaElement onto current Window because `MediaElement` doesn't work when it is not part of UI.


Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+			std::string url = param;
+			if (param.find("://") == std::string::npos) {
+				url = "ms-appx:///" + param;
+			}
+			Titanium::Media::AudioPlayer::set_url(url);
+			player__->Source = ref new Windows::Foundation::Uri(TitaniumWindows::Utility::ConvertString(url));
+		}
+
+
+		bool AudioPlayer::isPaused() TITANIUM_NOEXCEPT
```
sgtcoolguy: Looks like this and isPlaying could be implemented up in TitaniumKit level by doing get_state() == Titanium::Media::AudioState::Paused (and get_state() == Titanium::Media::AudioState::Playing)

Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+			if (paused) {
+				pause();
+			} else {
+				play();
+			}
+		}
+
+		std::chrono::milliseconds AudioPlayer::get_time() const TITANIUM_NOEXCEPT
+		{
+			return std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(player__->Position.Duration / 10000));
```
sgtcoolguy: don't we have some utility method or constant defined for the  special 10000 value?

feels like we do a bit of this in Windows where we convert their nanosecond time values to millisecond. I assume the std::chrono API could even do the conversion for us. We may want to introduce a TitaniumWindows::Utility method to convert their nanosecond readings to milliseconds.

Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+		}
+
+		void AudioPlayer::set_url(const std::string& param) TITANIUM_NOEXCEPT
+		{
+			std::string url = param;
+			if (param.find("://") == std::string::npos) {
```
sgtcoolguy: This looks familiar too :)

Again, we may want some utility method to convert relative paths to the special ms-appx URLs? I think we do this for the File proxy too?

Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+				url = "ms-appx:///" + param;
+			}
+			Titanium::Media::AudioPlayer::set_url(url);
```
sgtcoolguy: What do we do for other platforms? I think we just take the original value before we "fix" it and store that as the property value. Not saying that's the right approach, but we may want to retain it (which would mean moving this up to the top of the method).

Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+					event_arg.SetProperty("code", ctx.CreateNumber(0));
+					fireEvent("complete", event_arg);
+				});
+			} else if (event_name == "error") {
+				failed_event__ = player__->MediaFailed += ref new ExceptionRoutedEventHandler([this](Platform::Object^ sender, ExceptionRoutedEventArgs^ e) {
+					const auto ctx = get_context();
+					auto event_arg = ctx.CreateObject();
+					event_arg.SetProperty("message", ctx.CreateString(TitaniumWindows::Utility::ConvertString(e->ErrorMessage)));
+					event_arg.SetProperty("success", ctx.CreateBoolean(false));
+					event_arg.SetProperty("code", ctx.CreateNumber(-1));
```
sgtcoolguy: So it's not obvious at _all_ but I think there's a way to get an actual error code out of the e.ErrorMessage. At the bottom of this page: https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.exceptionroutedeventargs?f=255&MSPPError=-2147217396

They show how they manipulate the string to pull out the HResult where they eventually get a hex value in the string (i.e. "0xFFFFFFFF").  Then you can "decode" the value according to this: https://en.wikipedia.org/wiki/HRESULT

I'm not even sure we'd want to decode, we might just want to turn it into a 32-bit int as is.

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+	namespace Media {
 		Sound::Sound(const JSContext& js_context) TITANIUM_NOEXCEPT
-			: Titanium::Media::Sound(js_context)
-		{
+			: Titanium::Media::Sound(js_context) {
 			TITANIUM_LOG_DEBUG("TitaniumWindows::Media::Sound::ctor Initialize");
 		}
 
-		Sound::~Sound()
+		Sound::~Sound()
```
sgtcoolguy: remove trailing space

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
 		}
 
-		Sound::~Sound()
+		Sound::~Sound() 
 		{
 			TITANIUM_LOG_DEBUG("TitaniumWindows::Media::Sound::dtor");
 		}
 
-		void Sound::JSExportInitialize()
+		void Sound::JSExportInitialize()
```
sgtcoolguy: remove trailing space

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+				looping__ = isLooping();
+				playing__ = isPlaying();
+				paused__ = isPaused();
+				volume__ = player__->Volume;
+				duration__ = static_cast<std::uint32_t>(player__->NaturalDuration.TimeSpan.Duration / 10000); // convert 100 nanosecond to milliseconds
+
+				const auto ctx = get_context();
+				auto event_arg = ctx.CreateObject();
+
+				switch (player__->CurrentState) {
```
sgtcoolguy: This logic looks pretty much like the get_state() logic. Can it just call that?

Source/Media/src/Media/AudioPlayer.cpp
```diff
@@ -28,14 +33,184 @@ namespace TitaniumWindows
...
+			player__->IsLooping = false;
+			player__->CurrentStateChanged += ref new RoutedEventHandler([this](Platform::Object^ sender, RoutedEventArgs^ e) {
+				playing__ = isPlaying();
+				paused__ = isPaused();
+				volume__ = player__->Volume;
+				waiting__ = (player__->CurrentState == MediaElementState::Buffering);
+				const auto ctx = get_context();
+				auto event_arg = ctx.CreateObject();
+
+				switch (player__->CurrentState) {
```
sgtcoolguy: This logic looks pretty much like the get_state() logic. Can it just call that?

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+				case MediaElementState::Opening:
+					break;
+				case MediaElementState::Closed:
+					break;
+				}
+
+				fireEvent("change", event_arg);
+			});
+
+			// Add "hidden" MediaElement UI onto current Window, because it doesn't work when it is not on the UI.
```
sgtcoolguy: Can we move this code into some common place for these Media elements that need to secretly be hung on a Window? This is just duplicated code between them.

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+			}
+		}
+
+		void Sound::set_volume(const double& volume) TITANIUM_NOEXCEPT
+		{
+			Titanium::Media::Sound::set_volume(volume);
+			player__->Volume = volume;
+		}
+
+		void Sound::set_paused(const bool& paused) TITANIUM_NOEXCEPT
```
sgtcoolguy: This looks like a candidate for an implementation to just move up to TitaniumKit

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+			if (paused) {
+				pause();
+			} else {
+				play();
+			}
+		}
+
+		std::chrono::milliseconds Sound::get_time() const TITANIUM_NOEXCEPT
+		{
+			return std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(player__->Position.Duration / 10000));
```
sgtcoolguy: another place where we convert with the special constant...

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+		}
+
+		void Sound::set_url(const std::string& param) TITANIUM_NOEXCEPT
+		{
+			std::string url = param;
```
sgtcoolguy: another case of url massaging...

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+			Titanium::Media::Sound::set_looping(looping);
+			player__->IsLooping = looping;
+		}
+
+		bool Sound::isLooping() TITANIUM_NOEXCEPT
+		{
+			return player__->IsLooping;
+		}
+
+		bool Sound::isPaused() TITANIUM_NOEXCEPT
```
sgtcoolguy: move this and isPlaying() up to TitaniumKit using get_state()

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+		{
+			player__->Pause();
+		}
+
+		void Sound::play() TITANIUM_NOEXCEPT
+		{
+			player__->Play();
+		}
+
+		void Sound::reset() TITANIUM_NOEXCEPT
```
sgtcoolguy: This might be a candidate for default behavior to be moved up into TitaniumKit. Technically according to docs this just resets position back to beginning, doesn't mention actually stopping. Probably need to set the Position property.

Docs show a "time" property which I assume maps to MediaElement.Position, and doesn't say it's read-only. So if user's can set "time", we could move the impl up to TitaniumKit and have reset just set "time" property to 0.

Source/Media/src/Media/Sound.cpp
```diff
@@ -6,36 +6,202 @@
...
+					event_arg.SetProperty("code", ctx.CreateNumber(0));
+					fireEvent("complete", event_arg);
+				});
+			} else if (event_name == "error") {
+				failed_event__ = player__->MediaFailed += ref new ExceptionRoutedEventHandler([this](Platform::Object^ sender, ExceptionRoutedEventArgs^ e) {
+					const auto ctx = get_context();
+					auto event_arg = ctx.CreateObject();
+					event_arg.SetProperty("message", ctx.CreateString(TitaniumWindows::Utility::ConvertString(e->ErrorMessage)));
+					event_arg.SetProperty("success", ctx.CreateBoolean(false));
+					event_arg.SetProperty("code", ctx.CreateNumber(-1));
```
sgtcoolguy: Again, we could hack out he actual code from the ErrorMessage, maybe.

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
 			JSExport<VideoPlayer>::SetParent(JSExport<Titanium::Media::VideoPlayer>::Class());
 		}
 
+		void VideoPlayer::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) 
+		{
+			Titanium::Media::VideoPlayer::postCallAsConstructor(js_context, arguments);
+			player__ = ref new Windows::UI::Xaml::Controls::MediaElement();
+			player__->AreTransportControlsEnabled = true;
+			player__->Visibility = Windows::UI::Xaml::Visibility::Visible;
+			player__->AutoPlay = false;
```
sgtcoolguy: According to docs, the default is true for autoplay.

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
 		}
 
+		void VideoPlayer::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) 
+		{
+			Titanium::Media::VideoPlayer::postCallAsConstructor(js_context, arguments);
+			player__ = ref new Windows::UI::Xaml::Controls::MediaElement();
+			player__->AreTransportControlsEnabled = true;
+			player__->Visibility = Windows::UI::Xaml::Visibility::Visible;
+			player__->AutoPlay = false;
+			player__->IsLooping = true;
```
sgtcoolguy: do the other platforms autoloop by default?

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+		{
+			Titanium::Media::VideoPlayer::set_urls(urls);
+
+			if (urls.size() == 0) {
+				return;
+			}
+
+			// MediaElement can take only one url. Let's take the first one.
+			std::string url = urls.at(0);
+			if (url.find("://") == std::string::npos) {
```
sgtcoolguy: url massaging utility method...

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+		void VideoPlayer::set_autoplay(const bool& autoplay) TITANIUM_NOEXCEPT
+		{
+			Titanium::Media::VideoPlayer::set_autoplay(autoplay);
+			player__->AutoPlay = autoplay;
+		}
+
+		std::chrono::milliseconds VideoPlayer::get_duration() const TITANIUM_NOEXCEPT
+		{
+			// convert 100 nanosecond to milliseconds
+			return std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(player__->NaturalDuration.TimeSpan.Duration / 10000));
```
sgtcoolguy: nanosecond -> millisecond utility method

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+		}
+
+		std::chrono::milliseconds VideoPlayer::get_initialPlaybackTime() const TITANIUM_NOEXCEPT
+		{
+			return std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(0));
```
sgtcoolguy: default should be zero here, but the user could set this. This sounds like a property we need to keep track of and before we "play" we manipulate the Position property of the MediaElement to start at this time. So basically we need a field up in TitaniumKit, and we need to use normal getter/setters, but before we play for this first time, we set the "time" to this value.

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+		}
+
+		std::chrono::milliseconds VideoPlayer::get_endPlaybackTime() const TITANIUM_NOEXCEPT
```
sgtcoolguy: similar to start time above, this looks like a property we need to set/get up in TitaniumKit, and is a way to "clip" the end of the video. Unfortunately I don't really see any way to actually do that using MediaElement's API. So in this case, we may just have to ignore the value.

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+
+		std::chrono::milliseconds VideoPlayer::get_playableDuration() const TITANIUM_NOEXCEPT
+		{
+			return get_duration();
+		}
+
+		std::chrono::milliseconds VideoPlayer::get_currentPlaybackTime() const TITANIUM_NOEXCEPT
+		{
+			// convert 100 nanosecond to milliseconds
+			return std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(player__->Position.Duration / 10000));
```
sgtcoolguy: nanosceond -> millisecond conversion utility method

Source/Media/src/Media/VideoPlayer.cpp
```diff
@@ -28,14 +34,145 @@ namespace TitaniumWindows
...
+					fireEvent("complete", event_arg);
+				});
+			}
+			else if (event_name == "error") {
+				failed_event__ = player__->MediaFailed += ref new ExceptionRoutedEventHandler([this](Platform::Object^ sender, ExceptionRoutedEventArgs^ e) {
+					const auto ctx = get_context();
+					auto event_arg = ctx.CreateObject();
+					event_arg.SetProperty("message", ctx.CreateString(TitaniumWindows::Utility::ConvertString(e->ErrorMessage)));
+					event_arg.SetProperty("success", ctx.CreateBoolean(false));
+					event_arg.SetProperty("code", ctx.CreateNumber(-1));
```
sgtcoolguy: hack HREsult

Source/TitaniumKit/src/Media/Sound.cpp
```diff
@@ -235,14 +225,14 @@ namespace Titanium
 		TITANIUM_FUNCTION(Sound, setLooping)
```
sgtcoolguy: don't we have macros for for redirecting getter/setter methods to the getter/setter for properties?

sgtcoolguy:
test this please.

Source/Media/src/Media/Sound.cpp
infosia: I don't know why but `Sound` does not have `state` property unfortunately...

PR #400, opened 2015-08-11
Implement Ti.Media methods

infosia:
[TIMOB-18880](https://jira.appcelerator.org/browse/TIMOB-18880)

- beep
- hideCamera
- isMediaTypeSupported
- openMusicLibrary
- openPhotoGallery
- previewImage - *NOT IMPLEMENTED*
- queryMusicLibrary - *NOT IMPLEMENTED*
- switchCamera - *NOT IMPLEMENTED*
- setOverrideAudioRoute - *NOT IMPLEMENTED*
- requestAuthorization - *UNUSED*
- hideMusicLibrary - *UNUSED*
- saveToPhotoGallery
- showCamera
- startMicrophoneMonitor
- stopMicrophoneMonitor
- takePicture
- takeScreenshot
- vibrate

This PR adds new global event `windows.fileOpenFromPicker` for `Ti.Media.openPhotoGallery` and `Ti.Media.openMusicLibrary` because on Windows Phone [FileOpenPicker deactivates current app](https://msdn.microsoft.com/en-us/library/windows/apps/xaml/dn631755.aspx) and you need to restore latest state when file picker activates the app again.


infosia:
For `Ti.Media.beep()`, I created just a alias of `Ti.Media.createSound({url:'beep.wav'}).play()`. might need more tweak.

infosia:
Ready for merge, there are still some unimplemented functions which I think is not supported on the platform or functions which I couldn't find how to implement.

Source/TitaniumKit/include/Titanium/Media/Item.hpp
```diff
@@ -35,133 +35,133 @@ namespace Titanium
 			  @abstract albumArtist
 			  @discussion Artist credited for the album containing this item.
 			*/
-			TITANIUM_PROPERTY_IMPL_READONLY_DEF(std::string, albumArtist);
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, albumArtist);
```
sgtcoolguy: These properties are supposed to be read-only, but I see you're using setters to populate them. Is there any way to keep them read-only and still construct the object with values internally? It really feels to me like Titanium.Media.Item should be a struct/JSObject and not a Porxy subclass (despite what the docs say) since it has only read-only properties and not methods or events.

Maybe it doesn't matter here, and we can allow for these setters to get exposed, but in the cpp I think it does and we want to not expose the setters to JS...

Source/TitaniumKit/src/Media/Item.cpp
```diff
@@ -37,25 +37,25 @@ namespace Titanium
...
-		TITANIUM_PROPERTY_READ(Item, bool, isCompilation)
-		TITANIUM_PROPERTY_READ(Item, std::string, lyrics)
-		TITANIUM_PROPERTY_READ(Item, MusicMediaType, mediaType)
-		TITANIUM_PROPERTY_READ(Item, std::uint32_t, playCount)
-		TITANIUM_PROPERTY_READ(Item, std::chrono::seconds, playbackDuration)
-		TITANIUM_PROPERTY_READ(Item, std::string, podcastTitle)
-		TITANIUM_PROPERTY_READ(Item, std::uint32_t, rating)
-		TITANIUM_PROPERTY_READ(Item, std::uint32_t, skipCount)
-		TITANIUM_PROPERTY_READ(Item, std::string, title)
+		TITANIUM_PROPERTY_READWRITE(Item, std::string, albumArtist)
```
sgtcoolguy: I think we can keep the setters hidden to JS here, since you're using native setters to populate the items in our API internals...

Source/Media/src/Media.cpp
```diff
@@ -269,22 +693,62 @@ namespace TitaniumWindows
...
+			auto vibrate_timer = ref new DispatcherTimer();
+			vibrate_timer->Interval = time_span;
+			vibrate_timer->Stop();
+
+			vibrate_timer->Tick += ref new Windows::Foundation::EventHandler<Platform::Object^>([this, i](Platform::Object^ sender, Platform::Object^ event) {
+				// stop current timer
+				vibrate_timers__->GetAt(i)->Stop();
+
+				// return when it's a last timer
+				const size_t next = i + 1;
```
sgtcoolguy: Do we do any validation up in TitaniumKit about the length of the timers array?

first number is delay to start, then after that it's on time, off time, on time, off time, etc.

Seems to me like we should validate that there's at least two numbers, or we don't need to do anything. Also, if there's a trailing "off" time (i.e. if the array size is odd), we should ignore/truncate it off.

Source/Media/src/Media.cpp
```diff
@@ -269,22 +693,62 @@ namespace TitaniumWindows
...
+				if (next == vibrate_timers__->Size) {
+					vibrate_timers__->Clear();
+					return;
+				}
+				// vibrate device only when even number index
```
sgtcoolguy: comment and code disagree. I believe the code is right - "odd" indices are the vibration on times. (index 1, 3, 5, etc)

Source/Media/src/Media.cpp
```diff
@@ -269,22 +693,62 @@ namespace TitaniumWindows
...
 		const auto device = VibrationDevice::GetDefault();
-		Windows::Foundation::TimeSpan duration;
-		for (const auto msec : pattern) {
-			duration.Duration = msec.count() * 10000;
-			device->Vibrate(duration); // does this a blocking function so we can push durations like this?
+
+		vibrate_timers__->Clear();
+
+		for (size_t i = 0; i < pattern.size(); i++) {
+			const auto msec = pattern.at(i);
```
sgtcoolguy: this doesn't look to be used.

Source/TitaniumKit/src/MediaModule.cpp
```diff
@@ -1274,7 +1273,7 @@ namespace Titanium
 		const auto js_pattern_array = static_cast<std::vector<JSValue>>(js_pattern);
```
sgtcoolguy: Here's where I think it makes sense to do some validation/massaging of the array - if there's not at least two numbers, we should just ignore the call. If there's an odd number of items, drop the last one (a trailing "vibration off" time).

sgtcoolguy:
Yikes, this is a lot of code, and the windows phone handling of file pickers is rather convoluted since they suspend and continue the app to do so!

I won't pretend like I was able to pore through all of the code and test everything, so I'm going to just say we should make the few changes suggested and merge it and hope for the best :smile:

sgtcoolguy:
Merging for now, and we can address the comments after...

PR #409, opened 2015-08-19
Expose async methods by wrapping returns as JS Promises

sgtcoolguy:
This is a work in progress.

Right now, this appears to work when I build in VS2013 Debug using the NG app and the generated native wrappers. However, it fails on the command line using a 5.0.0 SDK build from master and the updated stub generation - complaining it's "unable to find the variable Promise".

Perhaps it's a difference in the Debug vs Release libs of JSC? I need to dig deeper.

Also, I need to clean up the commits here and remove all the noise of the new native wrappers in the NG example.


sgtcoolguy:
cant_find_promise

infosia:
Confirmed that Promises worked for me even on Release build. FYI I'm using my own JavaScriptCOre build so maybe there's wrong path or wrong library files set for CLI?

promise

sgtcoolguy:
I cleaned up the commits so it removes the noise of updating the NG example for testing.
Included in this are some additional error handling in some proxies I added. I was seeing an uncaught exception in native async code and wasn't sure which proxy in the Ti module it was. Eventually I found it was because my Promise test code had messed up the titanium_settings.ini (I wrote new contents over it in my JS code!) so the API log code failed to parse the new contents as ini (because it wasn't anymore). So the error handling likely will never be needed, but is helpful anyways just in case.

PR #415, opened 2015-08-24
Fix TIMOB-19350 tiapp property ti.ui.defaultunit ignored

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19350

Fixed a typo in method name in LayoutEngine. Added comments and fixed calculations for certain units.
Added a Titanium::AppModule::defaultUnit() method that pulls the units from Ti.App.Properties.getString("ti.ui.defaultunit"). If the property is empty, is "system" or is not a recognized unit, we default to "px".
We pass in the default unit to the LayoutEngine along with the ppi/dpi whenever we set a layout property.
Technically this is a potentially useless perf hit since we only really need the default units if there's no valid units specified for the value.

### DIFF
--- a/Source/TitaniumKit/src/App.cpp
+++ b/Source/TitaniumKit/src/App.cpp
@@ -8,6 +8,7 @@
 
 #include "Titanium/detail/TiImpl.hpp"
 #include "Titanium/App.hpp"
+#include "Titanium/App/Properties.hpp"
 
 namespace Titanium
 {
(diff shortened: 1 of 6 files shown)


Source/TitaniumKit/src/App.cpp
```diff
@@ -128,6 +130,36 @@ namespace Titanium
...
+			JSValue Properties_property = App.GetProperty("Properties");
+			TITANIUM_ASSERT(Properties_property.IsObject());  // precondition
+			JSObject Properties = static_cast<JSObject>(Properties_property);
+
+			auto props = Properties.GetPrivate<::Titanium::App::Properties>();
+			auto defaultUnit = props->getString("ti.ui.defaultunit", boost::optional<std::string>("px"));
+
+			if (!defaultUnit || *defaultUnit == "system")
+			{
+				defaultUnit__ == "px";
```
infosia: `defaultUnit__ = "px"` ?

sgtcoolguy: Ah good catch, thanks! :+1:

PR #419, opened 2015-08-26
TIMOB-19245 Allow mixing Ti.Ui.View and Windows UI components

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19245

This likely won't work for every use case, but the simplest ones of adding Ti.UI.View subclasses to an existing native UI element via Panel.Children.Append(view); and adding a native UIElement to a Ti.UI.View: view.add(native); do work now.

In the native proxy wrappers, we basically allow for a Ti.UI.View or a Windows.UI.Xaml.UIElement as arguments  to any native method expecting a Windows.UI.Xaml.UIElement. We might want to expand this to FrameworkElement args too.

In Windows Ti.UI.View/Window::add, we expect a Ti.Ui.View or a native wrapper that is wrapping a Windows.UI.Xaml.UIElement or subclass.
We may want to consider doing the same on other places (like remove()? Maybe even other properties/things that take View as an arg?)

Here's adding a native textblock to a Ti.Ui.Window:
```javascript
var win,
    TextBlock = require('Windows.UI.Xaml.Controls.TextBlock'),
	text = new TextBlock();

text.Text = "Hello, world!";
text.FontSize = 60;

win = Ti.UI.createWindow({});
win.add(text);
win.open();
```

Here's adding a Ti.Ui.Label to a native Canvas:
```javascript
var Window = require('Windows.UI.Xaml.Window'),
    Canvas = require('Windows.UI.Xaml.Controls.Canvas'),
    window = Window.Current,
    canvas = new Canvas(),
    text;

text = Ti.UI.createLabel({
    text: "Hello, world!",
    font: {
        fontSize: 60
    }
});
canvas.Children.Append(text);

window.Content = canvas;
window.Activate();
```


infosia:
Jenkins build is failing, looks like a compile error on x86?

PR #425, opened 2015-09-04
Add Ti.Buffer, Codec, and Streams TitaniumKit - NOT READY

infosia:
*NOT READY FOR MERGE*

TitaniumKit for Ti.Buffer, Codec, and Streams API. [TIMOB-19292](https://jira.appcelerator.org/browse/TIMOB-19292) and [TIMOB-18756](https://jira.appcelerator.org/browse/TIMOB-18756).


Source/TitaniumKit/include/Titanium/Buffer.hpp
```diff
@@ -0,0 +1,186 @@
...
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#ifndef _TITANIUM_BUFFER_HPP_
+#define _TITANIUM_BUFFER_HPP_
+
+#include "Titanium/Module.hpp"
+#include "Titanium/Codec.hpp"
```
sgtcoolguy: Probably better for build time to only include a Codec/Constants.hpp for the Type/ByteOrder enums.

In fact, you may not even need to do that, but in C++11 I believe we can now forward declare enums like we do classes.

```c++
namespace Titanium
{
    namespace Codec
    {
        enum class Type;
        enum class ByteOrder;
    }
    // ...
```

Source/TitaniumKit/include/Titanium/Buffer.hpp
```diff
@@ -0,0 +1,186 @@
...
+		bool js_setProperty(const JSString& property_name, const JSValue& value);
+
+	protected:
+#pragma warning(push)
+#pragma warning(disable : 4251)
+		bool isDataIndexProperty(const std::string& property_name) const;
+
+		JSValue value__;
+		std::vector<std::uint8_t>  data__;
+		Titanium::Codec::Type      type__;
```
sgtcoolguy: oops, looks like we hold members of these enum types, so forward declaration won't work...

So I guess, I'd recommend we pull out the enums into a Codec/Constants.hpp like we do for other enums (like in UI) and just include that up top.

Source/TitaniumKit/include/Titanium/BufferStream.hpp
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License.
+ * Please see the LICENSE included with this distribution for details.
+ */
+
+#ifndef _TITANIUM_BUFFERSTREAM_HPP_
+#define _TITANIUM_BUFFERSTREAM_HPP_
+
+#include "Titanium/Module.hpp"
```
sgtcoolguy: technically, I think this isn't needed since I'd assume IOStream includes it...

Source/TitaniumKit/include/Titanium/Codec.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+namespace Titanium
+{
+	using namespace HAL;
+
+	class Buffer;
+
+	namespace Codec
+	{
+		enum class TITANIUMKIT_EXPORT ByteOrder
```
sgtcoolguy: move the enums to a Constants.hpp file...

Source/TitaniumKit/include/Titanium/Codec.hpp
```diff
@@ -0,0 +1,289 @@
...
+			Double
+		};
+
+		/*!
+		  @struct
+		  @discussion Named parameters for Titanium.Codec.decodeNumber.
+		  This is an abstract type. Any object meeting this description can be used where this type is used.
+		  See http://docs.appcelerator.com/titanium/3.0/#!/api/DecodeNumberDict  
+		*/
+		struct DecodeNumberDict
```
sgtcoolguy: if they're used anywhere else, maybe move these structs out? Probably only used here for this class's API, so staying here would make the most sense.

Source/TitaniumKit/include/Titanium/Filesystem/FileStream.hpp
```diff
@@ -8,6 +10,7 @@
 #define _TITANIUM_FILESYSTEM_FILESTREAM_HPP_
 
 #include "Titanium/Module.hpp"
```
sgtcoolguy: unnecessary since including IOStream will bring it in...

Source/TitaniumKit/include/Titanium/IOStream.hpp
```diff
@@ -0,0 +1,89 @@
...
+	{
+
+	public:
+
+		/*!
+		  @method
+		  @abstract read
+		  @discussion Reads data from this stream into a buffer.
+		*/
+		virtual std::uint32_t read(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length) TITANIUM_NOEXCEPT;
```
sgtcoolguy: docs state this will return -1 if EOF is reached, so uint32_t won't work. technically we could use a boost::optional<std::uint32_t> to model the bytes read, with the null pointer value being EOF and we then translate that to a -1 for JS. Or we can just do a straight-up int32_t.

Source/TitaniumKit/include/Titanium/IOStream.hpp
```diff
@@ -0,0 +1,89 @@
...
+	using namespace HAL;
+
+	class Buffer;
+
+	/*!
+	  @class
+	  @discussion This is the Titanium IOStream Module.
+	  See http://docs.appcelerator.com/titanium/latest/#!/api/Titanium.IOStream
+	*/
+	class TITANIUMKIT_EXPORT IOStream : public Module, public JSExport<IOStream>
```
sgtcoolguy: Given the class hierarchy involved here for streams, do we need to do any weird composition stuff like we had to for the UI View hierarchy? We'll run into the same issue of say FileStream < IOStream < Module at the TitaniumKit level, and then may have Windows-specific subclasses of IOStream and FileStream.

I guess if we implement IOStream purely in cross-platform C++ only in the TitaniumKit layer, then maybe we avoid this issue? (i.e. we don't subclass a Windows-specific IOStream)

Source/TitaniumKit/src/Buffer.cpp
```diff
@@ -0,0 +1,397 @@
...
+	{
+		const auto source = sourceBuffer->get_data();
+		const auto actualLength = offset + sourceLength > data__.size() ? data__.size() - offset : sourceLength;
+		std::copy(source.begin() + sourceOffset, source.begin() + sourceOffset + actualLength, data__.begin() + offset);
+		return 0;
+	}
+
+	std::shared_ptr<Buffer> Buffer::clone(const std::uint32_t& offset, const std::uint32_t& length) TITANIUM_NOEXCEPT
+	{
+		const auto Titanium_property = get_context().get_global_object().GetProperty("Titanium");
```
sgtcoolguy: If we use just JS to generate a Ti.Buffer (i.e just do Ti.createBuffer();) then we can avoid the need to include TiModule.hpp above.

sgtcoolguy:
test this please.

infosia:
Codec and Buffer is ready. Stream API are WIP but it's ok to merge this PR for now as long as it passes all tests.

PR #426, opened 2015-09-09
TIMOB-19394 Ti.UI.View.add should support array of views

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19394

Not sure if we want to merge this immediately. Need more direction on whether we truly want to expand the API for other platforms like this and update our docs; or keep as-is and drop support in iOS; or further expand API to also support a single or array of objects holding a "view" and "position" key like iOS supports (but is not mentioned in the ticket or docs).


yuchi:
I would love to have it on Android too for react-titanium. /cc @FokkeZB

PR #427, opened 2015-09-11
Improve compile time of WindowsNativeModuleLoader for Native API access

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-19414

Basically we whittle down the set of headers to include in Plaftorm.Object.cpp and WindowsNativeModuleLoader based on the knowledge that we have about the inheritance tree and that a given type will include it's parent header in it's own.

The effects don't seem to be all to great. In a first-time build from scratch I saw a reduction in the compile/link time from about 5:35 to 5:29. We could do the same thing in the individual proxy/wrapper cpp files too, but generally speaking the list of includes there is much smaller, and it's slightly more work to filter there right now. Given the small reduction in time for the two files with the largest set of includes, it probably isn't worth it right now.

### DIFF
--- a/cli/lib/TODO.md
+++ b/cli/lib/TODO.md
@@ -1,9 +1,6 @@
 = TODO =
-- Async methods
-	- We skip them entirely. We need to convert to a Promise like object like WInJS does
-- Avoid recompiling everything on every rebuild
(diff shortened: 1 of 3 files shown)


infosia:
ok, I'd merge this for now as long as it effects.

PR #429, opened 2015-09-15
Implement Ti.Stream, IOStream, BufferStream and BlobStream

infosia:
[TIMOB-19493](https://jira.appcelerator.org/browse/TIMOB-19493), [TIMOB-19495](https://jira.appcelerator.org/browse/TIMOB-19495), [TIMOB-19494](https://jira.appcelerator.org/browse/TIMOB-19494) and [TIMOB-19496](https://jira.appcelerator.org/browse/TIMOB-19496).

### DIFF
--- a/Source/TitaniumKit/include/Titanium/Filesystem/FileStream.hpp
+++ b/Source/TitaniumKit/include/Titanium/Filesystem/FileStream.hpp
@@ -46,7 +46,7 @@ namespace Titanium
 			  @abstract isWriteable
 			  @discussion Indicates whether this stream is writeable.
 			*/
-			virtual bool isWriteable() TITANIUM_NOEXCEPT override;
+			virtual bool isWritable() TITANIUM_NOEXCEPT override;
 
 			/*!
 			  @method
--- a/Source/TitaniumKit/include/Titanium/IOStream.hpp
+++ b/Source/TitaniumKit/include/Titanium/IOStream.hpp
@@ -10,6 +10,7 @@
 #define _TITANIUM_IOSTREAM_HPP_
 
 #include "Titanium/Module.hpp"
+#include "Titanium/Filesystem/Constants.hpp"
 
 namespace Titanium
 {
@@ -33,6 +34,12 @@ namespace Titanium
 		  @discussion Reads data from this stream into a buffer.
 		*/
 		virtual std::int32_t read(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length) TITANIUM_NOEXCEPT;
+
+		/*!
+		  @method
+		  @abstract readAsync
+		  @discussion Override this if subclass supports async operation
+		*/
 		virtual void readAsync(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length, const std::function<void(const std::int32_t&)>&) TITANIUM_NOEXCEPT;
 
 		/*!
@@ -41,14 +48,20 @@ namespace Titanium
 		  @discussion Writes data from a buffer to this stream.
 		*/
 		virtual std::uint32_t write(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length) TITANIUM_NOEXCEPT;
+
+		/*!
+		  @method
+		  @abstract writeAsync
+		  @discussion Override this if subclass supports async operation
+		*/
 		virtual void writeAsync(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length, const std::function<void(const std::int32_t&)>&) TITANIUM_NOEXCEPT;
 
 		/*!
 		  @method
-		  @abstract isWriteable
+		  @abstract isWritable
 		  @discussion Indicates whether this stream is writeable.
 		*/
-		virtual bool isWriteable() TITANIUM_NOEXCEPT;
+		virtual bool isWritable() TITANIUM_NOEXCEPT;
 
 		/*!
 		  @method
@@ -64,6 +77,26 @@ namespace Titanium
 		*/
 		virtual void close() TITANIUM_NOEXCEPT;
 
+		/*!
+		  @method
+		  @abstract set_mode
+		  @discussion Set mode to open the stream in.
+		*/
+		virtual void set_modes(const std::unordered_set<Filesystem::MODE>& mode)
+		{
+			modes__ = mode;
+		}
+
+		/*!
+		  @method
+		  @abstract get_totalBytesProcessed
+		  @discussion Get total bytes processed
+		*/
+		virtual std::uint32_t get_totalBytesProcessed() const 
+		{
+			return totalBytesProcessed__;
+		}
+
 		IOStream(const JSContext&) TITANIUM_NOEXCEPT;
 		virtual ~IOStream()                  = default;
 		IOStream(const IOStream&)            = default;
@@ -77,13 +110,15 @@ namespace Titanium
 
 		TITANIUM_FUNCTION_DEF(read);
 		TITANIUM_FUNCTION_DEF(write);
-		TITANIUM_FUNCTION_DEF(isWriteable);
+		TITANIUM_FUNCTION_DEF(isWritable);
 		TITANIUM_FUNCTION_DEF(isReadable);
 		TITANIUM_FUNCTION_DEF(close);
 
 	protected:
 #pragma warning(push)
 #pragma warning(disable : 4251)
+		std::unordered_set<Filesystem::MODE> modes__;
+		std::uint32_t totalBytesProcessed__ { 0 };
 #pragma warning(pop)
 	};
 
--- a/Source/TitaniumKit/src/Blob.cpp
+++ b/Source/TitaniumKit/src/Blob.cpp
@@ -16,7 +16,7 @@ namespace Titanium
 	{
 	}
 
-	void Blob::construct(const std::vector<std::uint8_t> data) TITANIUM_NOEXCEPT
+	void Blob::construct(const std::vector<std::uint8_t>& data) TITANIUM_NOEXCEPT
 	{
 		height_ = 0;
 		width_ = 0;
@@ -128,6 +128,11 @@ namespace Titanium
 		const auto b = blob->getData();
 		data_.reserve(data_.size() + b.size());
 		data_.insert(data_.end(), b.begin(), b.end());
+	}
+
+	void Blob::release() TITANIUM_NOEXCEPT
+	{
+		data_.clear();
 	}
 
 	TITANIUM_PROPERTY_GETTER(Blob, length)	
--- a/Source/TitaniumKit/src/Blob.cpp
+++ b/Source/TitaniumKit/src/Blob.cpp
@@ -132,7 +132,9 @@ namespace Titanium
 
 	void Blob::release() TITANIUM_NOEXCEPT
 	{
-		data_.clear();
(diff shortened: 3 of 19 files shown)


Source/TitaniumKit/include/Titanium/Filesystem/FileStream.hpp
```diff
@@ -46,7 +46,7 @@ namespace Titanium
 			  @abstract isWriteable
 			  @discussion Indicates whether this stream is writeable.
 			*/
-			virtual bool isWriteable() TITANIUM_NOEXCEPT override;
+			virtual bool isWritable() TITANIUM_NOEXCEPT override;
```
sgtcoolguy: Should be isWriteable

Source/TitaniumKit/include/Titanium/IOStream.hpp
```diff
@@ -41,14 +48,20 @@ namespace Titanium
...
 		virtual void writeAsync(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length, const std::function<void(const std::int32_t&)>&) TITANIUM_NOEXCEPT;
 
 		/*!
 		  @method
-		  @abstract isWriteable
+		  @abstract isWritable
 		  @discussion Indicates whether this stream is writeable.
 		*/
-		virtual bool isWriteable() TITANIUM_NOEXCEPT;
+		virtual bool isWritable() TITANIUM_NOEXCEPT;
```
sgtcoolguy: Should be isWriteable

Source/TitaniumKit/src/Blob.cpp
```diff
@@ -130,6 +130,11 @@ namespace Titanium
 		data_.insert(data_.end(), b.begin(), b.end());
 	}
 
+	void Blob::release() TITANIUM_NOEXCEPT
+	{
+		data_.clear();
```
sgtcoolguy: So this clears the vector, but doesn't actually free the memory. We should probably attempt to free the memory here if we can, right?

Source/TitaniumKit/src/Filesystem/FileStream.cpp
```diff
@@ -31,7 +31,7 @@ namespace Titanium
 			return 0;
 		}
 
-		bool FileStream::isWriteable() TITANIUM_NOEXCEPT
+		bool FileStream::isWritable() TITANIUM_NOEXCEPT
```
sgtcoolguy: isWriteable

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -25,7 +25,7 @@ namespace Titanium
 
 	void IOStream::readAsync(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length, const std::function<void(const std::int32_t&)>& callback) TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_WARN("IOStream::readAsync: Unimplemented");
+		callback(read(buffer, offset, length));
```
sgtcoolguy: Is this actually running async? It doesn't appear to me that it is in any way... We're just calling a bound function with the result of synchronously running the read operation. Wouldn't we need to spin off some thread to read and then call the callback when it's done?

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -36,19 +36,17 @@ namespace Titanium
 
 	void IOStream::writeAsync(const std::shared_ptr<Buffer>& buffer, const std::uint32_t& offset, const std::uint32_t& length, const std::function<void(const std::int32_t&)>& callback) TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_WARN("IOStream::writeAsync: Unimplemented");
+		callback(write(buffer, offset, length));
```
sgtcoolguy: same as above. I don't think this is actually going to be async.

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -36,19 +36,17 @@ namespace Titanium
...
 	}
 
-	bool IOStream::isWriteable() TITANIUM_NOEXCEPT
+	bool IOStream::isWritable() TITANIUM_NOEXCEPT
```
sgtcoolguy: isWriteable

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -36,19 +36,17 @@ namespace Titanium
...
 	{
-		TITANIUM_LOG_WARN("IOStream::isWriteable: Unimplemented");
-		return false;
+		return modes__.find(Filesystem::MODE::READ) == modes__.end();
```
sgtcoolguy: This is basically saying it's writeable if READ wasn't specified. But I think a user could do (READ || WRITE) and we'd want to accept that? I'm not sure. The API I see in docs takes one mode, but we have a set here. I think we want to search for WRITE or APPEND here.

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -63,32 +61,38 @@ namespace Titanium
 
 		TITANIUM_ADD_FUNCTION(IOStream, read);
 		TITANIUM_ADD_FUNCTION(IOStream, write);
-		TITANIUM_ADD_FUNCTION(IOStream, isWriteable);
+		TITANIUM_ADD_FUNCTION(IOStream, isWritable);
```
sgtcoolguy: isWriteable

Source/TitaniumKit/src/IOStream.cpp
```diff
@@ -63,32 +61,38 @@ namespace Titanium
...
+		const auto buffer = buffer_object.GetPrivate<Buffer>();
+		if (buffer == nullptr) {
+			HAL::detail::ThrowRuntimeError("Titanium::IOStream::read", "Titanium::IOStream::read: Invalid arguments");
+		}
+		ENSURE_OPTIONAL_UINT_AT_INDEX(length, 2, buffer->get_length());
+		return get_context().CreateNumber(write(buffer, offset, length));
 	}
 
-	TITANIUM_FUNCTION(IOStream, isWriteable)
+	TITANIUM_FUNCTION(IOStream, isWritable)
```
sgtcoolguy: isWriteable

sgtcoolguy:
@infosia OK to merge after a few small changes - most notably the method name is 'isWriteable', not 'isWritable'

infosia:
@sgtcoolguy Yeah, I should have mentioned, the doc says `isWriteable` but the Android & iOS implementation actually uses `isWritable`. For instance: [TiStreamProxy.m#L140](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiStreamProxy.m#L140) & [BlobStreamProxy.java#L108](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/ti/modules/titanium/stream/BlobStreamProxy.java#L108)

Source/TitaniumKit/src/IOStream.cpp
infosia: Yes that's a good point. I would make this sync for default behavior because not all subclass needs actually a async operation...I consider that creating new thread for getting `std::vector` values sounds too much  (BlobStream, BuferStream) because it should be done pretty immediately. Only FileStream and Network stream needs async operation, actually.

Source/TitaniumKit/src/Blob.cpp
infosia: Do we have a better way to free the memory of `std::vector`?

Source/TitaniumKit/src/IOStream.cpp
infosia: ok, sounds reasonable.

Source/TitaniumKit/src/IOStream.cpp
yuchi: No no! The stream APIs must be async. Otherwise you'll get no guarantee over when things are called. NEVER mix sync with async behavior.

infosia: @yuchi My understanding is, that's exactly what async meant. I mean async has actually no guarantee over when callback is called, by definition.

infosia: @yuchi  Let me explain by code:

```javascript
function callback() {
    Ti.API.info('callback start');
}
function test() {
    Ti.API.info('readAsync start');
    Ti.Stream.readAsync(...,callback);
    Ti.API.info('readAsync end');
}
```

So, what would you expect for the output by this example?

Do you think it always goes following?

```
    Ti.API.info('readAsync start');
    Ti.API.info('readAsync end');
    Ti.API.info('callback start');
```

Actually there's no guarantee when callback is called. It could be like following if callback is done *before* *test()* is done. I believe this is exactly what *async* method works.

```
    Ti.API.info('readAsync start');
    Ti.API.info('callback start');
    Ti.API.info('readAsync end');
```

So, there's no guarantee when async callback gets called by definition. In regard to that, your program *should not* depends on the timing when callback gets called. 

/cc @sgtcoolguy

yuchi: I don’t think that’s what *async* means regarding the JS land.

AFAIK when something is async that means that you always have the current tick finishing before anything async actually happens.

I don‘t have pointers to official papers.

infosia: That's a good discussion point. I looked into the [ECMA-262 6th](http://www.ecma-international.org/publications/standards/Ecma-262.htm) specification but there's no definition about how async should work in order as far as I see. So I think there's no definition by the language spec.

So the next point is *how Titanium should work*. Does Titanium guarantee that async callback is not be called before next statement is executed? Let me explain by code:

```javascript
function A() { }
function B() {}
function C() {}
function callback() {
    C();
}
A();
Ti.Stream.readAll(..., callback);
B();
```

I this case, does Titanium *guarantees* it's executed like this?

```
A() -> B() -> C()
```

And, does Titanium *guarantees* following pattern never happen?
```
A() ->  C() -> B()
```

I guess it's actually a implementation dependent for both JavaScript VM and Titanium (C++/Java) wrapper, because there's no definition about how JavaScript VM should work in order for async function, and also Titanium doesn't have exact spec about how async function works.

yuchi: I’d bet my ~~balls~~ facial hair on the fact that in Node everything labelled as async will always act as such.

Found something interesting on [Bluebird *sync* build](https://github.com/petkaantonov/bluebird#what-is-the-sync-build).

yuchi: Highlight (emphasis is mine)

> This should pressure the CPU slightly less and thus the sync build should perform better. Indeed it does, but only marginally. The biggest performance boosts are from writing efficient Javascript, not from **compromising determinism**.

infosia: So the problem is there's no actual specification on how async should work. I guess this is a very common problem in the JavaScript world, and that's one of the reason why *Promises* was invented. I would suggest use of Promises anytime you need a guarantee, because Promise spec *promises* callback is done after next statement is executed.

```javascript
var promise = new Promise(function(resolve){
    resolve();
});
console.log('start');
promise.then(function(){
    console.log('done');
});
console.log('end');

// start 
// end
// done
```

infosia: I'm now trying to find a way to ensure Promise-like behavior for this. may be in subsequent PR.

yuchi: Great!

PR #430, opened 2015-09-17
Ti.Contacts

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-17914
https://jira.appcelerator.org/browse/TIMOB-19484
https://jira.appcelerator.org/browse/TIMOB-19483


Source/Ti/src/Contacts.cpp
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	void ContactsModule::JSExportInitialize() {
+		JSExport<ContactsModule>::SetClassVersion(1);
+		JSExport<ContactsModule>::SetParent(JSExport<Titanium::ContactsModule>::Class());
+	}
+
+#if (WINVER >= 0x0A00)
+	// Windows 10+ API!
+	std::shared_ptr<Titanium::Contacts::Group> listToGroup(const JSContext& context, Windows::ApplicationModel::Contacts::ContactList^ contact_list) TITANIUM_NOEXCEPT
```
infosia: I would mark this `static` because it's only used internally.

Source/Ti/src/Contacts.cpp
```diff
@@ -0,0 +1,355 @@
...
+		auto Group = context.CreateObject(JSExport<TitaniumWindows::Contacts::Group>::Class());
+		auto group = Group.CallAsConstructor();
+		auto group_ptr = group.GetPrivate<TitaniumWindows::Contacts::Group>();
+		group_ptr->construct(contact_list);
+
+		return group.GetPrivate<Titanium::Contacts::Group>();
+	}
+#endif
+
+	std::shared_ptr<Titanium::Contacts::Person> contactToPerson(const JSContext& context, Windows::ApplicationModel::Contacts::Contact^ contact) TITANIUM_NOEXCEPT
```
infosia: `static`, and should be enclosed by `#if (WINVER >= 0x0A00)`?

infosia:
How do I create a NMocha/NG project for Windows 10? I was using `node generate_project.js -p WindowsStore new NMocha` on Windows 10 but it gives me a VS project for Windows 8.1.

vs

infosia:
Looks good to merge, once internal functions noted above are declared as `static`.

sgtcoolguy:
The generate_project.js script is hard-coded to do Windows 8.1. We could add an arg to set the target version, but honestly I'm not sure if it'd just work or not.  Here's where it's hard-coded to 8.1: https://github.com/appcelerator/titanium_mobile_windows/blob/master/Tools/Scripts/generate_project.js#L91

I'm guessing the generator would also have to change: https://github.com/appcelerator/titanium_mobile_windows/blob/master/Tools/Scripts/generate_project.js#L82

infosia:
Hmm...I tried `DCMAKE_SYSTEM_VERSION=10.0` but cmake says "Visual Studio 12 2013" generator only supports "8.0" or "8.1". I don't have VS 2015 installed on my Windows 10 because I assumed it doesn't work well with cmake yet...

sgtcoolguy:
Merged manually

PR #432, opened 2015-09-21
Refactor & Allow mixing Windows UI component for View subclass

infosia:
Refactoring #419, which basically overrides `View.add` and `Window.add` only. This PR moves these native hook logic onto `WindowsViewLayoutDelegate` so that every View subclass supports mixing Windows UI components. Moving to layout delegate is also good for avoiding duplicate code across View subclasses.

- Refactor & Allow mixing Windows UI component for View subclass
- Move View::registerNativeUIWrapHook to layout delegate

```javascript
var win,
    TextBlock = require('Windows.UI.Xaml.Controls.TextBlock'),
	text = new TextBlock();

text.Text = "Hello, world!";
text.FontSize = 60;

win = Ti.UI.createWindow({});
win.add(text);
win.open();
```

```javascript
var Window = require('Windows.UI.Xaml.Window'),
    Canvas = require('Windows.UI.Xaml.Controls.Canvas'),
    window = Window.Current,
    canvas = new Canvas(),
    text;

text = Ti.UI.createLabel({
    text: "Hello, world!",
    font: {
        fontSize: 60
    }
});
canvas.Children.Append(text);

window.Content = canvas;
window.Activate();
```

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -17,6 +17,7 @@
 #include <cctype>
 #include <collection.h>
 
+#include "TitaniumWindows/UI/View.hpp"
 #include "TitaniumWindows/Utility.hpp"
 
 #define _USE_MATH_DEFINES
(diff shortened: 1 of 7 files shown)


Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -868,14 +897,14 @@ namespace TitaniumWindows
 			}
 
 			if (!is_scrollview__ && !std::isnan(component->Width) && !std::isnan(component->Height)) {
```
infosia: Following lines are just for a fix for line-break mismatch.

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -520,9 +549,9 @@ namespace TitaniumWindows
 				updateBackgroundImageSize();
 
 				// Update image source
```
infosia: Following lines are just for a fix for line-break mismatch.

PR #438, opened 2015-09-24
Fix TIMOB-19517 Windows: Native stub generation fails on 'float32' when requiring StackPanel

sgtcoolguy:
Handle all primitives found in metabase wherever we need to filter them out/convert them.

https://jira.appcelerator.org/browse/TIMOB-19517

Basically we need to avoid including primitives in the dependency tree for generating wrappers, and we  need to make sure we can convert primitives back and forth from JS.

### DIFF
--- a/cli/lib/stub.js
+++ b/cli/lib/stub.js
@@ -150,8 +150,22 @@ function addNewTypes(rawTypeName, existingTypes) {
 		types = expandTypes(rawTypeName);
 		for (var i = 0; i < types.length; i++) {
 			// skip template names, primitives
-			if (types[i].indexOf("!") == 0 || types[i] == 'bool' ||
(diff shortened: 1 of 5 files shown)


sgtcoolguy:
test this please.

sgtcoolguy:
test this please.

sgtcoolguy:
I didn't actually find any instances of "short", "single", "long", "long long", or "float" in the metabase, so didn't include those.

jhaynie:
ok, gotcha. weird given that it's a C based language.

sgtcoolguy:
Bizarre. The app crashes on startup on the build machine, but I'm not really sure why my changes would have any affect on that. I can't reproduce locally using the exact same script. Guess I'll just kick off a rebuild and hope it works.

sgtcoolguy:
test this please.

PR #441, opened 2015-09-29
Add VS 2015 and Windows 10 SDK option for CLI

infosia:
[TIMOB-19576](https://jira.appcelerator.org/browse/TIMOB-19576)

- Add `--vs-target` option to specify Visual Studio version
 - 12.0  ... Visual Studio 2013
 - 14.0 ... Visual Studio 2015
- Add Windows 10 sdk support  (`--wp-sdk 10`)

```
ti build --platform windows --vs-target 12.0 --target wp-emulator --device-id 8-1-1 --win-publisher-id 13AXB724-65X2-4X30-8994-C79399EDXX00 --wp-sdk 8.1
```

Note that although CLI accepts targeting to Windows 10 (`--wp-sdk 10`), it's failing for now because cmake doesn't support that yet. It's implemented for future release.

### DIFF
--- a/cli/commands/_build/initialize.js
+++ b/cli/commands/_build/initialize.js
@@ -49,6 +49,12 @@ function initialize(next) {
 	this.arch = this.cmakeArch == 'Win32' ? 'x86' : this.cmakeArch;
 	this.cmakeTarget = this.cmakePlatformAbbrev + '.' + this.arch;
 
+	if (argv['vs-target']) {
+		this.cmakeGeneratorName = argv['vs-target'] == '14' ? 'Visual Studio 14 2015' : 'Visual Studio 12 2013';
+	} else {
+		this.cmakeGeneratorName = 'Visual Studio 12 2013';
+	}
+
 	// directories
 	this.outputDir = argv['output-dir'] ? appc.fs.resolvePath(argv['output-dir']) : null;
 	this.wsCert = argv['ws-cert'] ? argv['ws-cert'] : null;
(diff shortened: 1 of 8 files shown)


cli/commands/_build/initialize.js
```diff
@@ -49,6 +49,12 @@ function initialize(next) {
 	this.arch = this.cmakeArch == 'Win32' ? 'x86' : this.cmakeArch;
 	this.cmakeTarget = this.cmakePlatformAbbrev + '.' + this.arch;
 
+	if (argv['vs-target']) {
+		this.cmakeGeneratorName = argv['vs-target'] == '14' ? 'Visual Studio 14 2015' : 'Visual Studio 12 2013';
```
sgtcoolguy: Do we really need this arg option? Can you even have multiple Visual Studio versions installed? (i.e. can we have 2013 and 2015 installed?) Even if you could, wouldn't you want to avoid having the user choose this and determine the right VS generator to use based on the chosen SDK?

infosia: Yes we can have both 2013 and 2015 installed. But right, we should be able to detect it.

infosia:
Implemented auto-detect for Visual Studio when user doesn't specify it. Implemented prompt for `vs-target` when user specifies wrong target.

```
Please select a valid vs-target value:
 1)  12.0
 2)  14.0
Select a value by number or name:
```

infosia:
@sgtcoolguy `windowslib` is good to go, would you bump up the version? Thanks.

PR #444, opened 2015-09-30
[TIMOB-19644] Fix: CLI build error is ignored

infosia:
[TIMOB-19644](https://jira.appcelerator.org/browse/TIMOB-19644) Related: [TIMOB-19643](https://jira.appcelerator.org/browse/TIMOB-19643)

When there's no DefaultIcon.png, CLI shows critical error on build but it does not stop build process.


garymathews:
CR: **PASS**
FR: **PASS**

PR #446, opened 2015-10-02
[TIMOB-19650] Skip generation of existing icons

garymathews:
- Skip the generation of existing icons

###### TEST CASE
- Place a required icon into ```<project>/app/assets/windows/```, for example ```Square150x150Logo.png```
- This icon should not be overwritten by the ```DefaultIcon.png``` after the project has been generated using ```appc run```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19650)

### DIFF
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -292,6 +292,16 @@ function copyResources(next) {
 
 			// TODO: Generate SplashScreen.scale-100.png?
 		];
+
+		// do not generate existing icons
+		for (i in missingIcons) {
+			var icon = missingIcons[i];
+			if (fs.existsSync(icon.file)) {
+				_t.logger.debug(__('%s already exists, skipping...', icon.file));
+				missingIcons.splice(i, 1);
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


infosia:
@sgtcoolguy , @garymathews , I'm not very sure about Titanium project spec but what about the icons in `<project>/platform/windows` and `<project>/`? For example don't we need to check `<project>/Square150x150Logo.png` and `<project>/platform/windows/Square150x150Logo.png` too?

garymathews:
@infosia From my testing it seems any icons/assets placed in those directories are ignored. They don't get copied over to the generated project. Only icons in the assets folder are copied.

infosia:
ok, we can tackle it in another ticket, I'll create new ticket if necessary. Merging this now...

sgtcoolguy:
@garymathews We definitely need another ticket/fix. We copy over the windows specific assets after any "general "assets: https://github.com/appcelerator/titanium_mobile_windows/blob/master/cli/commands/_build/copy.js#L340

PR #447, opened 2015-10-02
[TIMOB-18531] Set default value for Windows Publisher ID prompt

garymathews:
- Set a default value for the ```Windows Publisher ID``` prompt
- Do not set a default prompt value if ```--deploy-type``` is ```production```

###### TEST CASE
```
appc run -p windows
# builds project without prompting for 'Publisher ID'
```
```
appc run -p windows --deploy-type production
What is your Windows Publisher ID?:
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-18531)

### DIFF
--- a/cli/commands/_build/config/winPublisherId.js
+++ b/cli/commands/_build/config/winPublisherId.js
@@ -20,7 +20,7 @@ module.exports = function configOptionWindowsPublisherID(order) {
 
 	return {
 		abbr: 'I',
-		default: this.config.get('windows.publisherId'),
+		default: this.cli.argv['deploy-type'] !== 'production' ? '00000000-0000-1000-8000-000000000000' : this.config.get('windows.publisherId'),
 		desc: __('your Windows publisher ID, obtained from %s', 'https://dev.windows.com/en-us/Account/Management'.cyan),
 		hint: __('id'),
 		order: order,
@@ -29,8 +29,7 @@ module.exports = function configOptionWindowsPublisherID(order) {
 				promptLabel: __('What is your __Windows Publisher ID__?'),
 				validate: validate
 			}));
-		},
-		required: true,
+		}.bind(this),
 		validate: validate
 	};
 };


cli/commands/_build/config/winPublisherId.js
```diff
@@ -25,10 +27,14 @@ module.exports = function configOptionWindowsPublisherID(order) {
...
+			var field = {
 				promptLabel: __('What is your __Windows Publisher ID__?'),
 				validate: validate
-			}));
+			};
+			if (_t.cli.argv['deploy-type'] != 'production') {
+				field['default'] = '00000000-0000-1000-8000-000000000000';
+			}
+			callback(fields.text(field));
 		},
```
cb1kenobi: You don't have to do the `var _t = this`. You can just `.bind(this)` to the anon function.

cli/commands/_build/config/winPublisherId.js
```diff
@@ -25,10 +27,14 @@ module.exports = function configOptionWindowsPublisherID(order) {
...
 		order: order,
 		prompt: function (callback) {
-			callback(fields.text({
+			var field = {
 				promptLabel: __('What is your __Windows Publisher ID__?'),
 				validate: validate
-			}));
+			};
+			if (_t.cli.argv['deploy-type'] != 'production') {
+				field['default'] = '00000000-0000-1000-8000-000000000000';
```
cb1kenobi: Instead of creating a `field` var, you could simply do:

```javascript
callback(field.text({
    promptLabel: __('What is your __Windows Publisher ID__?'),
    validate: validate,
    default: this.cli.argv['deploy-type'] !== 'production' ? '00000000-0000-1000-8000-000000000000' : undefined
});
```

cb1kenobi: For production builds, why not default to `this.config.get('windows.phone.publisherGuid')`?

infosia:
I don't see any "fail" on Jenkins test.. test this please, again. :recycle:

infosia:
@garymathews What about https://github.com/appcelerator/titanium_mobile_windows/pull/447#discussion_r41060407 ? This sounds making sense to me.

infosia:
test this please

infosia:
Hmm. Jenkins test seems unstable. Test this please, again...

```
04:35:57 ERROR: Connection was broken: java.io.IOException: Connection aborted: org.jenkinsci.remoting.nio.NioChannelHub$MonoNioTransport@603a4010[name=Windows SDK Hyper-V]
```

infosia:
test this please

infosia:
test this please

cb1kenobi:
@infosia I just did a CR. Either you or @sgtcoolguy will need to do the FR.

sgtcoolguy:
@garymathews Dumb question, but it looks like this only gives you a default prompt value for non-production builds? I think we'd want to actually _not _prompt and spit out a warning saying that you don't have a publisher id set so we're using a default and please fill one in before you do production/store builds - and then just use the default without prompting so that Studio users may see the warning but can still run without setting the id. Basically, giving them a default prompt value is nice but we really need to warn and avoid prompting altogether if possible so we can just let users get started without having to fill that in before the first run.

cb1kenobi:
@sgtcoolguy For consistency sake, we don't "warn" and try to avoid prompting. Prompting is a valid response. The default prompt value makes some sense.

In the case of the publisher id, perhaps what we should do is move the publisher id to the tiapp.xml and not a command line option? I doubt the publisher id is going to change per project.

sgtcoolguy:
@cb1kenobi The publisher ID is likely to be constant for an app, but is actually more global than that. It's most likely going to be constant across apps.

The comparison was made to using a developer keypair for Android. In that case we don't warn and we don't prompt, we just use a dev one for non-production builds.

My point is that this was intended to be an improvement to avoid having to specify the publisher id to get started. While prompting is valid for CLI users, it means we don't get that improvement for Studio users - which is going to be our overwhelming user base here.

We don't _have_ to warn, but it may be useful to alert the user what's going on and they'll need the value eventually. An alternate approach is to not prompt at all here and mark the value as only required for production builds - and just use the placeholder value behind the scenes if it hasn't been specified.

cb1kenobi:
@sgtcoolguy I like that idea. The publisher id isn't really needed in dev builds, so all you guys need to do is remove the `required: true` for the publisher id option, then in the `target` option's callback, set the publisher id to required. Here's how the iOS build does it: https://github.com/appcelerator/titanium_mobile/blob/master/iphone/cli/commands/_build.js#L1008-L1050.

sgtcoolguy:
ping @garymathews

garymathews:
@sgtcoolguy, @cb1kenobi I've updated the PR so it now __does not__ display the prompt for development builds.

cb1kenobi:
LGTM

PR #445, opened 2015-10-02
CMake  Windows 10 RTM Support - WIP

infosia:
[TIMOB-19651](https://jira.appcelerator.org/browse/TIMOB-19651)

Update to CMake 3.3.2 with Windows 10 Support

```
cmake -G "Visual Studio 14 2015" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 ...
```

PR #448, opened 2015-10-05
Update CMake to Win10 feature branch from MS

sgtcoolguy:
Follow-on to #445 

Built the win10 branch version of cmake locally using cmake and VS 2013. Build worked fine for me, but the mocha tests are crashing intermittently (seems to be the case on the build server too). Needs to be tested on VS2015/Win10 too.


sgtcoolguy:
test this please.

infosia:
Thanks, but what's bad thing I've ran into is that CMake should be configured using "latest cmake" which is built on Windows 10, [according to Microsoft](https://github.com/Microsoft/CMake/issues/12). I'll take care of this, closing this PR for now. :+1:

PR #450, opened 2015-10-07
Update HAL to include u16String removal PR

sgtcoolguy:


sgtcoolguy:
@infosia Looks like this caused a test failure about encoding a string... http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_prs/1420/testReport/junit/(root)/codec/testEncodeString/

infosia:
ok, debugging... I will include this in PR #449. Closing.

PR #464, opened 2015-11-03
Implement missing TitaniumKit View properties

infosia:
[TIMOB-19837](https://jira.appcelerator.org/browse/TIMOB-19837)

- Add missing View properties
- Add TitaniumKit  Ti.3DMatrix

Note that I've added `Ti.3DMatrix` but there's unimplemented functions partially. It's added just to make `Ti.View` compiled for now.


infosia:
test this please

garymathews:
@infosia The tests pass (I tested it locally), it fails because the Jenkins server is trying to run the tests on a Windows 10 Emulator (8-1-1) when it should use (8-1-6) OR we should discard the Windows 10 Emulators from the list?

infosia:
test this please

Source/TitaniumKit/include/Titanium/UI/View.hpp
```diff
@@ -33,6 +37,85 @@ namespace Titanium
...
+			  @discussion A string describing the value (if any) of the view for the device's accessibility service.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::string, accessibilityValue);
+
+			/*!
+			  @property
+			  @abstract clipMode
+			  @discussion View's clipping behavior.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, clipMode);
```
sgtcoolguy: enum?

Source/TitaniumKit/include/Titanium/UI/View.hpp
```diff
@@ -33,6 +37,85 @@ namespace Titanium
...
+			  @discussion Whether view should be focusable while navigating with the trackball.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(bool, focusable);
+
+			/*!
+			  @property
+			  @abstract softKeyboardOnFocus
+			  @discussion Determines keyboard behavior when this view is focused.
+			*/
+			TITANIUM_PROPERTY_IMPL_DEF(std::uint32_t, softKeyboardOnFocus);
```
sgtcoolguy: enum?

Source/TitaniumKit/include/Titanium/UI/View.hpp
infosia: No, this is iOS specific property which we don't hold it as enum in TitaniumKit. We are usually do like this to define platform-specific enum property like "Titanium::UI::iOS::CLIP_MODE_DEFAULT" here.

Source/TitaniumKit/include/Titanium/UI/View.hpp
infosia: No, this is Android specific property which we don't hold it as enum in TitaniumKit. We are usually do like this to define platform-specific enum property like "Titanium::UI::Android::SOFT_KEYBOARD_DEFAULT_ON_FOCUS" here.

PR #470, opened 2015-11-09
Fix: DefaultLanguage property should be set (to "en") in manifest

infosia:
Fix for [TIMOB-19138](https://jira.appcelerator.org/browse/TIMOB-19138)

Set `DefaultLanguage` to `en` to suppress compiler warning.
Note: Would be nice to have platform-specific default language property, something like `ti.windows.defaultLanguage` FYI: [TIMOB-16356](https://jira.appcelerator.org/browse/TIMOB-16356)


garymathews:
CR: **PASS**
FR: **PASS**

PR #475, opened 2015-11-13
[TIMOB-19487] Fix missing icon generation

garymathews:
- Removed template icons overriding missing icons
- Fixed loop iteration that was affected by removed items
- Checks for icons in platform assets folder and project root folder
- Overwrites modified icons
- Added debug logging

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19487)


cli/commands/_build/copy.js
```diff
@@ -320,13 +320,22 @@ function copyResources(next) {
...
-			var iconInProject = path.join(this.projectDir, path.basename(icon.file));
-			if (fs.existsSync(iconInProject)) {
-				copyFile.call(_t, iconInProject, icon.file);
+		// if the app icon does not exist then check if it exists in the project root
+		// if it does not exist in project root then generate the missing icon
+		for (var i = missingIcons.length - 1; i >= 0; i--) {
+			var icon = missingIcons[i],
+				projectRootIcon = path.join(this.projectDir, path.basename(icon.file));
+			if (fs.existsSync(icon.file)) {
+				_t.logger.debug(__('%s already exists, skipping...', icon.file));
```
cb1kenobi: Inconsistent use of `this`. You should probably just use `this` instead of `_t` for readability.

garymathews:
@cb1kenobi Thanks, amended.

cli/commands/_build/copy.js
```diff
@@ -320,13 +320,22 @@ function copyResources(next) {
...
-			var iconInProject = path.join(this.projectDir, path.basename(icon.file));
-			if (fs.existsSync(iconInProject)) {
-				copyFile.call(_t, iconInProject, icon.file);
+		// if the app icon does not exist then check if it exists in the project root
+		// if it does not exist in project root then generate the missing icon
+		for (var i = missingIcons.length - 1; i >= 0; i--) {
+			var icon = missingIcons[i],
+				projectRootIcon = path.join(this.projectDir, path.basename(icon.file));
+			if (fs.existsSync(icon.file)) {
+				this.logger.debug(__('%s already exists, skipping...', icon.file));
```
infosia: I don't agree with this because this never happens now, because in this case `icon.file` points to the icon in the temporary build directory, and all assets are deleted from temporary directory on every build. This could cause problem in the future when we implement incremental build, because always-skipping this means whenever you change the icons in your project root, they are always ignored once it is copied into temporary build directory, right?

cli/commands/_build/copy.js
```diff
@@ -320,13 +320,22 @@ function copyResources(next) {
...
-		for (i in missingIcons) {
-			var icon = missingIcons[i];
-			var iconInProject = path.join(this.projectDir, path.basename(icon.file));
-			if (fs.existsSync(iconInProject)) {
-				copyFile.call(_t, iconInProject, icon.file);
+		// if the app icon does not exist then check if it exists in the project root
+		// if it does not exist in project root then generate the missing icon
+		for (var i = missingIcons.length - 1; i >= 0; i--) {
+			var icon = missingIcons[i],
+				projectRootIcon = path.join(this.projectDir, path.basename(icon.file));
```
infosia: I realized [TIMOB-19487](https://jira.appcelerator.org/browse/TIMOB-19487) requires support for "platform-specific" assets directory. I think we want to check assets under ` <project-dir>/platforms/windows/` too.

cli/commands/_build/copy.js
infosia: > We should overwrite any existing assets after every build?

I think we should, only when the asset in the project directory is updated. 

How about this:

1. Check if asset exists in the temporary build directory
    1. If not, and asset is found in the project root, copy it
    2. If not, and asset is not found in the project root, generate new one & copy it 
2. If exists, check if it's changed (timestamp or md5?)
    1. If it's not changed, skip it
    2. If it's changed, and asset is found in the project root, copy it
    3. If it's changed, and asset is not found in the project root, generate new one & copy it

cli/commands/_build/copy.js
infosia: ok, looks like `<project-dir>/platforms/PLATFORM/DefaultIcon.png` is not supported yet in the titanium CLI. I think we want to do this in the different ticket & next release.

infosia: Just confirmed we don't need to support platform-specific DefaultIcon. (See comment in [TIMOB-19487](https://jira.appcelerator.org/browse/TIMOB-19487))

garymathews:
Ok so I've updated the PR to do the following:

1. Check if the icon exists in the ```app/assets/windows``` folder
  1. If it exists, check if it exists in the build folder.
  2. Overwrite the build folder icon if it's md5 hash is different.
  3. Otherwise, if it doesn't exist in the build folder copy it over to the build folder.

2. If the icon does NOT exist in the ```app/assets/windows``` folder, check for the icon in the project root directory.
  1. If it exists, check if it exists in the build folder.
  2. Overwrite the build folder icon if it's md5 hash is different.
  3. Otherwise, if it doesn't exist in the build folder copy it over to the build folder.

infosia:
ok looks good. Will merge this into 5_1_x soon too.

PR #478, opened 2015-11-19
Update build.js script to allow setting target SDK version and msbuild

sgtcoolguy:
Related to [TIMOB-19704](https://jira.appcelerator.org/browse/TIMOB-19704) Set up Win 10 Build

Rewrite the build.js script to use commander for arg/option parsing, try to get it set up so we can target 8.1 or 10.0 Win SDK, with MSBuild 12.0 or 14.0 via args


sgtcoolguy:
Merged manually.

PR #480, opened 2015-11-20
Hack setup.js script so we could pass override source URLs for gtest,…

sgtcoolguy:
… boost or JSC on command line

Related to https://jira.appcelerator.org/browse/TIMOB-19704


sgtcoolguy:
Manually merged.

PR #481, opened 2015-11-23
New YAML API doc generator

infosia:
Fix for [TIMOB-20022](https://jira.appcelerator.org/browse/TIMOB-20022)

- Add new YAML API doc generator
- Explicitly declares "unimplemented" methods/properties

To generate correct API docs,  I'm planning to take a different path for our YAML generator. 

- TItaniumKit module class should declare its name in Titanium like `Titanium.UI.View`, because there's some difference between C++ class name and Titanium module name. (i.e. `Titanium.UI.2DMatrix` vs `Titanium::UI::TwoDMatrix`)
- Titanium platform implementation should declare "what's unimplemented" explicitly, because there's no way to auto-detect what's not supported, by just looking into C++ method signature.

To achieve this, I combined doxygen grouping comments `@ingroup` and empty C++ macro (`TITANIUM_FUNCTION_UNIMPLEMENTED` and `TITANIUM_PROPERTY_UNIMPLEMENTED`) which is only used for documentation. For example:

```c++
/*
  @class FilesystemModule
  @ingroup Titanium.Filesystem
*/
class TITANIUMWINDOWS_FILESYSTEM_EXPORT FilesystemModule final ... {
public:
  TITANIUM_FUNCTION_UNIMPLEMENTED(openStream); 
  TITANIUM_PROPERTY_UNIMPLEMENTED(....); 
...
};
```


infosia:
Ready for review, `node ti_win_yaml.js` should work like before, with removing unimplemented APIs.

garymathews:
CR: **PASS**
FR: **PASS**

PR #488, opened 2015-11-26
Disable AudioCategory on Windows 10

infosia:
Fix compiler warning on Windows 10.

Although it's not quite clear which values are deprecated, VS 2015 compiler produces warning on use of `AudioCategory::BackgroundCapableMedia`. 

The document from MS says:

[MediaElement class](https://msdn.microsoft.com/library/windows/apps/windows.ui.xaml.controls.mediaelement.aspx)
> To let audio continue to play when your app is in the background, set the AudioCategory property to BackgroundCapableMedia. This also requires declaring the "Audio" background task capability in the app manifest. **These values are deprecated in Windows 10**.

But what does "*Those values*" exactly mean?

In their API document, MS says it's *supported* in Windows 10. See [AudioCategory enumeration](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.media.audiocategory.aspx) and [AudioCategory property](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.controls.mediaelement.audiocategory.aspx).

Mystery.

### DIFF
--- a/Source/Media/src/Media/AudioPlayer.cpp
+++ b/Source/Media/src/Media/AudioPlayer.cpp
@@ -38,7 +38,11 @@ namespace TitaniumWindows
 			Titanium::Media::AudioPlayer::postCallAsConstructor(js_context, arguments);
 
 			player__ = ref new Windows::UI::Xaml::Controls::MediaElement();
+#pragma warning(push)
+#pragma warning(disable : 4973)
+			// Note: BackgroundCapableMedia is deprecated in Windows 10
 			player__->AudioCategory = AudioCategory::BackgroundCapableMedia;
+#pragma warning(pop)
 			player__->AreTransportControlsEnabled = true;
 			player__->Visibility = Windows::UI::Xaml::Visibility::Collapsed; // Hide UI
 			player__->AutoPlay = false;


garymathews:
Is this just deprecated for Windows 10 Store apps or both Windows 10 Store and Phone?

infosia:
AFIK when you build for Windows 10, there's no "Windows 10 Phone" target. There's only Windows 10 Store app when you create Universal App.

sgtcoolguy:
The feature is supported in Windows 10, but is deprecated and there's a document on using background audio (I think using background tasks). I don't think we want to just remove the line in Win10+. We probably want to disable the deprecation warning on Windows 10 for this specific instance until/unless we use the new API/flow for doing the background audio.

I'd suggest we turn off the compiler warning for this file (or stifle it in some way) on Win 10 compiles, and open a ticket for replacing the mechanism to use the new API in the future. See https://msdn.microsoft.com/en-us/library/windows/apps/mt282140.aspx and https://social.msdn.microsoft.com/Forums/windowsapps/en-US/46432468-b968-4570-a0ce-48babe838f2d/uwpbackground-audio-with-mft-in-windows-10

infosia:
@sgtcoolguy ok, updated to disable warning. New ticket is [TIMOB-20080](https://jira.appcelerator.org/browse/TIMOB-20080).

PR #490, opened 2015-12-02
[TIMOB-19929][TIMOB-19930] Implement Ti.Filesystem.File.append() and fix Ti.Filesystem.File.deleteDirectory()

garymathews:
- Implemented ```Titanium.Filesystem.File.append()``` functionality into TitaniumKit
- Implemented ```Titanium.Filesystem.File.deleteDirectory()``` functionality into TitaniumKit
- Fixed recursive deletion of files and folders for ```Titanium.Filesystem.File.deleteDirectory()```
- Added test cases

[TIMOB-19929](https://jira.appcelerator.org/browse/TIMOB-19929)
[TIMOB-19930](https://jira.appcelerator.org/browse/TIMOB-19930)

### DIFF
--- a/Source/Titanium/src/File.cpp
+++ b/Source/Titanium/src/File.cpp
@@ -213,11 +213,8 @@ namespace TitaniumWindows
 			return path_;
 		}
 
-		std::shared_ptr<Titanium::Filesystem::File> File::get_parent() const TITANIUM_NOEXCEPT
+		std::shared_ptr<Titanium::Filesystem::File> File::get_file_ptr(const std::string& path) const TITANIUM_NOEXCEPT
 		{
-			const std::string path = path_;
-			const std::string parent = path.substr(0, path.find_last_of("\\"));
-
 			JSValue Titanium_property = get_context().get_global_object().GetProperty("Titanium");
 			TITANIUM_ASSERT(Titanium_property.IsObject());  // precondition
 			JSObject Titanium = static_cast<JSObject>(Titanium_property);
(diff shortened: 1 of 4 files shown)


Source/Titanium/src/File.cpp
```diff
@@ -230,7 +227,15 @@ namespace TitaniumWindows
...
 			JSObject File = static_cast<JSObject>(File_property);
 
-			return File.CallAsConstructor(parent).GetPrivate<Titanium::Filesystem::File>();
+			return File.CallAsConstructor(path).GetPrivate<Titanium::Filesystem::File>();
+		}
+
+		std::shared_ptr<Titanium::Filesystem::File> File::get_parent() const TITANIUM_NOEXCEPT
+		{
+			const std::string path = path_;
+			const std::string parent = path.substr(0, path.find_last_of("\\"));
```
sgtcoolguy: We shouldn't be using a hard-coded separator here. There's a method to get the file separator for the platform (Ti.Filesystem.separator), and we should be using that.

sgtcoolguy: Oops, just noticed this is in the Windows-specific subclass. So scratch that. But it does make me think then that there's a way to implement get_parent() at the TitaniumKit level using some combination of Ti.File.get_nativePath(), Ti.Filesystem.separator, and Ti.Filesystem.createFile()

Source/Titanium/src/File.cpp
```diff
@@ -364,17 +369,17 @@ namespace TitaniumWindows
...
-				//if (native_file->isDirectory()) {
-				//	if (!native_file->deleteDirectory(recursive)) {
-				//		return false;
-				//	}
-				//} else {
-				//	if (!native_file->deleteFile()) {
-				//		return false;
-				//	}
-				//}
+				auto file = get_file_ptr(path_ + "\\" + value);
```
sgtcoolguy: Actually this also looks like we could do in the TitaniumKit layer. Basically  the isFolder guard above becomes isDirectory(), getDirectoryListing() is already a method on Ti.Filesystem.File, and we can do Ti.Filesystem.getFile(path, value) to get a handle on the child.

garymathews: I'll update the PR and implement this into TitaniumKit.

infosia:
Merged by #493

PR #492, opened 2015-12-03
Win10 Fixes

sgtcoolguy:
- try to make it so we can build/test against win10 for build machine.
- reference the win10 versions of the libraries when building for win10.
- collect the set of available windows phone sdks from windowslib/info
- fix the default tile for win10 to an asset we generate
- don't do post filtering of emulator list (requires appcelerator/windowslib#26)


infosia:
Looks good, but I can't test this now because it seems recent Visual Studio 2015 Update 1 broke my WP8.1 tools. `AppDeployCmd` doesn't work for me. :(

```
PS C:\Program Files (x86)\Microsoft SDKs\Windows Phone\v8.1\Tools\AppDeploy> .\AppDeployCmd.exe /EnumerateDevices
Error: The system cannot find the file specified. (Exception from HRESULT: 0x80070002)
```

@garymathews Could you test/review this PR if your environment is safe?

garymathews:
@infosia Sure, although I also have the Visual Studio 2015 Update 1. I'll still take a look though.

PR #495, opened 2015-12-08
[TIMOB-19927] Implement Titanium.Blob width and height

garymathews:
- Implemented ```Titanium.Blob.width``` and ```Titanium.Blob.height``` functionality
- Updated NMocha ```ti.blob.test.js``` to test ```width``` and ```height``` of ```Logo.png```

###### TEST CASE
```Javascript
var blob = Ti.Filesystem.getFile('Logo.png').read();
Ti.API.info('Blob->width: ' + blob.width);
Ti.API.info('Blob->height: ' + blob.height);
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19927)

### DIFF
--- a/Examples/NMocha/src/Assets/ti.blob.test.js
+++ b/Examples/NMocha/src/Assets/ti.blob.test.js
@@ -79,15 +79,15 @@ describe("Titanium.Blob", function () {
         finish();
     });
     it("width", function (finish) {
-        var blob = Ti.Filesystem.getFile('app.js').read();
+        var blob = Ti.Filesystem.getFile('Logo.png').read();
         should(blob.width).be.a.Number;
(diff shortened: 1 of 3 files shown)


Examples/NMocha/src/Assets/ti.blob.test.js
```diff
@@ -79,15 +79,15 @@ describe("Titanium.Blob", function () {
         finish();
     });
     it("width", function (finish) {
-        var blob = Ti.Filesystem.getFile('app.js').read();
```
sgtcoolguy: I think it'd still be useful to have test cases for width/height of a non-image Blob to ensure it returns the default values of 0.

PR #504, opened 2015-12-15
AttributedString for Label

infosia:
Fix for [TIMOB-19964](https://jira.appcelerator.org/browse/TIMOB-19964)

- Implement  ATTRIBUTE_UNDERLINE
- Implement ATTRIBUTE_LINK

Note: I also implemented *ATTRIBUTE_EXPANSION*, *ATTRIBUTE_WRITING_DIRECTION*, *ATTRIBUTE_KERN* and *ATTRIBUTE_LIGATURE*, but those properties does not work on both Windows Phone & Windows Store. I'm not sure why.

```javascript
var win = Titanium.UI.createWindow({
    backgroundColor: 'black',
});

var text = 'Bacon ipsum dolor Appcelerator Titanium rocks! Have you tried Hyperloop yet? Sit amet fatback leberkas salami sausage tongue strip steak.';

var attr = Titanium.UI.createAttributedString({
    text: text,
    attributes: [
        {
            type: Titanium.UI.ATTRIBUTE_FONT,
            value: { fontSize: 26 },
            range: [text.indexOf('Appcelerator'), ('Appcelerator').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_FOREGROUND_COLOR,
            value: "red",
            range: [text.indexOf('Appcelerator'), ('Appcelerator').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_EXPANSION,
            value: 0.25,
            range: [text.indexOf('Hyperloop'), ('Hyperloop').length]
        },
        {
            type: Ti.UI.ATTRIBUTE_WRITING_DIRECTION,
            value: Ti.UI.ATTRIBUTE_WRITING_DIRECTION_RIGHT_TO_LEFT,
            range: [text.indexOf('Hyperloop'), ('Hyperloop').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_KERN,
            value: 5,
            range: [text.indexOf('Hyperloop'), ('Hyperloop').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_LINK,
            value: 'https://github.com/appcelerator/hyperloop',
            range: [text.indexOf('Hyperloop'), ('Hyperloop').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_FOREGROUND_COLOR,
            value: "red",
            range: [text.indexOf('Hyperloop'), ('Hyperloop').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_FOREGROUND_COLOR,
            value: "blue",
            range: [text.indexOf('Titanium'), ('Titanium').length]
        },
        {
            type: Titanium.UI.ATTRIBUTE_FOREGROUND_COLOR,
            value: "yellow",
            range: [text.indexOf('rocks!'), ('rocks!').length]
        },
    ]
});

var label = Titanium.UI.createLabel({
    attributedString: attr
});

win.add(label);
win.open();
```


Source/UI/src/Label.cpp
```diff
@@ -187,35 +216,84 @@ namespace TitaniumWindows
...
 							case Titanium::UI::ATTRIBUTE_TYPE::FONT: 
 							{
 								const auto font = Titanium::UI::js_to_Font(static_cast<JSObject>(attribute.value));
 								TitaniumWindows::UI::ViewHelper::SetFont<Windows::UI::Xaml::Documents::Span^>(get_context(), span, font);
 								break;
 							}
+							case Titanium::UI::ATTRIBUTE_TYPE::EXPANSION:
+							{
+								const auto value = static_cast<double>(attribute.value);
+								if (value <= -0.5) {
```
sgtcoolguy: Looks like you're using the percentage values from the Windows docs to correlate these values, which seems like a good approach to me. We may want to let @jawa9000 know via some ticket that possible value range here for Windows and what it actually will do. I can't find exact docs for iOS, but it appears that they allow floats, with 0.0 being "default", but I'm not sure of the valid range of values. Here we're allowing -0.5 to 1.0 floats, but really lumping them into discrete possible values from an enumeration (so both -100 and -0.5 are the same, or .126 and .25).

It's possible we may want to be a little smarter about comparisons so say .126 is treated like .125 rather than lumped in with 0.25 range. (so maybe check distance to find the "nearest" value rather than simple <= checks).

Source/UI/src/Label.cpp
```diff
@@ -187,35 +216,84 @@ namespace TitaniumWindows
...
+							case Titanium::UI::ATTRIBUTE_TYPE::WRITING_DIRECTION:
+								if (attribute_styles.find(Titanium::UI::ATTRIBUTE_STYLE::WRITING_DIRECTION_LEFT_TO_RIGHT) != attribute_styles.end()) {
+									run->FlowDirection = FlowDirection::LeftToRight;
+								} else if (attribute_styles.find(Titanium::UI::ATTRIBUTE_STYLE::WRITING_DIRECTION_RIGHT_TO_LEFT) != attribute_styles.end()) {
+									run->FlowDirection = FlowDirection::RightToLeft;
+								}
+								break;
+							case Titanium::UI::ATTRIBUTE_TYPE::KERN:
+								run->CharacterSpacing = static_cast<std::int32_t>(attribute.value);
+								Typography::SetKerning(label__, true);
```
sgtcoolguy: should set kerning false if value is 0, true otherwise.

Source/UI/src/Label.cpp
```diff
@@ -187,35 +216,84 @@ namespace TitaniumWindows
...
+							}
+							case Titanium::UI::ATTRIBUTE_TYPE::WRITING_DIRECTION:
+								if (attribute_styles.find(Titanium::UI::ATTRIBUTE_STYLE::WRITING_DIRECTION_LEFT_TO_RIGHT) != attribute_styles.end()) {
+									run->FlowDirection = FlowDirection::LeftToRight;
+								} else if (attribute_styles.find(Titanium::UI::ATTRIBUTE_STYLE::WRITING_DIRECTION_RIGHT_TO_LEFT) != attribute_styles.end()) {
+									run->FlowDirection = FlowDirection::RightToLeft;
+								}
+								break;
+							case Titanium::UI::ATTRIBUTE_TYPE::KERN:
+								run->CharacterSpacing = static_cast<std::int32_t>(attribute.value);
```
sgtcoolguy: We have a cross-platform value (and doc) problem here. According to our docs, the value is in pixels for iOS (http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI-property-ATTRIBUTE_KERN) but looking at the code, we do no conversions and the iOS docs state it's points (https://developer.apple.com/library/ios/documentation/UIKit/Reference/NSAttributedString_UIKit_Additions/#//apple_ref/doc/constant_group/Character_Attributes)

the Windows docs state they take values in 1/1000 em: https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.documents.textelement.characterspacing

So I think we need to do some conversion of the value first here. I think we have to assume they're specifying pt and convert to equivalent of 1/1000 ems if we want to accept values people have used from iOS and treat them roughly equivalent.

cc @jawa9000

sgtcoolguy:
@infosia Looks pretty good. But I think we need to do conversion of the kern values for character spacing; and maybe do a closest value comparison for the expansion values rather than a hard <= check.

Once addressed, please feel free to merge (and ticket can be resolved for Release 6.0.0)

@jawa9000 I've opened https://jira.appcelerator.org/browse/TIDOC-2403 for the expansion value clarification, and https://jira.appcelerator.org/browse/TIDOC-2404 for kern.

PR #472, opened 2015-11-11
[TIMOB-19103] Make sure build ends with 0 warnings

infosia:
Fix for [TIMOB-19103](https://jira.appcelerator.org/browse/TIMOB-19103)

Make sure build ends with 0 warnings. Note that this introduces new temporary directory for build VS project files `$HOME/.titanium/vsbuild`.

- Remove `CMAKE_BUILD_TYPE`, because it has no effect for Visual Studio
- Change `tempBuildDir` to `$HOME/.titanium/vsbuild` to disable `MSB8029` warning
-  Add `WebHostHidden` to disable C4453 warning
- Remove duplicate resource

PR #505, opened 2015-12-15
Add clean command hook - WIP

infosia:
Fix for [TIMOB-20085](https://jira.appcelerator.org/browse/TIMOB-20085)

Note: This requires PR https://github.com/appcelerator/titanium_mobile/pull/7628 to be merged.

~~Attempt to hook `clean` command so that it deletes Visual Studio build directory.~~

~~Unfortunately this does not work yet because `clean.post` event is not fired on `ti clean`. I think it is because `ti clean` command does not load hooks. I saw hooks are loaded on `ti build` but not on `ti clean`.~~


sgtcoolguy:
@infosia It looks like clean.post and clean.windows.post should be fired: https://github.com/appcelerator/titanium_mobile/blob/bc85170157d3bebc5de1d61a9fe6e34bce84a8c9/cli/commands/clean.js#L109

But maybe because clean itself is a hook it behaves oddly? I don't think the other platforms have hooks on clean. @cb1kenobi is the guy who'd know what's going on here.

cb1kenobi:
@sgtcoolguy @infosia The code looks right, but if you look at the `clean` command's code, it only fires `clean.post` if you explicitly specify the platform (i.e. `--platforms windows`) or if the `build` directory exists.

For the latter, this is probably a bug. We should probably change https://github.com/appcelerator/titanium_mobile/blob/bc85170157d3bebc5de1d61a9fe6e34bce84a8c9/cli/commands/clean.js#L117 to a simple `else` and remove the existing `else` block.

infosia:
From what I can see, [clean.js](https://github.com/appcelerator/titanium_mobile/blob/bc85170157d3bebc5de1d61a9fe6e34bce84a8c9/cli/commands/clean.js#L117 ) is successfully removing its `buildDir`, which is `path.join(cli.argv['project-dir'], 'build')`.

But the problem is that Titanium CLI is assuming `path.join(cli.argv['project-dir'], 'build')` for the build directory, but Titanium CLI for Windows needs additional directory for the Visual Studio (it's needed because we had permission issue or something, as I recall), and we need a care for it. 

[validate.js](https://github.com/appcelerator/titanium_mobile_windows/blob/master/cli/commands/_build/validate.js#L104)

```javascript
	// try to build under temp if the path is shorter and we have write access
	var home = process.env.HOME || process.env.USERPROFILE || process.env.APPDATA;
	var ti_home = path.join(home, '.titanium');
	var tempBuildDir = path.join(ti_home, 'vsbuild');
```

And we need to delete this `tempBuildDir` at `ti clean`.

cb1kenobi:
@infosia All build files should go in the `build` directory. Not the home directory. I would revisit having the Visual Studio files in the `build` directory and determine if the permission issue is a fluke, just your machine, or a proper concern.

sgtcoolguy:
@cb1kenobi The build is done in the home directory temporarily due to path length restrictions on Windows. Typical project locations (specifically in a Studio workspace) will eventually generate paths too long for the MS toolchain. We added in special code for Windows where we basically move the build off to a temp location that shorter to avoid that issue and then copy it back once done.

I'm not sure when it got moved to HOME/.titanium/vsbuild/<project-name>, but originally I had it doing it in %TEMP%/<project-name>

cb1kenobi:
@sgtcoolguy Yeah, that rings a bell. That's unfortunate.

infosia:
@sgtcoolguy we've moved `%TEMP%/` to under HOME as of #472 because Visual Studio produces warning for it. (`MSB8029`)

infosia:
@cb1kenobi @sgtcoolguy Do you have any good idea to workaround for this? I think that the original issue is that Ti CLI does not load hooks during `ti clean` and thus the hook can't receive `clean.post` event.

infosia:
Pushed https://github.com/appcelerator/titanium_mobile/pull/7628, for giving platform-SDK a chance to load hooks before `ti clean`.

cli/commands/_clean.js
```diff
@@ -0,0 +1,17 @@
...
+ * @copyright
+ * Copyright (c) 2016 by Appcelerator, Inc. All Rights Reserved.
+ *
+ * @license
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+var path = require('path');
+
+exports.scanHooks = function(logger, config, cli) {
```
sgtcoolguy: I think this file is unnecessary (or should be).

PR #508, opened 2015-12-17
Implement ImageView properties

infosia:
Fix for [TIMOB-19954](https://jira.appcelerator.org/browse/TIMOB-19954)

- Implement `ImageView.defaultImage`, which enables to place placeholder image while waiting `image` loaded. especially useful when you need to grab `image` from network. 
- Implement `ImageView.toBlob` (but it does not work)
- Implement `View.toImage` (but it does not work)

Currently `toBlob` and `toImage` is always returning empty data. The problem is that they are using `Windows::UI::Xaml::Media::Imaging::RenderTargetBitmap::RenderAsync` to capture images from the component, but it always return `nullptr` when you call  `GetPixelsAsync`. I'm not sure if this API is actually implemented or working.

```javascript
var win = Ti.UI.createWindow({background:'black'});

var view = Ti.UI.createImageView({
    defaultImage: 'Logo.png', /* placeholder until image is loaded */
    image: 'http://api.randomuser.me/portraits/women/2.jpg'
});
win.add(view);

view.addEventListener('load', function () {
    Ti.API.info('image loaded');
});
var button = Ti.UI.createButton({ title: 'toBlob', top: 0, left: 0 });
button.addEventListener('click', function () {
    view.toImage(function () {
        Ti.API.info('toImage callback?');
    });
});
win.add(button);
win.open();
```

### DIFF
--- a/Source/UI/src/ImageView.cpp
+++ b/Source/UI/src/ImageView.cpp
@@ -9,7 +9,8 @@
 #include "TitaniumWindows/UI/ImageView.hpp"
 #include "TitaniumWindows/Utility.hpp"
 #include "LayoutEngine/LayoutEngine.hpp"
-
+#include "TitaniumWindows/UI/View.hpp"
+#include "Titanium/Blob.hpp"
 #include <ppltasks.h>
 
 namespace TitaniumWindows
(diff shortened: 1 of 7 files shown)


Source/UI/src/ImageView.cpp
```diff
@@ -193,5 +239,31 @@ namespace TitaniumWindows
...
+		std::shared_ptr<Titanium::Blob> ImageView::toBlob(JSValue callback) TITANIUM_NOEXCEPT
+		{
+			return this->toImage(callback, false);
+		}
+
+		std::shared_ptr<Titanium::Blob> ImageView::toImage(JSValue callback, const bool& honorScaleFactor) TITANIUM_NOEXCEPT
+		{
+			if (callback.IsObject() && static_cast<JSObject>(callback).IsFunction()) {
+				TitaniumWindows::UI::View::ToImage(image__, callback, get_object());
+			} else {
```
sgtcoolguy: This should work without a callback as well, we'd just need to basically do a concurrency::event.wait() in the impl...

infosia: Yeah, I tried that but I saw Xaml `DispatchedHandler` didn't work well with `concurrency::event` for any  task_continuation_context, and I ended up providing async function only. :(

PR #509, opened 2015-12-19
Fix: Ti.Codec.encodeNumber hangs on Windows 10

infosia:
Fix for [TIMOB-20155](https://jira.appcelerator.org/browse/TIMOB-20155)

I'm not very sure what's happening there, but use of ` task_continuation_context::use_arbitrary()` for async function fixed the issue. Seems like something has been changing regarding threading as of Window 10.

```javascript
var buffer = Ti.createBuffer({
    length: 8
});
Ti.Codec.encodeNumber({
    source: 78187493530,
    dest: buffer,
    type: Ti.Codec.TYPE_LONG
});
```

### DIFF
--- a/Examples/NMocha/src/Assets/ti.buffer.test.js
+++ b/Examples/NMocha/src/Assets/ti.buffer.test.js
@@ -352,8 +352,7 @@ describe("buffer", function() {
 	    finish();
 	});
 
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
Merged manually after fixing up the two skipped test's name to better reflect the encoding scenario they tested (my bad from earlier)

PR #511, opened 2015-12-19
[TIMOB-20086] Set environment variables

garymathews:
- Allow ```setup.js``` to set user environment variables

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-20086)


sgtcoolguy:
Merged manually.

PR #514, opened 2015-12-28
Fix: Window.fullscreen doesn't work on Windows 10

infosia:
Fix for [TIMOB-20166](https://jira.appcelerator.org/browse/TIMOB-20166)

```javascript
    fullscreen: true,
    backgroundColor: 'green'
});

var label = Ti.UI.createLabel({text:'Click here to exit fullscreen'});

// exit fullscreen when label is clicked
label.addEventListener('click', function () {
    win.fullscreen = false;
});
win.add(label);

win.open();
```


garymathews:
CR: **PASS**
FR: **PASS**

PR #515, opened 2015-12-29
Use SplitView for TabGroup on Windows 10

infosia:
Initial cut for [TIMOB-20164](https://jira.appcelerator.org/browse/TIMOB-20164)

This PR makes `Ti.UI.TabGroup` to use `Xaml::Controls::SplitView` instead of `Xaml::Controls::ListView` on Windows 10.

So, following image shows how it looks like on Windows 10 Mobile. I would say it's pretty much similar to how it works when you use Xaml `ListView`. The only difference is I use Xaml `Button` to realize Tab item (where "Tab 1" and "Tab 2" is placed) and this may not give "natural" look & feel compared to Xaml `ListView`.

surface_pro_3
And here's how our current "`ListView` version" looks like...I would say it's more natural look & feel.

surface_pro_3 2

I used combination of `Button` & `SplitView` in this PR but in fact I could even use `ListView` for the left pane of `SplitView,` and then it will give exactly same look & feel than before. But is that worth it? It gives no functional difference. The good thing when you use `SplitView` might be that it may give more option to implement, which is specific to Windows 10 such as `CompactPaneLength`, `OpenPaneLength` and `IsOpenPane` property.

### DIFF
--- a/Source/UI/include/TitaniumWindows/UI/TabGroup.hpp
+++ b/Source/UI/include/TitaniumWindows/UI/TabGroup.hpp
@@ -84,11 +84,13 @@ namespace TitaniumWindows
 
 #if defined(IS_WINDOWS_PHONE)
 			Windows::UI::Xaml::Controls::Pivot^ pivot__;
+#elif defined(IS_WINDOWS_10)
+			Windows::UI::Xaml::Controls::SplitView^ splitView__;
+			Windows::UI::Xaml::Controls::StackPanel^ splitPane__;
+
 #else
 			//
 			// We're going to use ListView for tab navigation because there's no Pivot for Windows Store app.
-			// On Windows 10 we might want to use SplitView instead.
-			// http://igrali.com/2015/04/12/getting-started-with-splitview-control-in-universal-apps/
 			//
 			Windows::UI::Xaml::Controls::ListView^ sectionView__;
 			Windows::UI::Xaml::Data::CollectionViewSource^ sectionViewSource__;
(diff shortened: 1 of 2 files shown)


infosia:
test this please

sgtcoolguy:
Argh, the windows 10 test is failing because the emulator needs a pin to deploy to it again. Looks like it re-requires this occasionally, perhaps after restarts? It should only ever ask the first time and then came back from a checkpoint with the pin/connection already established.

sgtcoolguy:
@infosia It sounds like we actually prefer the current implementation? I do agree it looks more natural with the ListView than with SplitView using buttons.

I think we just want to make sure we have the best mapping for TabGroup in the platform. It appears to me like the Windows docs explicitly state using Pivots for Tabs: https://msdn.microsoft.com/en-us/library/windows/apps/dn997788.aspx

That appears to be the UI pattern we use on Win 8.1 phone, and we use this ListView for Win 8.1 desktop. By default, we use the ListView for Windows 10 Desktop and Mobile (just because of how the macro check worked). Perhaps we just want to switch Windows 10 to use the Windows 8.1 phone code path so it also uses pivots? That may make most sense to me.

The SplitView appears to be more of a Nav Pane/Hamburger menu implementation along the side. Maybe we'd want to (eventually) just expose this in a Windows specific namespace so users could make use of it in case they need one, i.e. Ti.Windows.SplitView?  Similar to Ti.UI.iOS.SplitWindow.

infosia:
ok make sense, I close this PR for now & will push new PR to use Pivots on Windows 10.

PR #516, opened 2015-12-29
[TIMOB-20162] Obtain mapservicetoken from tiapp.xml

garymathews:
- Load windows specific properties into ```_app_prop_.json``` from ```tiapp.xml```
- Obtain and use ```mapservicetoken``` from ```_app_prop_.json```

###### EXAMPLE tiapp.xml
```XML
...
<windows>
  <mapservicetoken>VrSrmXR8B5bgklWrs0CK_w</mapservicetoken>
</windows>
...
```

###### TEST CASE
```Javascript
    view = Ti.Map.createView({
        mapType: Titanium.Map.NORMAL_TYPE,
        region: {latitude:37.390749, longitude:-122.081651}
    });
win.add(view);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-20162)

### DIFF
--- a/Source/Map/src/View.cpp
+++ b/Source/Map/src/View.cpp
@@ -10,6 +10,7 @@
 #include "TitaniumWindows/Utility.hpp"
 #include "TitaniumWindows/Map/Annotation.hpp"
 #include "TitaniumWindows/UI/WindowsViewLayoutDelegate.hpp"
+#include "TitaniumWindows/App/Properties.hpp"
 #include "Titanium/detail/TiLogger.hpp"
 #include "Titanium/detail/TiImpl.hpp"
 #include "TitaniumWindows/WindowsMacros.hpp"
@@ -41,9 +42,26 @@ namespace TitaniumWindows
 #if defined(IS_WINDOWS_PHONE) || defined(IS_WINDOWS_10)
 			mapview__ = ref new MapControl();
 
-			// Set service token
-			// TODO : Load this from app properties, windows_map_service_token?
-			mapview__->MapServiceToken = "VrSrmXR8B5bgklWrs0CK_w";
+			// obtain mapservicetoken
+			auto titanium_js = get_context().get_global_object().GetProperty("Titanium");
+			TITANIUM_ASSERT(titanium_js.IsObject());
+			auto titanium_obj = static_cast<JSObject>(titanium_js);
+
+			auto app_js = titanium_obj.GetProperty("App");
+			TITANIUM_ASSERT(app_js.IsObject());
+			auto app_obj = static_cast<JSObject>(app_js);
+
+			auto properties_js = app_obj.GetProperty("Properties");
+			TITANIUM_ASSERT(properties_js.IsObject());
+			auto properties_obj = static_cast<JSObject>(properties_js);
+			auto properties_ptr = properties_obj.GetPrivate<TitaniumWindows::App::Properties>();
+			TITANIUM_ASSERT(properties_ptr);
+
+			std::string mapservicetoken = properties_ptr->getString("mapservicetoken").get();
+			TITANIUM_ASSERT_AND_THROW(!mapservicetoken.empty(), "Windows mapservicetoken must be defined in tiapp.xml");
+
+			// set mapservicetoken
+			mapview__->MapServiceToken = TitaniumWindows::Utility::ConvertString(mapservicetoken);
 
 			Titanium::Map::View::setLayoutDelegate<TitaniumWindows::UI::WindowsViewLayoutDelegate>();
 
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -565,6 +565,12 @@ function copyResources(next) {
 			Object.keys(this.tiapp.properties).forEach(function (prop) {
 				props[prop] = this.tiapp.properties[prop].value;
 			}, this);
+			this.tiapp.windows && Object.keys(this.tiapp.windows).forEach(function (prop) {
+				// ignore appxmanifest
+				if (prop !== 'manifest') {
+					props[prop] = this.tiapp.windows[prop];
+				}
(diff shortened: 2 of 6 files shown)


Source/Map/src/View.cpp
```diff
@@ -41,9 +41,14 @@ namespace TitaniumWindows
 #if defined(IS_WINDOWS_PHONE) || defined(IS_WINDOWS_10)
 			mapview__ = ref new MapControl();
 
-			// Set service token
-			// TODO : Load this from app properties, windows_map_service_token?
-			mapview__->MapServiceToken = "VrSrmXR8B5bgklWrs0CK_w";
+			// obtain mapservicetoken
+			auto mapservicetoken = get_context().JSEvaluateScript("Ti.App.Properties.getString('mapservicetoken')");
+			TITANIUM_ASSERT_AND_THROW(mapservicetoken.IsString(), "Expected a String");
+			TITANIUM_ASSERT_AND_THROW(mapservicetoken.IsNull(), "Windows mapservicetoken must be defined in tiapp.xml");
```
infosia: This assertion fails every time when `mapservicetoken.IsString()` equals true. You'll need to reverse the order & check if it's *not* null...

```diff
+			TITANIUM_ASSERT_AND_THROW(!mapservicetoken.IsNull(), "Windows mapservicetoken must be defined in tiapp.xml");
+			TITANIUM_ASSERT_AND_THROW(mapservicetoken.IsString(), "Expected a String");
```

Source/Map/src/View.cpp
```diff
@@ -41,9 +41,14 @@ namespace TitaniumWindows
 #if defined(IS_WINDOWS_PHONE) || defined(IS_WINDOWS_10)
 			mapview__ = ref new MapControl();
 
-			// Set service token
-			// TODO : Load this from app properties, windows_map_service_token?
-			mapview__->MapServiceToken = "VrSrmXR8B5bgklWrs0CK_w";
+			// obtain mapservicetoken
+			auto mapservicetoken = get_context().JSEvaluateScript("Ti.App.Properties.getString('mapservicetoken')");
```
infosia: I would prefer dealing with `Titanium::App::Properties` directly like below...maybe just a matter of preference though.

```c++
  auto Properties = get_context().JSEvaluateScript("Ti.App.Properties");
  TITANIUM_ASSERT(Properties.IsObject());
  auto Properties_ptr = static_cast<JSObject>(Properties).GetPrivate<TitaniumWindows::App::Properties>();
  auto mapservicetoken = Properties_ptr->getString("mapservicetoken");
  if (mapservicetoken) {

  }
```

cli/commands/_build/copy.js
```diff
@@ -565,6 +565,12 @@ function copyResources(next) {
 			Object.keys(this.tiapp.properties).forEach(function (prop) {
 				props[prop] = this.tiapp.properties[prop].value;
 			}, this);
+			this.tiapp.windows && Object.keys(this.tiapp.windows).forEach(function (prop) {
+				// ignore appxmanifest
+				if (prop !== 'manifest') {
+					props[prop] = this.tiapp.windows[prop];
+				}
+			}, this);
```
infosia: Is there any reason we don't use [`property` element](http://docs.appcelerator.com/platform/latest/#!/guide/tiapp.xml_and_timodule.xml_Reference-section-29004921_tiapp.xmlandtimodule.xmlReference-property) here?

If you need platform-specific property, I think we want to do like this:

```xml
<windows>
  <property name="mapservicetoken">VrSrmXR8B5bgklWrs0CK</property>
</windows>
```

Source/Map/src/View.cpp
sgtcoolguy: Additionally, you may want to give a more detailed message if the value isn't a string, like "Expected string value for Windows map service token"

cli/commands/_build/copy.js
sgtcoolguy: I'm not sure we've really established a style for platform-specific properties yet. So far, we do tend to be leaning towards the tag name being the key rather than a generic property tag like we do at the top-level. cc @jhaynie @pec1985

Source/Map/src/View.cpp
garymathews: ```Properties_ptr->getString("mapservicetoken");```

Is inaccessible since it's private

infosia: Good point...I thought I used to "expose" all of those stub functions as *public* (like [this](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/include/Titanium/Buffer.hpp#L79)) but seems like we are not doing so. I think best practice here is making functions public, and make member variables protected for best encapsulation with convenience.

Source/Map/src/View.cpp
```diff
@@ -41,9 +42,26 @@ namespace TitaniumWindows
...
+			TITANIUM_ASSERT(app_js.IsObject());
+			auto app_obj = static_cast<JSObject>(app_js);
+
+			auto properties_js = app_obj.GetProperty("Properties");
+			TITANIUM_ASSERT(properties_js.IsObject());
+			auto properties_obj = static_cast<JSObject>(properties_js);
+			auto properties_ptr = properties_obj.GetPrivate<TitaniumWindows::App::Properties>();
+			TITANIUM_ASSERT(properties_ptr);
+
+			std::string mapservicetoken = properties_ptr->getString("mapservicetoken").get();
```
garymathews: @infosia Any idea why this produces:

```
LNK2019	unresolved external symbol "__declspec(dllimport) public: virtual class boost::optional ...
```

Even when I reference the ```TitaniumWindows_Ti``` library to ```TitaniumWindows_Map```

garymathews:
test this please

infosia:
ok looks good, merging :tada:

PR #519, opened 2016-01-04
Fix: Window width & height for Store App doesn't work

infosia:
Fix for [TIMOB-20187](https://jira.appcelerator.org/browse/TIMOB-20187)

*This works only on Windows 10 Store App*.

```javascript
var win = Ti.UI.createWindow({width:600, height:800, backgroundColor:'green'}),
    lbl = Ti.UI.createLabel({ text: 'Hello, World' });
win.add(lbl);
win.open();
```


garymathews:
CR: **PASS**
FR: **PASS**

PR #528, opened 2016-01-08
[TIMOB-20163] Windows: Add Contact group functionality for Windows 10 - WIP, NOT READY TO MERGE

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-20163

WIP, need to fix up Ti.Contacts.Person because our API seems to allow a person to be created outside a group whereas the WIndows APIs don't.

It appears that We're echoing an iOS style API where a contact exists and can be a member of multiple groups. Windows appears to require a strict hierarchy where a contact is a member of a list. We might have to generate a default list to place people who haven't been added to a specific group.


sgtcoolguy:
@infosia I could use another opinion on what to do here. I'm seeing that Windows effectively only allows an app to create a single ContactList. After that it throws an exception saying "Request is not supported". I can't figure out a good way to implement Ti.Contacts.Group. I can create a single contact list for an app and throw Ti.Contacts.Person/Contacts into it so we store contacts, but I don't see any way to actually assign groups to them or to store groups. I'm thinking we may actually need to cheat and use Ti.Database to effectively store the groups ourselves.

ContactList _seems_ like the best fit, but again an app can only have one. (they don't tell you that anywhere in the docs). I looked at hacking ContactAnnotation and ContactAnnotationList, but we can't assign a name to the ContactAnnotationList. Maybe we could hack the ProviderProperties map on Contact? https://msdn.microsoft.com/en-us/library/windows/apps/windows.applicationmodel.contacts.contact.providerproperties.aspx

infosia:
Can't we just say "Grouping is not supported for Windows" when ContactList only supports single list?

sgtcoolguy:
Merged manually after squashing.

PR #529, opened 2016-01-11
[TIMOB-20191] e.source of a table click event should return table row

infosia:
Fix for [TIMOB-20191](https://jira.appcelerator.org/browse/TIMOB-20191)

```javascript
var tableData = [{ title: 'Apples' }, { title: 'Bananas' }, { title: 'Carrots' }, { title: 'Potatoes' }];
    data: tableData
});
table.addEventListener('click', function (e) {
    Ti.API.info(e.source != table); // should be true
    Ti.API.info(e.source == e.row); // should be true
    Ti.API.info(e.source.title == e.row.title); // should be true
});
win.add(table);
win.open();
```


sgtcoolguy:
Too bad there's no way to automate this test....

PR #532, opened 2016-01-13
[TIMOB-19946] Implement Ti.UI.AlertDialog.hide() functionality

garymathews:
- Implemented the ability to hide active alert dialog's

###### TEST CASE
```Javascript
var a = Ti.UI.createAlertDialog({title: 'Alert', message: 'This will hide in 3 seconds'});
a.show();

// hide alert after 3 seconds
setTimeout(function () {
    a.hide();
}, 3000);

// show alert after another 3 seconds
setTimeout(function () {
    a.show();
}, 6000);
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19946)

### DIFF
--- a/Source/UI/src/AlertDialog.cpp
+++ b/Source/UI/src/AlertDialog.cpp
@@ -95,7 +95,8 @@ namespace TitaniumWindows
 
 				// show first dialog and then proceed to next after finishing click event
 				if (dialog_queue__.size() == 1) {
-					concurrency::create_task(dialog__->ShowAsync()).then(on_click__);
+					dialog_task__ = dialog__->ShowAsync();
+					concurrency::create_task(dialog_task__).then(on_click__);
 				}
 
 			} catch (::Platform::COMException^ ce) {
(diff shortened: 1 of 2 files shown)


garymathews:
test this please

sgtcoolguy:
@garymathews This doesn't appear to work for me. I tested on Windows 10 mobile and got an "Access is denied" exception when the second show() call happened. This is because the hide() call cancelled the task and removed the dialog from the queue, but *the dialog never closed!* It appears to be a limitation/difference between Store/desktop and Phone/mobile? http://stackoverflow.com/questions/28795850/how-to-dismiss-a-messagedialog-from-code-in-windows-phone-8-1

So either we should:
 - use macros to turn off the hide() impl used here for phones to avoid the scenario I'm seeing where it dumps the dialog from the queue without closing it. and maybe log something saying you can't hide alert dialogs programatically on phone;
 - **or** we may need to look into using a different mechanism for the dialog other than MessageDialog. I saw remarks pointing to https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.xaml.controls.contentdialog.aspx if you need to programatically close the dialog.

garymathews:
test this please

Source/UI/src/AlertDialog.cpp
```diff
@@ -105,5 +106,19 @@ namespace TitaniumWindows
 				detail::ThrowRuntimeError("Ti.UI.AlertDialog", "Exception during show()");
 			}
 		}
+
+		void AlertDialog::hide() TITANIUM_NOEXCEPT
+		{
+#ifndef IS_WINDOWS_PHONE
```
sgtcoolguy: I still think this won't guard against windows 10 mobile...

garymathews:
test this please

Source/UI/src/AlertDialog.cpp
```diff
@@ -105,5 +106,19 @@ namespace TitaniumWindows
 				detail::ThrowRuntimeError("Ti.UI.AlertDialog", "Exception during show()");
 			}
 		}
+
+		void AlertDialog::hide() TITANIUM_NOEXCEPT
+		{
+#if defined(IS_WINDOWS_PHONE)
```
garymathews: It seems the ```IS_WINDOWS_MOBILE``` condition is still satisfied on a Windows 10 Universal Desktop application. I've removed it for now so ```IS_WINDOWS_PHONE``` will have to surpass.

infosia:
Tested this on Windows 10 Mobile but I got ""Access is denied" exception.

garymathews:
@infosia Updated.

garymathews:
test this please

infosia:
Looks good to me. Will merge once all Jenkins tests go through.

infosia:
test this please

PR #534, opened 2016-01-19
Support differing capabilities per platform/SDK version

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-20231

### DIFF
--- a/Tools/Scripts/build/test.js
+++ b/Tools/Scripts/build/test.js
@@ -168,11 +168,7 @@ function addTiAppProperties(sdkVersion, next) {
 			content.push('\t<property name="presetInt" type="int">1337</property>');
 			content.push('\t<property name="presetDouble" type="double">1.23456</property>');
 			content.push('\t<windows><manifest><Capabilities><Capability Name=\"internetClient\" />');
-			if (sdkVersion == WIN_10) {
-				content.push('\t<uap:Capability Name=\"contacts\" />');
-			} else {
-				content.push('\t<m3:Capability Name=\"contacts\" />');
-			}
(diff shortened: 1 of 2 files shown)


infosia:
Is  there some more example that I can test? against From what I can see it converts `<Capability Name="contacts" />` to  `<m3:Capability Name="contacts" />` or `<uap:Capability Name="contacts" />` automatically depending on the target platform, right? 

```xml
<!-- Windows Phone 8.1 only -->
<manifest target="phone" version="8.1">
    <Capability Name="contacts" />
    <Capability Name="location" />
</manifest>
<!-- Windows 10 only -->
<manifest target="win10" version="10.0">
    <Capability Name="contacts" />
</manifest>
```

sgtcoolguy:
@infosia Correct, I updated the ticket with more details: https://jira.appcelerator.org/browse/TIMOB-20231

I use a whitelist of possible capabilities. If you enter in an invalid capability name, we ignore it. You also don't need to use namespaces tags, just the vanilla `<Capability Name="whatever" />` tag.

So some good tests would be to:
- make sure if you enter in random name we ignore it
- that if you specify "contacts" as a capability it uses the right namespaced tag in the appxmanifest for Win 8.1 Phone or Win 10 (you'll know if it doesn't break the build :smile:)
- That if you specify a good capability name that isn't namespaces it passes through fine.
- That if you specify a valid capability for Windows 10 but specify that it targets 8.1/phone we ignore it to avoid breaking the build.

We may want to actually spit out some warning text for the first and last use cases - right now it silently ignores those tags.

infosia:
ok looks good, merging.

PR #535, opened 2016-01-20
[TIMOB-19957] Implement Ti.UI.ProgressBar.backgroundColor

garymathews:
- Implement ```Titanium.UI.ProgressBar.backgroundColor```
- Implement ```Titanium.UI.ProgressBar.backgroundDisabledColor```

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow({}),
    progressBar = Ti.UI.createProgressBar({top: 100, backgroundColor:'green', backgroundDisabledColor:'red'});

win.add(progressBar);
win.open();

```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-19957)

### DIFF
--- a/Source/UI/src/ProgressBar.cpp
+++ b/Source/UI/src/ProgressBar.cpp
@@ -46,7 +46,18 @@ namespace TitaniumWindows
 			panel__->Orientation = Windows::UI::Xaml::Controls::Orientation::Vertical;
 
 			bar__ = ref new Windows::UI::Xaml::Controls::ProgressBar();
-			bar__->Background = ref new Windows::UI::Xaml::Media::SolidColorBrush(Windows::UI::Colors::Gray);
+			backgroundColorBrush__ = ref new Windows::UI::Xaml::Media::SolidColorBrush(Windows::UI::Colors::Gray);
+			bar__->Background = backgroundColorBrush__;
+			bar__->IsEnabledChanged += ref new Windows::UI::Xaml::DependencyPropertyChangedEventHandler(
+				[=](Platform::Object^ sender, DependencyPropertyChangedEventArgs^ e) {
+					auto control = dynamic_cast<Windows::UI::Xaml::Controls::Control^>(sender);
+					if (control->IsEnabled) {
+						control->Background = backgroundColorBrush__;
+					} else {
+						control->Background = backgroundDisabledColorBrush__;
+					}
+				}
+			);
 
 			label__ = ref new Windows::UI::Xaml::Controls::TextBlock();
 			label__->Text = "";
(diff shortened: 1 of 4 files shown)


Source/UI/src/ProgressBar.cpp
```diff
@@ -109,5 +120,18 @@ namespace TitaniumWindows
 			bar__->Foreground = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
 		}
 
+		void ProgressBar::set_backgroundColor(const std::string& colorName) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ProgressBar::set_backgroundColor(colorName);
+			backgroundColorBrush__ = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
+			bar__->Background = backgroundColorBrush__;
```
sgtcoolguy: We likely need to check if bar->IsEnabled here before setting it?

Source/UI/src/ProgressBar.cpp
```diff
@@ -109,5 +120,18 @@ namespace TitaniumWindows
...
+		{
+			Titanium::UI::ProgressBar::set_backgroundColor(colorName);
+			backgroundColorBrush__ = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
+			bar__->Background = backgroundColorBrush__;
+		}
+		
+		void ProgressBar::set_backgroundDisabledColor(const std::string& colorName) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ProgressBar::set_backgroundDisabledColor(colorName);
+			backgroundDisabledColorBrush__ = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
```
sgtcoolguy: Check if bar__->IsEnabled here and if not, set the bar__->Background to this brush?

Source/UI/src/ProgressBar.cpp
```diff
@@ -109,5 +120,22 @@ namespace TitaniumWindows
 			bar__->Foreground = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
 		}
 
+		void ProgressBar::set_backgroundColor(const std::string& colorName) TITANIUM_NOEXCEPT
+		{
+			if (bar__->IsEnabled) {
```
sgtcoolguy: This if guard should only be around the actual bar__->Background setter code. I think e still want to create the brush and set the property internally.

Source/UI/src/ProgressBar.cpp
```diff
@@ -109,5 +120,22 @@ namespace TitaniumWindows
...
+				Titanium::UI::ProgressBar::set_backgroundColor(colorName);
+				backgroundColorBrush__ = ref new Windows::UI::Xaml::Media::SolidColorBrush(WindowsViewLayoutDelegate::ColorForName(colorName));
+				bar__->Background = backgroundColorBrush__;
+			}
+		}
+		
+		void ProgressBar::set_backgroundDisabledColor(const std::string& colorName) TITANIUM_NOEXCEPT
+		{
+			if (!bar__->IsEnabled) {
```
sgtcoolguy: Again, we want to retain the color and brush, just check if it's not enabled and if not, then call bar__->Background = backgroundDisabledColorBrush__;

infosia:
Sorry for late reply but I just noticed that background color should have been implemented in [WindowsViewLayoutDelegate.cpp#L1093](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/UI/src/WindowsViewLayoutDelegate.cpp#L1093). Because Xaml ProgressBar is subclass of Control, this logic should be working now but you mean it's not? I'm interested in why WindowsViewLayoutDelegate does not work...

infosia:
I've confirmed that it's already implemented in [WindowsViewLayoutDelegate](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/UI/src/WindowsViewLayoutDelegate.cpp#L1093) at some point (maybe [TIMOB-19921](https://jira.appcelerator.org/browse/TIMOB-19921)) . So I'm afraid to tell you I would close this ticket for now. I see that it does not work quite well due to the layout issue, but it should be another ticket. I'll open new ticket for it soon.

PR #536, opened 2016-01-21
If DB.execute returns an immediate SQLITE_DONE result, return null ra…

sgtcoolguy:
…ther than an empty ResultSet

I'm not sure if this will fix https://jira.appcelerator.org/browse/TIMOB-20222 by itself, but it appears to at least play a role in that bug. Alloy seems to rely on db.execute returning a null value when "pragma table_info("table")" is called for a nonexistent table, which is specifically what Android does if a SELECT or PRAGMA returns a cursor with 0 columns. So in our case, if we get a value of SQL_DONE and no rows, we return a JS null instead of an empty ResultSet. This implementation may not be correct. I might need to instead call sqlite3_column_count(stmt) and if it's 0 then return null?

### DIFF
--- a/Source/TitaniumKit/src/Database/DB.cpp
+++ b/Source/TitaniumKit/src/Database/DB.cpp
@@ -22,13 +22,13 @@ namespace Titanium
 			TITANIUM_LOG_DEBUG("DB:: ctor ", this);
 		}
 
-		DB::~DB() 
+		DB::~DB()
 		{
 			TITANIUM_LOG_DEBUG("DB:: dtor ", this);
 			close();
 		}
 
-		void DB::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) 
+		void DB::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments)
 		{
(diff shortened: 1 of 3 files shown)


infosia:
Just changed assignee to @garymathews as Gary should have better understanding with this. Thanks!

sgtcoolguy:
I just pushed another commit to return null if statement immediately returns SQLITE_DONE and has a column count of 0, rather than just an immediate SQLITE_DONE.

Source/TitaniumKit/src/Database/DB.cpp
```diff
@@ -176,11 +176,16 @@ namespace Titanium
 			resultSet->setDatabase(this);
 			const auto insert_result = resultSets__.emplace(statement, resultSet);
 			TITANIUM_ASSERT(insert_result.second);
+
 			int affectedRows = 0;
 			if (stepResult == SQLITE_DONE) {
+				int columns = sqlite3_column_count(statement);
```
garymathews:
```C++
const int
```

garymathews:
FR: **PASS**
- Fixed [TIMOB-20222](https://jira.appcelerator.org/browse/TIMOB-20222)

PR #537, opened 2016-01-26
Reduce memory consumption at TitaniumKit startup

infosia:
Fix for [TIMOB-20256](https://jira.appcelerator.org/browse/TIMOB-20256).

## Background

From the starting point of HAL, we have been declaring proxy constant using C++ object member by holding `JSValue`.

```c++
class Widget : public JSExportObject, public JSExport<Widget> {
  private:
      JSValue myProperty_1__;
      JSValue myProperty_2__;
};
```

In this case we are creating these instances at proxy constructor:

```c++
Widget::Widget(const JSContext& js_context) HAL_NOEXCEPT
  : JSExportObject(js_context)
  , myProperty_1__(js_context.CreateObject())
  , myProperty_2__(js_context.CreateNumber()) {
}
```

Getter would be implemented like below. We are storing `JSValue` in C++ member, and this makes sense because we don't want to create new JSValue every time getter is called.

```c++
JSValue Widget::js_get_myProperty1() const HAL_NOEXCEPT {
  return myProperty_1__;
}

void Widget::JSExportInitialize() {
  JSExport<Widget>::AddValueProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1)));
  JSExport<Widget>::AddValueProperty("myProperty2", std::mem_fn(&Widget::js_get_myProperty2)));
}
```

## Problem

The problem is, by this design, **we had to create *all* constants at proxy constructor**. Even when we don't use it at all.

```c++
class Widget : public JSExportObject, public JSExport<Widget> {
  private:
      JSValue myProperty_1__;
      JSValue myProperty_2__;

     // ...
     // what if we have 1,000 properties? Do we create *all* of them at constructor?
     // ...
      JSValue myProperty_1000__;
};
```

## Solution

Load properties on demand and cache the result if it's a read-only constant. That means **if constant is never used, no `JSValue` is created**. I introduced `JSExport<T>::AddConstantProperty` for the constant property instead of `JSExport<T>::AddValueProperty`. If you declare object property is constant, the "getter" callback will be called only once and the result is cached. This solution should be easy to implement for proxy developers, which should minimize transition cost to introduce cache mechanics.

```diff
-  JSExport<Widget>::AddValueProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1));
+  JSExport<Widget>::AddConstantProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1));
```

In this case getter would be like below, note that it's safe & efficient when you create new JSValue at getter, it's automatically cached and reused by HAL. In fact this getter is called only once when user uses this property for the first time. Again, this means **if this property is never used, no `JSValue` associated with this property is created**.

```c++
JSValue Widget::js_get_myProperty1() const HAL_NOEXCEPT {
  return get_context().CreateObject();
}
```


infosia:
Closing, as taken over by #539

PR #539, opened 2016-01-26
 Reduce memory consumption at TitaniumKit startup

infosia:
Fix for [TIMOB-20256](https://jira.appcelerator.org/browse/TIMOB-20256).

## Background

From the starting point of HAL, we have been declaring proxy constant using C++ object member by holding `JSValue`.

```c++
class Widget : public JSExportObject, public JSExport<Widget> {
  private:
      JSValue myProperty_1__;
      JSValue myProperty_2__;
};
```

In this case we are creating these instances at proxy constructor:

```c++
Widget::Widget(const JSContext& js_context) HAL_NOEXCEPT
  : JSExportObject(js_context)
  , myProperty_1__(js_context.CreateObject())
  , myProperty_2__(js_context.CreateNumber()) {
}
```

Getter would be implemented like below. We are storing `JSValue` in C++ member, and this makes sense because we don't want to create new JSValue every time getter is called.

```c++
JSValue Widget::js_get_myProperty1() const HAL_NOEXCEPT {
  return myProperty_1__;
}

void Widget::JSExportInitialize() {
  JSExport<Widget>::AddValueProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1)));
  JSExport<Widget>::AddValueProperty("myProperty2", std::mem_fn(&Widget::js_get_myProperty2)));
}
```

## Problem

The problem is, by this design, **we had to create *all* constants at proxy constructor**. Even when we don't use it at all.

```c++
class Widget : public JSExportObject, public JSExport<Widget> {
  private:
      JSValue myProperty_1__;
      JSValue myProperty_2__;

     // ...
     // what if we have 1,000 properties? Do we create *all* of them at constructor?
     // ...
      JSValue myProperty_1000__;
};
```

## Solution

Load properties on demand and cache the result if it's a read-only constant. That means **if constant is never used, no `JSValue` is created**. I introduced `JSExport<T>::AddConstantProperty` for the constant property instead of `JSExport<T>::AddValueProperty`. If you declare object property is constant, the "getter" callback will be called only once and the result is cached. This solution should be easy to implement for proxy developers, which should minimize transition cost to introduce cache mechanics.

```diff
-  JSExport<Widget>::AddValueProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1));
+  JSExport<Widget>::AddConstantProperty("myProperty1", std::mem_fn(&Widget::js_get_myProperty1));
```

In this case getter would be like below, note that it's safe & efficient when you create new JSValue at getter, it's automatically cached and reused by HAL. In fact this getter is called only once when user uses this property for the first time. Again, this means **if this property is never used, no `JSValue` associated with this property is created**.

```c++
JSValue Widget::js_get_myProperty1() const HAL_NOEXCEPT {
  return get_context().CreateObject();
}
```


Source/TitaniumKit/src/MapModule.cpp
```diff
@@ -12,156 +12,177 @@
...
+		return get_context().CreateNumber(Titanium::Map::Constants::to_underlying_type(Titanium::Map::GOOGLE_PLAY_SERVICE_STATE::VERSION_UPDATE_REQUIRED));
 	}
 
 	TITANIUM_PROPERTY_GETTER(MapModule, SUCCESS)
 	{
-		return success__;
+		return get_context().CreateNumber(Titanium::Map::Constants::to_underlying_type(Titanium::Map::GOOGLE_PLAY_SERVICE_STATE::SUCCESS));
+	}
+
+	TITANIUM_PROPERTY_GETTER(MapModule, Camera)
```
sgtcoolguy: So I see how you're avoiding holding content values in fields and just generating a value in the getters. But it looks like the approach you're using for classes/types is still pretty similar to before. We "set" the JSClass in a field and wrap it in an object we return (and cache like constants from then on?).

I assume this is so that we can set a platform-specific subclass before we generate the JS object that wraps it? Why not just do the same thing as the constants and never hold it in a field, but instead require the platform-specific subclass of the module to override the class/type property getter to instantiate the JSClass when called?

Because otherwise, yes we're eliminating holding every constant in-memory immediately - but we're not avoiding pre-loading all the types/proxy classes - they'll get set at program startup into fields and then wrapped in an object on-demand.

sgtcoolguy:
@infosia Looks like a good approach, so we're not keeping around every immutable constant property in-memory forever (I also have to assume we'd have a copy of each field in each "instance" of a type as well then and not just at the "class" level?)

Are the constants stored at the JSClass level (i.e. is the constant shared across all instances)?

In any case, I think it'd be wise to take this the whole way and basically instantiate the JSClass on property access like we do for constant values rather than hold a JSClass field for types in a given module's namespace (i.e. we shouldn't set/hold FilesystemModule as a field on TiModule, but instantiate it on-demand and cache that instance for good)

Another thing to consider is whether we even _want_ to cache forever like this. What's the overhead of re-creating JSValue constants? JSClass constants? Maybe there's not enough overhead for things like Numbers and Strings, but there is for JSClasses. Or maybe we want to consider some LRU cache that can free up memory over time?

Source/TitaniumKit/src/MapModule.cpp
infosia: Yes, that's a good point to discuss.  Like you suggest, we actually don't have to hold the reference to the JSClass and creates it on demand. The reason because we're holding JSClass like before is, I was trying to keep basic design of the `ApplicationBuilder` to make code migration easier and also keep how TitaniumKit deals with cross-platform using subclass. The one good thing when you hold JSClass (vs JSObject) is JSClass static in HAL. It means everytime you call `JSExport<SomeClass>::Class()`, it returns same instance.

infosia:
> I also have to assume we'd have a copy of each field in each "instance" of a type as well then and not just at the "class" level?

Yes, we were holding new instances for each field on each JSExportObject instances. Now it's cached in JSClass level that is static. And now it's only loaded on demand. I think that's huge.

> Or maybe we want to consider some LRU cache that can free up memory over time?

Yes, that's a very good point. Since currently HAL is keeping reference to the fields in the [JSExportClass::constants_cache__](https://github.com/appcelerator/HAL/pull/60/files#diff-4387606ff3ca17a99bee85029db34673R141) using `std::unordered_map`, it's actually kept forever once it's loaded for now. LRU cache would be a very strong candidate to tackle next.

sgtcoolguy:
@infosia I think this is a very good start, and we'll likely want to open another ticket or two to further pursue better memory usage such as the LRU of cached values, or maybe doing on-demand class generation. But since this should be a huge improvement as-is, I'm going to merge it.

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-20316

PR #542, opened 2016-01-29
Fix: File.write can't handle UTF-8 string

infosia:
Fix for [TIMOB-20288](https://jira.appcelerator.org/browse/TIMOB-20288).

```javascript
var filePath = Ti.Filesystem.applicationDataDirectory + "test.txt";
var file = Ti.Filesystem.getFile(filePath);
file.write("日本語 : Japanese");
alert(file.read().text);
```


garymathews:
FR: **PASS**
CR: **PASS**

PR #545, opened 2016-02-02
Fix: ProgressBar default layout should be SIZE, not FILL

infosia:
Fix for [TIMOB-20282](https://jira.appcelerator.org/browse/TIMOB-20282)

According to [Transitioning to the New UI Layout System](http://docs.appcelerator.com/platform/latest/#!/guide/Transitioning_to_the_New_UI_Layout_System), the default layout of `ProgressBar` should be `Ti.UI.SIZE`, not `Ti.UI.FILL`. For instance following example should show green background only under the progress bar & text, and it should not fill parent view.

```javascript
var pb = Ti.UI.createProgressBar({
    min: 0,
    max: 10,
    value: 5,
    color: 'blue',
    backgroundColor: 'green',
    backgroundDisabledColor: 'gray',
    message: 'Downloading 0 of 10'
});
var win = Ti.UI.createWindow({backgroundColor:'yellow'});
win.add(pb);
win.open();
pb.show();
```

windows_8_1

### DIFF
--- a/Source/UI/include/TitaniumWindows/UI/ProgressBar.hpp
+++ b/Source/UI/include/TitaniumWindows/UI/ProgressBar.hpp
@@ -53,7 +53,7 @@ namespace TitaniumWindows
 
 			ProgressBar(const JSContext&) TITANIUM_NOEXCEPT;
 
-			virtual ~ProgressBar()                  = default;
+			virtual ~ProgressBar();
 			ProgressBar(const ProgressBar&)            = default;
 			ProgressBar& operator=(const ProgressBar&) = default;
 #ifdef TITANIUM_MOVE_CTOR_AND_ASSIGN_DEFAULT_ENABLE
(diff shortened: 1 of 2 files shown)


yuchi:
I don’t remember the default value of the width of a `ProgressBar` to be `SIZE`.

sgtcoolguy:
@yuchi it appears that SIZE is the default for height and width for iOS: https://github.com/appcelerator/titanium_mobile/blob/1607237763f9e490e41d98b8d711ac49eeb01043/iphone/Classes/TiUIProgressBarProxy.m

And the docs also state that "auto" is "SIZE" for ProgressBar width/height: http://docs.appcelerator.com/platform/latest/#!/guide/Transitioning_to_the_New_UI_Layout_System-section-30088148_TransitioningtotheNewUILayoutSystem-AutoSizeViews

PR #547, opened 2016-02-04
Implement Blob imageAs* methods

infosia:
First cut for [TIMOB-20257](https://jira.appcelerator.org/browse/TIMOB-20257)

```javascript
var blob = Ti.Filesystem.getFile(Ti.Filesystem.applicationDirectory, 'Square150x150Logo.png').read();
var cropped   = blob.imageAsCropped({width:100, height:100, x:50, y:50});
var resized   = blob.imageAsResized(100, 100);
var thumbnail = blob.imageAsThumbnail(50);
```

`imageAsThumbnail` with `borderSize` and `cornerRaidus` is not implemented.

### DIFF
--- a/Source/Titanium/src/Blob.cpp
+++ b/Source/Titanium/src/Blob.cpp
@@ -103,7 +103,7 @@ namespace TitaniumWindows
 		}
 	}
 
-	std::shared_ptr<Titanium::Blob> Blob::transformImage(const std::uint32_t& width, const std::uint32_t height, const Titanium::UI::Dimension& crop) TITANIUM_NOEXCEPT
+	std::shared_ptr<Titanium::Blob> Blob::transformImage(const std::uint32_t& scaledWidth, const std::uint32_t scaledHeight, const Titanium::UI::Dimension& crop) TITANIUM_NOEXCEPT
 	{
 		using namespace Windows::Graphics::Imaging;
 		using namespace Windows::Storage::Streams;
@@ -123,11 +123,11 @@ namespace TitaniumWindows
 			return BitmapDecoder::CreateAsync(instream);
 		}, task_continuation_context::use_arbitrary()).then([outstream](BitmapDecoder^ decoder){
 			return BitmapEncoder::CreateForTranscodingAsync(outstream, decoder);
-		}, task_continuation_context::use_arbitrary()).then([width, height, crop](BitmapEncoder^ encoder){
+		}, task_continuation_context::use_arbitrary()).then([scaledWidth, scaledHeight, crop](BitmapEncoder^ encoder){
 
 			// scaling
-			encoder->BitmapTransform->ScaledWidth  = width;
-			encoder->BitmapTransform->ScaledHeight = height;
+			encoder->BitmapTransform->ScaledWidth  = scaledWidth;
+			encoder->BitmapTransform->ScaledHeight = scaledHeight;
 
 			// cropping
 			BitmapBounds bounds;
@@ -156,8 +156,8 @@ namespace TitaniumWindows
 		auto blob = Blob.CallAsConstructor();
 		auto blob_ptr = blob.GetPrivate<TitaniumWindows::Blob>();
 		blob_ptr->construct(data, mimetype_);
-		blob_ptr->width_  = width;
-		blob_ptr->height_ = height;
+		blob_ptr->width_  = crop.width;
+		blob_ptr->height_ = crop.height;
 
 		return blob.GetPrivate<Titanium::Blob>();
 	}
--- a/Source/Titanium/src/Blob.cpp
+++ b/Source/Titanium/src/Blob.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "TitaniumWindows/Blob.hpp"
+#include "Titanium/UI/Dimension.hpp"
 #include "TitaniumWindows/Utility.hpp"
 #include "TitaniumWindows/GlobalObject.hpp"
 #include <ppltasks.h>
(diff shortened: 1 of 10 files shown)


Source/Titanium/src/Blob.cpp
```diff
@@ -97,4 +102,108 @@ namespace TitaniumWindows
...
+		}
+
+		auto Blob = get_context().CreateObject(JSExport<TitaniumWindows::Blob>::Class());
+		auto blob = Blob.CallAsConstructor();
+		auto blob_ptr = blob.GetPrivate<TitaniumWindows::Blob>();
+		blob_ptr->construct(data, mimetype_);
+		blob_ptr->width_  = crop.width;
+		blob_ptr->height_ = crop.height;
+
+		return blob.GetPrivate<Titanium::Blob>();
```
sgtcoolguy: can't we just return blob_ptr here?

Source/Titanium/src/Blob.cpp
```diff
@@ -97,4 +102,108 @@ namespace TitaniumWindows
...
+	{
+		using namespace Windows::Graphics::Imaging;
+		using namespace Windows::Storage::Streams;
+
+		const auto instream  = ref new InMemoryRandomAccessStream();
+		const auto outstream = ref new InMemoryRandomAccessStream();
+		const auto writer = ref new DataWriter(instream);
+		const auto reader = ref new DataReader(outstream->GetInputStreamAt(0));
+
+		writer->WriteBytes(Platform::ArrayReference<std::uint8_t>(&data_[0], data_.size()));
```
sgtcoolguy: Is there any way to better pipe the data along? Does the ArrayReference avoid duplicating the Blob's data into the stream/writer in-RAM? Just worried we might be holding onto multiple copies of the data/stream at the same time and not cleaning up as we go/minimizing RAM usage so large images might just eat it all/cause a crash?

Source/Titanium/src/Blob.cpp
```diff
@@ -51,8 +57,7 @@ namespace TitaniumWindows
 			std::string contents = global->readRequiredModule(get_object(), path_);
 			auto buffer = CryptographicBuffer::ConvertStringToBinary(TitaniumWindows::Utility::ConvertString(contents), BinaryStringEncoding::Utf8);
 			data_ = TitaniumWindows::Utility::GetContentFromBuffer(buffer);
-		}
-		else {
+		} else {
 			data_ = TitaniumWindows::Utility::GetContentFromFile(file);
```
sgtcoolguy: If we know the content is from a file on disk, can we delay/load the data on-demand? We should probably try and avoid reading the data into memory if we don't need it yet. i.e. if we're going to generate variants on an image using imageAs* methods below, we can point a stream at the file itself as the input stream for the BitmapDecoder - rather than read the file into a buffer/array in memory and then point the BitmapDecoder at the array? (or does the ArrayReference stuff avoid any issues there?

Source/TitaniumKit/include/Titanium/Blob.hpp
```diff
@@ -8,6 +8,7 @@
 #define _TITANIUM_BLOB_HPP_
 
 #include "Titanium/Module.hpp"
+#include "Titanium/UI/Dimension.hpp"
```
sgtcoolguy: can we use a forward declaration of the struct here instead of the include? I believe we can do that in C++11 now...

Source/Titanium/src/Blob.cpp
```diff
@@ -97,4 +102,108 @@ namespace TitaniumWindows
...
+		auto blob = Blob.CallAsConstructor();
+		auto blob_ptr = blob.GetPrivate<TitaniumWindows::Blob>();
+		blob_ptr->construct(data, mimetype_);
+		blob_ptr->width_  = crop.width;
+		blob_ptr->height_ = crop.height;
+
+		return blob.GetPrivate<Titanium::Blob>();
+	}
+
+	std::shared_ptr<Titanium::Blob> Blob::imageAsCropped(const Titanium::UI::Dimension& options) TITANIUM_NOEXCEPT
```
sgtcoolguy: Feels like maybe we should just declare transformImage as a method platform impls will need to implement and move imageAsCropped/imageAsResized (at least) to TitaniumKit level

Source/TitaniumKit/include/Titanium/UI/ViewLayoutDelegate.hpp
```diff
@@ -158,8 +158,8 @@ namespace Titanium
 
 			  Default: 0
 			*/
-			virtual uint32_t get_borderRadius() const TITANIUM_NOEXCEPT;
-			virtual void set_borderRadius(const uint32_t& borderRadius) TITANIUM_NOEXCEPT;
+			virtual double get_borderRadius() const TITANIUM_NOEXCEPT;
```
sgtcoolguy: do we really support non-integer radius values?

Source/Titanium/src/Blob.cpp
infosia: That's a good question. I think Blob can not implement lazy-loading from a File because  Blob is just representing chunk of binary data, and it should be independent from File even when Blob is constructed from a File (`Ti.File read()`). Blob should be able to forget about source of the data, and Blob should not assume File is there once blob is constructed. For instance, what happens when File is deleted after Blob is created? I think Blob should still hold the binary data even in this case.

```
var file = Ti.Filesystem.getFile(...);
var blob = file.read();
var length = blob.length;

file.deleteFile();

// this should work, right?
should(blob.length).be.eql(length);
```

And to realize this, Blob should keep the data in memory.

Source/TitaniumKit/include/Titanium/UI/ViewLayoutDelegate.hpp
infosia: I believe yes, because iOS is [defining it as `float`](https://developer.apple.com/library/ios/documentation/GraphicsImaging/Reference/CALayer_class/#//apple_ref/occ/instp/CALayer/cornerRadius) .

Source/Titanium/src/Blob.cpp
infosia: This sounds making sense to me. Yes I expect `DataWriter::WriteBytes` duplicates copy of the data even when it's created from `ArrayReference`. I wish I could create random access stream from `ArrayReference` directly without copying data, to make it work with `BitmapDecorder.CreateAsync`... but as far as I see there's no such thing in WinRT API! :(

PR #552, opened 2016-02-08
Implement borderRadius for View

infosia:
Fix for [TIMOB-18707](https://jira.appcelerator.org/browse/TIMOB-18707)

```javascript
var window = Ti.UI.createWindow({ backgroundColor: 'green' });
var view = Ti.UI.createView({
    borderRadius: 5,
    borderColor: 'blue',
    borderWidth: 10,
    backgroundColor: 'gray',
    width:  '90%',
    height: '90%',
    layout: 'vertical'
});

    text: 'Hello',
    borderRadius: 2,
    borderColor: 'yellow',
    borderWidth: 2,
    width: '50%',
    textAlign: Ti.UI.TEXT_ALIGNMENT_CENTER
});
view.add(label);

    title: 'Hello',
    borderColor: 'yellow',
    borderWidth: 2
});
view.add(button);

window.add(view);
window.open();
```


Source/UI/src/Window.cpp
```diff
@@ -40,7 +40,8 @@ namespace TitaniumWindows
 			layoutDelegate__->set_defaultHeight(Titanium::UI::LAYOUT::FILL);
 			layoutDelegate__->set_defaultWidth(Titanium::UI::LAYOUT::FILL);
 
-			getViewLayoutDelegate<WindowsViewLayoutDelegate>()->setComponent(canvas__);
+			// Window is considered as "root" of the view and it does'nt support border.
```
sgtcoolguy: doesn't

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -698,19 +695,22 @@ namespace TitaniumWindows
...
-					auto color = get_borderColor().empty() ? get_backgroundColor() : get_borderColor();
-					if (color.empty()) {
-						color = "black";
-					}
-					borderColorBrush__ = ref new SolidColorBrush(ColorForName(color));
-					control->BorderBrush = borderColorBrush__;
-				}
+				// Xaml::Control descendant has its own border property. 
+				// Use it then, it usually works better than Xaml::Border. Note that it doesn't support border radius though...
+				const auto control = dynamic_cast<Control^>(component__);
```
sgtcoolguy: I was going to note that we'd need to set both border width and color for a colored border to show up this way, but given that the Ti API assumes a width default of 0, they'll need to anyways...

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -1016,11 +1010,17 @@ namespace TitaniumWindows
...
 		}
 
-		void WindowsViewLayoutDelegate::setComponent(FrameworkElement^ component)
+		void WindowsViewLayoutDelegate::setComponent(FrameworkElement^ component, const bool& enableBorder)
 		{
 			TITANIUM_ASSERT(component__ == nullptr);
 
 			component__  = component;
+
+			if (enableBorder) {
```
sgtcoolguy: Don't we pretty much _always_ want to enable a border? borderRadius is a property on View, so all UI classes can have it...

infosia: > Don't we pretty much always want to enable a border? 

Yes, as long as it doesn't cause runtime/compile error. I'm still testing but luckily seems like most of components are eligible to have Xaml `Border` as "parent" even when we disables border with width=0. The reason I use `enableBorder` is that only `Ti.Window` doesn't accept `Border`. It looks like Xaml root frame (`Xaml::Controls::Frame` doesn't accept `Border` as a page content).

infosia: > borderRadius is a property on View, so all UI classes can have it...

Ah I see, we should be able to always enclose with Xaml `Border`. This sounds making sense, but the reason I prefer Xaml `Control`'s border property over Xaml `Border` is that I'm observing some controls don't render well `Border`. For instance, `Button` doens't render well with `Border`. I'll take screenshots to show how it looks.

infosia:
Here's a screenshot when you use `Xaml::Controls::Border` over `Xaml::Controls::Button`. It doesn't actually renders correct hit area. Also, I found that `Border::borderRadius` doesn't work even in this case.

button

infosia:
Updated, tested on all components. 

All of views except `View` doesn't support `borderRadius`. Windows doesn't render those well even when they're enclosed by Xaml `Border`. For Xaml `Control` classes they have their own `Border` property, but some components doesn't render border or break rendering. For `Window`, you can't enable border because it causes runtime error. For `ListView` and `TableView`, list item disappears when you enable border property.

component | borderColor | borderWidth | borderRadius
------------ | ------------- | ------------ | -------------
View | :o: | :o: | :o:
ActivityIndicator |  :o: | :o: | :x:
Button |  :o: | :o: | :x:
ImageView | :o: | :o: | :x:
Label | :o: | :o: | :x:
ScrollView | :o: | :o: | :x:
ScrollableView | :o: | :o: | :x:
Slider |  :o: | :o: | :x:
Switch |  :o: | :o: | :x:
TextArea |  :o: | :o: | :x:
TextField |  :o: | :o: | :x:
ProgressBar |  :o: | :o: | :x:
WebView |  :o: | :o: | :x:
Window | :x: | :x: | :x: 
ListView | :x: | :x: | :x: 
TableView | :x: | :x: | :x: 

Sample screenshot:
border1

sgtcoolguy:
@infosia Thanks for checking all these out! My concern here is that borderRadius doesn't work for ImageView/Button. In particular I think it's fairly common for people to use borderRadius to basically turn a square image into a circular one (or at least rounded off), or to create circular buttons.

For example, the corporate directory example app uses both use cases, see the screenshots here: https://github.com/appcelerator-se/corporate-directory

it would appear that setting CornerRadius should work for at least buttons:
 - http://windowsapptutorials.com/windows-phone/how-to-create-an-image-round-button-in-windows-phone-app/
 - http://stackoverflow.com/questions/16341788/how-to-clip-content-with-rounded-corners-in-windows-store-app

infosia:
> it would appear that setting CornerRadius should work for at least buttons:

Yeah, I saw some articles mention styling round border on `Button` using `Border`... but as far as I tried, it doesn't work well. For instance setting `Border` around `Button` ends up not only rendering wrong areas but also `CornerRadius` doesn't work.  :disappointed_relieved: I suspect there are some incompatibilities between WPF, WP 8.0 and WP 8.1 (+ Win10 Mobile).

button

sgtcoolguy:
@infosia so from a quick test of a Win 10 universal app in designer, I added a Border and a Button inside it. Then you need to start tweaking both the Border and button to avoid double borders. So you set BorderThickness to 0 on the button to remove it's natural one; then you set the BorderThickness on the Border itself instead. You set the width/height on the Border, and set the Button to "Stretch" (fill) horizontally and vertically. You can then set a CornerRadius on the Border to make it rounded.

Here's the raw XAML xml and what it looks like in designer:
```xml
 <Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
        <Border BorderBrush="Black" CornerRadius="20" BorderThickness="10" HorizontalAlignment="Left" Height="100" Margin="133,253,0,0" VerticalAlignment="Top" Width="100">
            <Button x:Name="button" Content="Push Me!" HorizontalAlignment="Stretch" VerticalAlignment="Stretch" BorderThickness="0"/>
        </Border>
    </Grid>
```
Here's what it looks like in designer:
capture


And on the phone emulator:
emulator

Now what I do see that isn't nice, is if you get to a round enough radius to do circles with a button inside, the button's corners will "poke through":
poke
In this case, I think you could get around it by matching the button bg color to the surrounding bg color? From looks of it, the grid is white, but the button has a 20% alpha black bg color set.

sgtcoolguy:
@infosia Ok, so I also found a somewhat hacky way to do rounded corner images as well. If I place an Image inside a Border, the image draws over the top of the border itself. Not good.

But you can generate a Border and set it's background brush to an ImageBrush pointing at an image source:

```xml
<Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
        <Border BorderBrush="Black" CornerRadius="50" BorderThickness="10" HorizontalAlignment="Left" Height="100" Margin="133,253,0,0" VerticalAlignment="Top" Width="100">
            <Border.Background>
                <ImageBrush Stretch="Fill" ImageSource="http://api.randomuser.me/portraits/women/1.jpg"/>
            </Border.Background>
        </Border>
    </Grid>
```
Designer:
border_image_design

Emulator:
border_image_emulator

Now, I think this actually isn't too bad and could work for ImageView when using border* properties, and NOT doing animations. But I think if you're using the "images" property to use a set of images in an animation, that likely wouldn't play nicely with this approach. Which seems like a reasonable limitation: either use a static image and we can do border, or use a series of images and you can't do borders but can do animations.

infosia:
@sgtcoolguy wow awesome ideas! I'll give it a shot.

infosia:
I've been trying to mimic the way XAML xml does by doing it programatically. I started from `Button`, but making `Grid->Border->Button` layout tree renders "clicked area" incorrectly. For instance following example shows what happens when you click the button with following settings.

```c++
void Button::postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments)
{
	Titanium::UI::Button::postCallAsConstructor(js_context, arguments);	
	Titanium::UI::Button::setLayoutDelegate<WindowsViewLayoutDelegate>();

	button__ = ref new Controls::Button();
	button__->VerticalAlignment = VerticalAlignment::Stretch;
	button__->HorizontalAlignment = HorizontalAlignment::Stretch;
	button__->BorderThickness = 0;

	border__ = ref new Controls::Border();
	border__->Child = button__;
	border__->CornerRadius = 30;
	border__->BorderThickness = 10;

	parent__ = ref new Controls::Grid();
	parent__->Children->Append(border__);
	parent__->SetColumn(border__, 0);
	parent__->SetRow(border__, 0);

	getViewLayoutDelegate<WindowsViewLayoutDelegate>()->setComponent(parent__);
}
```

windows_8_1

we want workaround this by changing color of the area, or making it transparent maybe?

infosia:
I guess it's not a good idea to disable color of the "clicked area" because that way user doesn't get the response of the interaction...




Not so bad?

sgtcoolguy:
@infosia do we need the Border to be contained in a Grid to layout properly? Can we just use the Border with a child button?

So your comments above are about how it renders when you are clicking the button? The blue color is the "clicked" state? It's certainly not pretty, but I'm not sure what we can really do about it other than hacking the click event ourselves to listen and fill the Border bg color/brush with the click bg color of the button entirely (and then switch it back after the click state goes away). I'd say for now let's live with the ugliness it causes on click and open another ticket to make that nicer looking by investigating doing a bg fill of the parent Border during click state.

infosia:
@sgtcoolguy,

> do we need the Border to be contained in a Grid to layout properly? Can we just use the Border with a child button?

Yeah, but I observed `borderRadius` doesn't work in that case. It's kind of weird. :(

> So your comments above are about how it renders when you are clicking the button? The blue color is the "clicked" state?

Yes. ok, let's go on this way for now!

infosia:
build win 8.1

infosia:
Updated, I was able to implement `borderRadius` for `Label`, `ImageView` and `Button`.

component | borderColor | borderWidth | borderRadius
------------ | ------------- | ------------ | -------------
View | :o: | :o: | :o:
Button |  :o: | :o: | :o:
ImageView | :o: | :o: | :o:
Label | :o: | :o: | :o:
ActivityIndicator |  :o: | :o: | :x:
ScrollView | :o: | :o: | :x:
ScrollableView | :o: | :o: | :x:
Slider |  :o: | :o: | :x:
Switch |  :o: | :o: | :x:
TextArea |  :o: | :o: | :x:
TextField |  :o: | :o: | :x:
ProgressBar |  :o: | :o: | :x:
WebView |  :o: | :o: | :x:
Window | :x: | :x: | :x: 
ListView | :x: | :x: | :x: 
TableView | :x: | :x: | :x: 

corporate

button

Should be ready once Jenkins passes this.

PR #553, opened 2016-02-08
[TIMOB-20370] Windows: Running desktop app doesn't show logs in console

sgtcoolguy:
Requires appcelerator/windowslib#31 be merged, version bumped and released and then put in appcelerator/titanium_mobile master branch's package.json and merged.

JIRA ticket: https://jira.appcelerator.org/browse/TIMOB-20370

Basically when you launched a desktop app before it'd just launch it and close the CLI process. Now we don't exit and we do some magic voodoo to make the app exempt from the network isolation loopback IP blocking that's done by default for Windows Runtime apps. (See https://msdn.microsoft.com/en-us/library/windows/apps/hh780593.aspx )

So there is still a bit of a difference between phone emulator and desktop apps in that we poll constantly to know when the emulator is killed and then kill the CLI process. We DO NOT do the same thing for desktop apps yet. We likely should.

### DIFF
--- a/cli/hooks/ws-run.js
+++ b/cli/hooks/ws-run.js
@@ -2,7 +2,7 @@
  * Runs an app on a Windows Store emulator.
  *
  * @copyright
- * Copyright (c) 2015 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2015-2016 by Appcelerator, Inc. All Rights Reserved.
  *
  * @license
  * Licensed under the terms of the Apache Public License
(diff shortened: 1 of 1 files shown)


cli/hooks/ws-run.js
```diff
@@ -160,25 +160,38 @@ exports.init = function (logger, config, cli) {
...
 					logger.info(__('Installing the application'));
 					windowslib.winstore.install(projectDir, opts, next);
+				}, function (next) {
+					logger.info(__('Making the application exempt from loopback IP restrictions for logging'));
+					windowslib.winstore.getAppxPackages(opts, function (err, packages) {
+						if (err) {
+							return next(err);
+						}
+
+						appc.subprocess.run('CheckNetIsolation.exe', ['LoopbackExempt', '-a', '-n=' + packages[appId].PackageFamilyName], function (code, out, err) {
```
infosia: I think we want to check if `packages` contains `appId` property and return appropriate error message if not there.

Why don't you include this logic into `windowslib` by the way?

sgtcoolguy: You'e right, I moved it to windowslib: appcelerator/windowslib#32

infosia:
Awesome, this works for me with https://github.com/appcelerator/windowslib/pull/32. ok to merge once https://github.com/appcelerator/windowslib/pull/32 merged.

PR #556, opened 2016-02-11
Fix TIMOB-20382 - font caching was broken on lookups of cached font

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-20382


infosia:
Looks good, merging.

PR #559, opened 2016-02-12
DO NOT MERGE - Run unit tests on store app (ws-local)

sgtcoolguy:
This is not intended to be merged!

This changes our batch scripts to only build store libraries and run the unit tests against ws-local target. This is a sanity check to see if it works at all before I fold it into the full build.

https://jira.appcelerator.org/browse/TIMOB-20263


sgtcoolguy:
build win 8.1

sgtcoolguy:
merged manually.

PR #563, opened 2016-02-18
Make sure .winmd has at least one public entry point

infosia:
[TIMOB-20192](https://jira.appcelerator.org/browse/TIMOB-20192)

Attempt to fix "submission error 2001" described in https://msdn.microsoft.com/en-us/library/windows/apps/mt445539.aspx.

This PR just tries to make sure each `.winmd` file has at least one entry point (`ref class`). To test this, use [Ildasm.exe](https://msdn.microsoft.com/en-us/library/aa309387%28v=vs.71%29.aspx) to find a entry point in each `winmd` files. The other way to test is that usually `winmd` files are nearly `1KB` when it has no entry but it should be over `3KB` or so when it has an entry. Note that from what I saw only `TitaniumWindows_*.DLL` has its `winmd`, and `HAL`, `LayoutEngine` and `TitaniumKit` doesn't have it. (I expect it's because they do not use any Windows API?)

FYI `Ildasm.exe` is located under `c:\Program Files(x86)\Microsoft SDKs\Windows\v8.1A\bin\NETFX 4.1.5 Tools\` for me on Windows 10.

### DIFF
--- a/Source/Filesystem/CMakeLists.txt
+++ b/Source/Filesystem/CMakeLists.txt
@@ -48,6 +48,7 @@ endif()
 set(SOURCE_Filesystem
   include/TitaniumWindows/Filesystem.hpp
   src/Filesystem.cpp
+  src/TitaniumEntryPoint.cpp
   )
 
 source_group(Titanium\\Filesystem FILES ${SOURCE_Filesystem})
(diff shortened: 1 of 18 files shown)


sgtcoolguy:
build win 8.1

PR #566, opened 2016-02-23
Implement Background Service

infosia:
First cut for background service implementation. [TIMOB-18952](https://jira.appcelerator.org/browse/TIMOB-18952)

Initial goal is to enable HTTP access and geolocation in background.
Here's a sample code:

```javascript
    backgroundColor: 'blue'
});
var view = Ti.UI.createView({ layout: 'vertical' });

//
// Enable location tracking in background
//
Ti.App.Windows.BackgroundService.locationServiceEnabled = true;

//
// Callback for background task
// Note that you can't use setTimeout/setInterval in background, as well as any UI interaction (of course).
// Also, note that Ti.Geolocation events does not work in background. Use locationServiceEnabled instead.
//
var callback = function (e) {
    //
    // e.source ... task object that invokes this event
    // e.cancel ... indicate that this task is canceled
    // e.location ... Current location. Available only when locationServiceEnabled is true.
    // 
    //
    if (e.canceled) {
        Ti.API.info("Task canceled");
    } else {
        Ti.API.info("Do something useful");
        if (e.location) {
            // location contains latitude, longitude etc.
            // See LocationCoodinates for details (http://docs.appcelerator.com/platform/latest/#!/api/LocationCoordinates)
            Ti.API.info('Current geolocation: ' + JSON.stringify(e.location));
        }
        Ti.API.info("Get something via HTTP then");
        var url = "http://www.appcelerator.com";
            onload: function (e) {
                Ti.API.info("Received data size: " + this.responseData.size);
            },
            onerror: function (e) {
                Ti.API.info('Error: ' + e.error);
            },
            timeout: 5000
        });
        client.open("GET", url);
        client.send();
    }
};

var task;
var registerBtn = Ti.UI.createButton({
    title: "register timer task"
});
var unregisterBtn = Ti.UI.createButton({
    title: "unregister timer task"
});
var unregisterAllBtn = Ti.UI.createButton({
    title: "unregister ALL tasks"
});

registerBtn.addEventListener('click', function () {
    //
    // BackgroundService.registerTimerTask(callback, interval_minutes, is_one_shot);
    //
    // Create timer task, that repeats at 15-minute intervals.
    // Note that Windows doesn't accept interval less than 15 minutes.
    //
    task = Ti.App.Windows.BackgroundService.registerTimerTask(callback, 15, false);
});

unregisterBtn.addEventListener('click', function () {
    //
    // When you unregister the task, task invokes callback with canceled = true.
    //
    task.unregister();
});

unregisterAllBtn.addEventListener('click', function () {
    //
    // Unregister all tasks that is associated with this application.
    //
    Ti.App.Windows.BackgroundService.unregisterAllTasks();
});

view.add(registerBtn);
view.add(unregisterBtn);
view.add(unregisterAllBtn);

win.add(view);
win.open();
```

##### IMPORTANT NOTE

To enable background service, both Windows Phone and Windows Store apps require an `timer` BackgroundTask extension as well as `internetClient` and `location` capabilities in the `tiapp.xml`. (See [Windows-specific configuration](https://wiki.appcelerator.org/display/guides2/tiapp.xml+and+timodule.xml+Reference#tiapp.xmlandtimodule.xmlReference-Windows-specificapplicationproperties) for details) Also, when you use geolocation, make sure your devices enables GPS.

###### tiapp.xml Background  Task Extension

```xml
<windows>
  <manifest>
    <Extensions>
        <Extension Category="windows.backgroundTasks" EntryPoint="TitaniumWindows_Ti.BackgroundServiceTask">
          <BackgroundTasks>
            <Task Type="timer" />
          </BackgroundTasks>
        </Extension>
    </Extensions>
    <Capabilities>
      <DeviceCapability Name="location" />
    </Capabilities>
  </manifest>
</windows>
```

### DIFF
--- a/Examples/NMocha/src/Assets/ti.app.windows.backgroundservice.test.js
+++ b/Examples/NMocha/src/Assets/ti.app.windows.backgroundservice.test.js
@@ -1,0 +1,44 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2011-2016 by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+var should = require('./should');
+
+describe('Ti.App.Windows.BackgroundService', function () {
+
+    it('API', function (finish) {
+        should(Ti.App.Windows).be.an.Object;
+        should(Ti.App.Windows.BackgroundService).be.an.Object;
+        should(Ti.App.Windows.BackgroundService.registerTimerTask).be.a.Function;
+        should(Ti.App.Windows.BackgroundService.unregisterAllTasks).be.a.Function;
+        finish();
+    });
+
+    it('locationServiceEnabled', function (finish) {
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.Boolean;
+        Ti.App.Windows.BackgroundService.locationServiceEnabled = true;
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.true;
+        Ti.App.Windows.BackgroundService.locationServiceEnabled = false;
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.false;
+        finish();
+    });
+
+    it('registerTimerTask', function (finish) {
+        var task = Ti.App.Windows.BackgroundService.registerTimerTask(function() {
+            Ti.API.info('registerTimerTask');
+        }, 15, true);
+        task.unregister();
+        finish();
+    });
+
+    it('unregisterAllTasks', function (finish) {
+        var task = Ti.App.Windows.BackgroundService.registerTimerTask(function() {
+            Ti.API.info('unregisterAllTasks');
+        }, 15, true);
+        Ti.App.Windows.BackgroundService.unregisterAllTasks();
+        finish();
+    });
+
+});
--- a/Source/Network/include/TitaniumWindows/Network/HTTPClient.hpp
+++ b/Source/Network/include/TitaniumWindows/Network/HTTPClient.hpp
@@ -56,7 +56,6 @@ namespace TitaniumWindows
 			virtual void setRequestHeader(const std::string& name, const std::string& value) TITANIUM_NOEXCEPT override final;
 			// properties
 			virtual std::string get_allResponseHeaders() const TITANIUM_NOEXCEPT override final;
-			virtual void set_timeout(const std::chrono::milliseconds& timeout) TITANIUM_NOEXCEPT override final;
 			
 		protected:
 #pragma warning(push)
@@ -72,14 +71,10 @@ namespace TitaniumWindows
 			concurrency::cancellation_token_source cancellationTokenSource__;
 			//  timeoutRegistrationToken__ - used to remove the timeout handler
 			Windows::Foundation::EventRegistrationToken timeoutRegistrationToken__;
-			// dispatcherTimer__ - WinRT timer used to cancel a HTTP connection if timeout specified
-			Windows::UI::Xaml::DispatcherTimer^ dispatcherTimer__;
 			// responseStream__ - holds response string, the stream is not exposed
 			Windows::Storage::Streams::IBuffer^ responseStream__;
 			// responseDataLen__ - count of character contained in data vector
 			long responseDataLen__ { 0 };
-			// timeoutSpan__ - the span in milliseconds during which the request is active
-			Windows::Foundation::TimeSpan timeoutSpan__;
 			// responseHeaders__ - the collection of key value pairs returned from the server
 
 			std::map<const std::string, std::string> responseHeaders__;
@@ -92,7 +87,7 @@ namespace TitaniumWindows
 			bool disposed__ { false };
 #pragma warning(pop)
 
-			void startDispatcherTimer();
+			void configureTimeout();
 			task<Windows::Storage::Streams::IBuffer^> HTTPClient::HTTPResultAsync(Windows::Storage::Streams::IInputStream^ stream, concurrency::cancellation_token token);
 			Windows::Storage::Streams::Buffer^ charVecToBuffer(std::vector<std::uint8_t> char_vector);
 
--- a/Source/Sensors/src/Geolocation.cpp
+++ b/Source/Sensors/src/Geolocation.cpp
@@ -64,18 +64,26 @@ namespace TitaniumWindows
 				const auto latitude = args->Position->Coordinate->Point->Position.Latitude;
 				const auto longitude = args->Position->Coordinate->Point->Position.Longitude;
 				const auto altitude = args->Position->Coordinate->Point->Position.Altitude;
-				const auto altitudeAccuracy = args->Position->Coordinate->AltitudeAccuracy->Value;
-				const auto speed = args->Position->Coordinate->Speed->Value;
+				const auto altitudeAccuracy = args->Position->Coordinate->AltitudeAccuracy;
+				const auto speed = args->Position->Coordinate->Speed;
 
-				Utility::RunOnUIThread([=](){
-					lastGeolocation__.SetProperty("latitude", get_context().CreateNumber(latitude));
-					lastGeolocation__.SetProperty("longitude", get_context().CreateNumber(longitude));
-					lastGeolocation__.SetProperty("altitude", get_context().CreateNumber(altitude));
-					lastGeolocation__.SetProperty("altitudeAccuracy", get_context().CreateNumber(altitudeAccuracy));
-					lastGeolocation__.SetProperty("speed", get_context().CreateNumber(speed));
+				const auto func = [=](){
+					const auto ctx = get_context();
+					lastGeolocation__.SetProperty("latitude", ctx.CreateNumber(latitude));
+					lastGeolocation__.SetProperty("longitude", ctx.CreateNumber(longitude));
+					lastGeolocation__.SetProperty("altitude", ctx.CreateNumber(altitude));
+					lastGeolocation__.SetProperty("altitudeAccuracy", altitudeAccuracy ? ctx.CreateNumber(altitudeAccuracy->Value) : ctx.CreateNumber(0));
+					lastGeolocation__.SetProperty("speed", speed ? ctx.CreateNumber(speed->Value) : ctx.CreateNumber(0));
 
 					this->fireEvent("location", lastGeolocation__);
-				});
+				};
+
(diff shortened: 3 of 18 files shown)


Source/Network/include/TitaniumWindows/Network/HTTPClient.hpp
```diff
@@ -72,14 +71,10 @@ namespace TitaniumWindows
 			concurrency::cancellation_token_source cancellationTokenSource__;
 			//  timeoutRegistrationToken__ - used to remove the timeout handler
 			Windows::Foundation::EventRegistrationToken timeoutRegistrationToken__;
-			// dispatcherTimer__ - WinRT timer used to cancel a HTTP connection if timeout specified
-			Windows::UI::Xaml::DispatcherTimer^ dispatcherTimer__;
```
infosia: Note: Basically eliminating `Windows::UI::Xaml::DispatcherTimer` because it doesn't work when app is in background.

Source/Sensors/src/Geolocation.cpp
```diff
@@ -64,18 +64,26 @@ namespace TitaniumWindows
...
 
 					this->fireEvent("location", lastGeolocation__);
-				});
+				};
+
+				// Check if we are in background, in that case we do not run it on UI thread.
+				if (static_cast<bool>(get_context().JSEvaluateScript("Ti.App.Windows.BackgroundService.suspended;"))) {
+					func();
+				} else {
+					Utility::RunOnUIThread(func);
```
infosia: Note: Making sure run callback on UI thread only when app is foreground.

Source/Ti/src/App/BackgroundService.cpp
```diff
@@ -0,0 +1,290 @@
...
+	void BackgroundServiceTask::Run(IBackgroundTaskInstance^ taskInstance)
+	{
+		const auto deferral = Platform::Agile<BackgroundTaskDeferral>(taskInstance->GetDeferral());
+		const auto location_enabled = TitaniumWindows::App::WindowsXaml::BackgroundService::Get_LocationServiceEnabled();
+		if (location_enabled) {
+
+			const auto geolocator_ = ref new Geolocator();
+			geolocator_->MovementThreshold = 1;
+			geolocator_->ReportInterval = 0;
+			geolocator_->DesiredAccuracy = PositionAccuracy::High;
```
infosia: Note: `Ti.Geolocation` doesn't work here, seems like background task is executed in different process. In that sense we share information between app and background task using `LocalSettings` .

Source/Ti/src/App/BackgroundService.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+namespace TitaniumWindows
+{
+	namespace App
+	{
+		namespace WindowsXaml 
+		{
+			Platform::String^ BackgroundService::TASK_ENTRYPOINT_NAME { "TitaniumWindows_Ti.BackgroundServiceTask" };
+			Platform::String^ BackgroundService::LAST_GEOLOCATION_KEY { "TitaniumWindows_Ti.BackgroundServiceTask.LastGeolocation" };
+			Platform::String^ BackgroundService::GEOLOCATION_ENABLED_KEY{ "TitaniumWindows_Ti.BackgroundServiceTask.GeolocationEnabled" };
```
infosia: Note: This is an actual key in `LocalSettings`. We might want to prefix something like "titanium." to indicate it's for internal use?

infosia:
build win 10

Source/Ti/include/TitaniumWindows/App/BackgroundServiceTask.hpp
```diff
@@ -0,0 +1,78 @@
...
+				void construct(const std::uint32_t& id, BackgroundTaskRegistration^ registration, JSObject& callback);
+				void unregister();
+				void taskCompleted(const bool& canceled = false);
+
+				/*!
+				  @property
+				  @abstract taskId
+				  @discussiona taskId that is associated with this task. Read-Only.
+				*/
+				TITANIUM_PROPERTY_IMPL_READONLY_DEF(std::uint32_t, taskId);
```
sgtcoolguy: Looks to be the equivalent of Ti.Android.Service.serviceInstanceId?

Source/Ti/include/TitaniumWindows/App/BackgroundServiceTask.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+				/*!
+				  method
+				  @abstract unregister
+				  @discussiona Unregisters this background task.
+				*/
+				TITANIUM_FUNCTION_DEF(unregister);
```
sgtcoolguy: So, looks like for Android we can start/stop a service instance, and can register/unregister broadcast receivers at the Ti.Android API level (or start and stop a service as well).

Is unregister() the equivalent of stopping an Android service? Just curious how much we can align the APIs to have a more cross-platform solution eventually...

Source/Ti/include/TitaniumWindows/AppModule.hpp
```diff
@@ -44,6 +44,8 @@ namespace TitaniumWindows
 		AppModule& operator=(AppModule&&) = default;
 #endif
 
+		TITANIUM_PROPERTY_READONLY_DEF(Windows);
```
sgtcoolguy: So, right now the APIs to create a service on Android live off Ti.Android, but it looks like you're hanging them off Ti.App.Windows for the Win equivalent? We may want to move it up to just Ti.Windows

sgtcoolguy: Ah, now I see. The iOS API lives off Ti.App.iOS. Hmm. Looks like we already have an inconsistency in our APIs here anyways. So we may not need to deal with it here/now, but should likely open a ticket for the SDK teams to try and come up with a more standardized API for services across platforms.

Source/Ti/src/App/BackgroundService.cpp
sgtcoolguy: hmm, yeah good question. Not sure if we store internal keys in Android intent usage on the Android side and could emulate that? cc @hieupham007 

Otherwise, I'd do something like Titanium.App.Windows.BackgroundService.* as prefix, whatever the surfaced API namespace is...

Source/Ti/include/TitaniumWindows/App/BackgroundServiceTask.hpp
sgtcoolguy: Since I see this API is a little more closely aligned to the iOS one, do we also want a stop() method? (or can we even do one?)

Source/Ti/src/App/BackgroundService.cpp
```diff
@@ -0,0 +1,290 @@
...
+					hasCurrent = iter->MoveNext();
+				}
+
+				return nullptr;
+			}
+
+			void BackgroundService::UnregisterTask(const std::uint32_t& id)
+			{
+				const auto taskName = Windows::ApplicationModel::Package::Current->Id->Name + "." + id;
+				const auto iter = BackgroundTaskRegistration::AllTasks->First();
```
sgtcoolguy: There's no handy for loop for this type of iterator?

Source/TitaniumKit/src/GeolocationModule.cpp
```diff
@@ -56,151 +91,151 @@ namespace Titanium
...
 	TITANIUM_PROPERTY_GETTER(GeolocationModule, PROVIDER_GPS)
 	{
-			return provider_gps__;
+		return get_context().CreateNumber(Titanium::Geolocation::Constants::to_underlying_type(Titanium::Geolocation::PROVIDER::GPS));
 	}
 
 	TITANIUM_PROPERTY_GETTER(GeolocationModule, PROVIDER_NETWORK)
 	{
-			return provider_network__;
+		return get_context().CreateNumber(Titanium::Geolocation::Constants::to_underlying_type(Titanium::Geolocation::PROVIDER::NETOWORK));
```
sgtcoolguy: Typo in the constant name, should be `Titanium::Geolocation::PROVIDER::NETWORK`

Source/Ti/src/App/BackgroundService.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+				pos.timestamp = std::chrono::milliseconds(static_cast<std::chrono::milliseconds::rep>(values->Data[8]));
+
+				return pos;
+			}
+
+			TITANIUM_FUNCTION(BackgroundService, registerTimerTask)
+			{
+				ENSURE_OBJECT_AT_INDEX(callback, 0);
+				ENSURE_OPTIONAL_UINT_AT_INDEX(freshnessTime, 1, 15);
```
sgtcoolguy: Do we want to do more enforcement here to verify that the value is a minimum of 15, and if it isn't we log a warning and make it 15 minutes?

Examples/NMocha/src/Assets/ti.app.windows.backgroundservice.test.js
```diff
@@ -0,0 +1,44 @@
...
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.Boolean;
+        Ti.App.Windows.BackgroundService.locationServiceEnabled = true;
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.true;
+        Ti.App.Windows.BackgroundService.locationServiceEnabled = false;
+        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.false;
+        finish();
+    });
+
+    it('registerTimerTask', function (finish) {
+        var task = Ti.App.Windows.BackgroundService.registerTimerTask(function() {
```
sgtcoolguy: The API differs from both Android and iOS in that it takes a callback function, whereas they take a JS object with an "url" property which has a value with a relative URL to a JS file defining the service. We may want to eventually along them closer.

sgtcoolguy:
Looks good for a first cut. I think as a whole we need to try and see if we can't align iOS/Android/Windows more closely with regard to services, the APIs between them are pretty divergent right now. We should also ask Jurgen to try out a master build of the SDK after this lands to get his feedback and give him some sample usage to try out - and we can see if it'll meet his needs.

Buder:
Reading though your comments, just FYI, testing now with Phone 8.1 will let you know. 

But one thing I will be needing is a notification API, I looked at the parity its from master but I cannot find any way what so ever to create a notification, not local or from push.. (if push I need to register a callback that handles the push from servers), any idea or am I missing something?)
I know that is not part of the ticket but for me to have a working background service I have to be able to notify the user that there is new information that arrived, this is done from the HTTP call where I report the location and get a reply if there is new news to be notified to user.

thanks

PR #571, opened 2016-02-29
Implement Ti.Locale properties

infosia:
[TIMOB-20478](https://jira.appcelerator.org/browse/TIMOB-20478)

Implement Ti.Locale properties:

- `Ti.Locale.currentCountry`
- `Ti.Locale.currentLanguage`
- `Ti.Locale.currentLocale`

### DIFF
--- a/Source/Ti/include/TitaniumWindows/Locale.hpp
+++ b/Source/Ti/include/TitaniumWindows/Locale.hpp
@@ -23,10 +23,21 @@ namespace TitaniumWindows
 	class TITANIUMWINDOWS_TI_EXPORT Locale final : public Titanium::Locale, public JSExport<Locale>
 	{
 	public:
-		TITANIUM_FUNCTION_UNIMPLEMENTED(getCurrencyCode);
-		TITANIUM_FUNCTION_UNIMPLEMENTED(getCurrencySymbol);
-		TITANIUM_FUNCTION_UNIMPLEMENTED(getLocaleCurrencySymbol);
 		TITANIUM_FUNCTION_UNIMPLEMENTED(formatTelephoneNumber);
+
+		/*!
+		@method
+		@abstract getCurrencyCode
+		@discussion Returns the ISO 3-letter currency code for the specified locale.
+		*/
+		virtual std::string getCurrencyCode(const std::string& locale) TITANIUM_NOEXCEPT;
+
+		/*!
+		@method
+		@abstract getCurrencySymbol
+		@discussion Returns the currency symbol for the specified currency code.
+		*/
+		virtual std::string getCurrencySymbol(const std::string& currencyCode) TITANIUM_NOEXCEPT;
 
 		/*!
 		  @method
--- a/Source/Ti/include/TitaniumWindows/Locale.hpp
+++ b/Source/Ti/include/TitaniumWindows/Locale.hpp
@@ -23,18 +23,17 @@ namespace TitaniumWindows
 	class TITANIUMWINDOWS_TI_EXPORT Locale final : public Titanium::Locale, public JSExport<Locale>
 	{
 	public:
-		TITANIUM_PROPERTY_UNIMPLEMENTED(currentCountry);
-		TITANIUM_PROPERTY_UNIMPLEMENTED(currentLanguage);
-		TITANIUM_PROPERTY_UNIMPLEMENTED(getCurrencyCode);
+		TITANIUM_FUNCTION_UNIMPLEMENTED(getCurrencyCode);
 		TITANIUM_FUNCTION_UNIMPLEMENTED(getCurrencySymbol);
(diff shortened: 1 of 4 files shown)


Source/Ti/src/Locale.cpp
```diff
@@ -14,6 +17,16 @@ namespace TitaniumWindows
 		: Titanium::Locale(js_context)
 	{
 		TITANIUM_LOG_DEBUG("TitaniumWindows::Locale::ctor");
+
+		currentCountry__ = TitaniumWindows::Utility::ConvertString(GlobalizationPreferences::HomeGeographicRegion);
```
sgtcoolguy: HomeGeographicRegion is documented as just the identifier for the GeographicRegion the user has set. It doesn't actually say what sort of format it's in. The docs for GeographicRegion specify that it can be constructed with a 2-letter string, a 3-letter strong, or a 3-digit code.

I think we _may_ want to covert this so a GeographicRegion object and then explicitly grab the 2-letter code since our docs state that's the value we return for currentCountry.

See https://msdn.microsoft.com/en-us/library/windows/apps/windows.globalization.geographicregion

Source/Ti/src/Locale.cpp
```diff
@@ -14,6 +17,16 @@ namespace TitaniumWindows
...
 		TITANIUM_LOG_DEBUG("TitaniumWindows::Locale::ctor");
+
+		currentCountry__ = TitaniumWindows::Utility::ConvertString(GlobalizationPreferences::HomeGeographicRegion);
+		currentLocale__  = TitaniumWindows::Utility::ConvertString(GlobalizationPreferences::Languages->GetAt(0));
+
+		const auto pos = currentLocale__.find("-");
+		if (pos != std::string::npos) {
+			currentLanguage__ = currentLocale__.substr(0, pos);
+		} else {
+			currentLanguage__ = currentLocale__;
```
sgtcoolguy: If the language value has no dash, it likely holds just the language, not the country code. In that case we likely should do this for the else block:
```
currentLocale__ =  "-" + currentCountry_;
```

Source/Ti/include/TitaniumWindows/Locale.hpp
```diff
@@ -23,18 +23,17 @@ namespace TitaniumWindows
 	class TITANIUMWINDOWS_TI_EXPORT Locale final : public Titanium::Locale, public JSExport<Locale>
 	{
 	public:
-		TITANIUM_PROPERTY_UNIMPLEMENTED(currentCountry);
-		TITANIUM_PROPERTY_UNIMPLEMENTED(currentLanguage);
-		TITANIUM_PROPERTY_UNIMPLEMENTED(getCurrencyCode);
+		TITANIUM_FUNCTION_UNIMPLEMENTED(getCurrencyCode);
```
sgtcoolguy: So I know this is a nice simple PR that improves our API coverage, but it made me look deeper to see if we could do more.

It seems like `getLocaleCurrencySymbol( locale )` should be implemented at the TitaniumKit level to just call:
```
getCurrencySymbol(getCurrencyCode(locale));
```

That way it's one more method we don't need to override.

Also, it looks to me like we should be able to fairly easily implement getCurrencyCode() with https://msdn.microsoft.com/en-us/library/windows/apps/windows.globalization.geographicregion

Basically split the local string input by "-" and take the two letters after it, and construct a GeographicRegion from it, then return the value of the GeographicRegion.CurrenciesInUse property's first member?

Getting the currency symbol may be a little more difficult.  Maybe we can hack https://msdn.microsoft.com/en-us/library/windows/apps/windows.globalization.numberformatting.currencyformatter.aspx and have it format something like 0.00 and drop all spaces, digits and periods or commas and just assume what's left is the symbol?

Source/Ti/src/Locale.cpp
infosia: You mean,

```c++
currentLanguage__ = currentLocale__; // e.g. "en"
currentLocale__   = currentLocale__ + "-" + currentLanguage__; // e.g. "en-US"
```

right?

infosia:
Added rest of `Ti.Locale` functions except `formatTelephoneNumber`. [TIMOB-20480](https://jira.appcelerator.org/browse/TIMOB-20480)

- Ti.Locale.getCurrencyCode
- Ti.Locale.getCurrencySymbol
- Ti.Locale.getLocaleCurrencySymbol

infosia:
Seems like some network error is happening on the unit tests....keep trying to run them again.

infosia:
build win 10

infosia:
build win 8.1

Source/Ti/src/Locale.cpp
sgtcoolguy: yes :+1:

PR #575, opened 2016-03-07
Create test case based on stackoverflow issue

sgtcoolguy:
http://stackoverflow.com/questions/35843982/appcelerator-windows-phone-xml-parsing-error#new-answer

### DIFF
--- a/Examples/NMocha/src/Assets/ti.xml.test.js
+++ b/Examples/NMocha/src/Assets/ti.xml.test.js
@@ -1068,4 +1068,25 @@ describe("Titanium.XML", function () {
         should(thirdNewAttr.specified).be.true;
         finish();
     });
+
+	it("evaluate", function(finish) {
+		var xml = Ti.XML.parseString("<returnXml>\n" +
+"    <methodStatus>\n" +
+"        <statusType>success</statusType>\n" +
+"    </methodStatus>\n" +
+"    <parametersReturn />\n" +
+"</returnXml>");
+		var statusType = xml.evaluate("/returnXml/methodStatus/statusType");
+		should(statusType).be.an.Object;
+		should(statusType.item(0).text).eql("success"); // Isn't it textContent?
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
@infosia Can you take a look into why this would be failing? This came from a stack overflow issue a user posted where the code works on Android but not Windows Phone.

infosia:
According to the [API document](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.XML), `Ti.XML.parseString` returns `Ti.XML.Document` which doesn't have `evaluate` function. So `TypeError: xml.evaluate is not a function` seems to be a expected behavior. `xml.getElementsByTagName` should be working even in this case though. I'll take a look.

infosia:
Confirmed that `getElementsByTagName` works. `evaluate` is not implemented on Windows since it implements DOM Level 2. [Ti.XML API document](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.XML) also states it supports DOM Level 2. Do we officially support Level 3 API?

```javascript
var xml = Ti.XML.parseString("<returnXml>\n" +
"    <methodStatus>\n" +
"        <statusType>success</statusType>\n" +
"    </methodStatus>\n" +
"    <parametersReturn />\n" +
"</returnXml>");
var statusTypeList = xml.getElementsByTagName("statusType");
Ti.API.info(JSON.stringify(statusTypeList.item(0).textContent,null,2));
```

sgtcoolguy:
@infosia The docs may be out of date and maybe we _do_ support DOM Level 3 on Android (and maybe even iOS)? cc @hieupham007 @pec1985 @jawa9000 (Can any of you confirm/deny?)

If that's the case it would be good to open a JIRA for Windows Phone to support level 3 as well (in the 5.4 or later timeframe).

infosia:
Created [TIMOB-20549](https://jira.appcelerator.org/browse/TIMOB-20549) for tracking this.

sgtcoolguy:
Closing this PR for now

PR #576, opened 2016-03-08
Make sure to clear the cache on suspend

infosia:
[TIMOB-20316](https://jira.appcelerator.org/browse/TIMOB-20316)

https://github.com/appcelerator/HAL/pull/61 introduced new way to clear the constant cache: `JSExport<T>::EvictCache()`. We might want to delete the cache on app suspension to prevent app from using memory when it's not active, and this PR is for it.

Note that this PR needs https://github.com/appcelerator/HAL/pull/61 to be merged. This PR is also meant to showcase how to use `JSExport<T>::EvictCache()`...as you can see you need to call `JSExport` with template parameter `<T>`, which make it a bit redundant to code. Feel free to comment if you have better ideas for this.

### DIFF
--- a/Source/Titanium/src/TitaniumWindows.cpp
+++ b/Source/Titanium/src/TitaniumWindows.cpp
@@ -358,6 +358,66 @@ namespace TitaniumWindows
 			}
 		}
 
+		// hmm, is there any better way to do clear the cache?
+		// we might want to provide more unified way for this
+		JSExport<TitaniumWindows::TiModule>::EvictAllCache();
+		JSExport<TitaniumWindows::GlobalString>::EvictAllCache();
+		JSExport<TitaniumWindows::UIModule>::EvictAllCache();
+		JSExport<TitaniumWindows::API>::EvictAllCache();
+		JSExport<TitaniumWindows::AppModule>::EvictAllCache();
+		JSExport<TitaniumWindows::Codec>::EvictAllCache();
+		JSExport<TitaniumWindows::Locale>::EvictAllCache();
+		JSExport<TitaniumWindows::Platform>::EvictAllCache();
+		JSExport<TitaniumWindows::Gesture>::EvictAllCache();
+		JSExport<TitaniumWindows::Geolocation>::EvictAllCache();
+		JSExport<TitaniumWindows::Accelerometer>::EvictAllCache();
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
test this please

infosia:
build win 8.1

sgtcoolguy:
test this please

infosia:
test this please

sgtcoolguy:
The Windows 10 machine kept crashing during the test: downloadLargeFile. Looks like our HTTPClient tests are a bit unstable right now.

sgtcoolguy:
@infosia I'm a bit concerned that the commit here: https://github.com/appcelerator/titanium_mobile_windows/commit/e81224f523b377c7f2d4b24a61db6368bc21a04c may have destabilized the HTTPClient tests. 

I think the new timeout handling code is unstable, and may throw an uncaught exception sometimes?

I have no idea, so here's some of my thoughts:
- Maybe we need to explicitly create a task and check when it's cancelled in a loop using smaller wait times until the timeout is done)? 
- Or maybe it's because we don't "recreate" the cancellation token source after cancelling in the async timeout task like we used to?
https://msdn.microsoft.com/en-us/library/dd984117.aspx

sgtcoolguy:
test this please

sgtcoolguy:
I opened #587 to try and fix the issue. The tests passed immediately, so maybe that fixed it (recreating the cancellationTokenSource__ on timeout).

sgtcoolguy:
build win 10

garymathews:
build win 10

infosia:
test this please

infosia:
> Jenkins Windows 10 SDK PR Build failed
> Titanium.Network.decodeURIComponent
> expected undefined to be a function

I don't know what's happening there...

test this please

infosia:
test this please

infosia:
Ah I see what's going on there.  `Ti.Network.decodeURIComponent` is lost because [TiModule](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/src/TiModule.cpp#L514) setup those on application startup, and lost them when you clear `Ti` object cache. It should have been re-built when you re-create it. 

I think I need to introduce more easier way to do this...close this PR for now.

PR #572, opened 2016-02-29
[TIMOB-20481] Support l10n application name

infosia:
[TIMOB-20481](https://jira.appcelerator.org/browse/TIMOB-20481)

Add localization support for application name. You can use Titanium's [string localization resources](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.Locale-method-getString) to localize application name. You need to use `app_name` for application name, and `app_description` for application description. These entries are optional. Note that you may need to uninstall existing app in order to see updated application name.

## i18n/en/strings.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<resources>
    <string name="app_name">Titanium Test App</string>
    <string name="app_description">This is awesome app</string>
</resources>
```

## i18n/ja/strings.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<resources>
    <string name="app_name">Titanium テスト</string>
</resources>
```

PR #579, opened 2016-03-10
Fix: default app_name/description should be added only when there's no locale strings

infosia:
Related to #572

Fix: Duplicate `app_name` entry build error happens when you build app with i18n application name. Default `app_name/description` should be added only when there's no locale strings.

## i18n/en/strings.xml
```xml
<resources>
    <string name="app_name">Titanium Test App</string>
    <string name="app_description">This is awesome app</string>
</resources>
```

### DIFF
--- a/cli/commands/_build/generate.js
+++ b/cli/commands/_build/generate.js
@@ -67,13 +67,15 @@ function generateI18N(next) {
 		// process default app_name
 		if (data[locale].strings && !data[locale].strings.app_name) {
 			data[locale].strings.app_name = data.en.app.appname;
-		} else {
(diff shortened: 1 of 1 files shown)


cli/commands/_build/generate.js
```diff
@@ -67,13 +67,15 @@ function generateI18N(next) {
 		// process default app_name
 		if (data[locale].strings && !data[locale].strings.app_name) {
 			data[locale].strings.app_name = data.en.app.appname;
-		} else {
```
infosia: Note: This `else {` condition includes `data[locale].strings && data[locale].strings.app_name`, which causes duplicate app_name...

PR #586, opened 2016-03-22
[TIMOB-20600] Fix: Window.close throws "unknown exception"

infosia:
Fix for [TIMOB-20600](https://jira.appcelerator.org/browse/TIMOB-20600)

```javascript
var win1 = Ti.UI.createWindow({ backgroundColor: 'yellow' }),
    win2 = Ti.UI.createWindow({ backgroundColor: 'blue' }),
    win3 = Ti.UI.createWindow({ backgroundColor: 'gray' });

win1.addEventListener('open', function (e) {
    win2.open();
    setTimeout(function () {
        win3.open();
        setTimeout(function () {
            win2.close();
            setTimeout(function () {
                win3.close();
                win1.backgroundColor = 'green';
                // and then pushing "back" hardware button should exit the app
            }, 1000);
        }, 1000);
    }, 1000);
});

win1.open();
```

### DIFF
--- a/Source/UI/src/Window.cpp
+++ b/Source/UI/src/Window.cpp
@@ -134,33 +134,55 @@ namespace TitaniumWindows
 			// disable all events further because it doesn't make sense.
 			disableEvents();
 
-			auto rootFrame = dynamic_cast<Windows::UI::Xaml::Controls::Frame^>(Windows::UI::Xaml::Window::Current->Content);
+			const auto rootFrame = dynamic_cast<Windows::UI::Xaml::Controls::Frame^>(Windows::UI::Xaml::Window::Current->Content);
+			const auto top_window = window_stack__.back();
+			const auto is_top_window = (top_window.get() == this); // check if window.close has been issued onto the top window
+
 			if (!get_exitOnClose() && window_stack__.size() > 1) {
-				rootFrame->GoBack();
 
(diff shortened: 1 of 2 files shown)


Source/UI/src/Window.cpp
```diff
@@ -134,33 +134,55 @@ namespace TitaniumWindows
...
-				rootFrame->GoBack();
 
-				auto top_window = window_stack__.back();
+				// Check if stack-top is not a current Window.
+				// This usually means new windows is opened before current windows is closed,
+				// or user explicitly closes a window which is not on top.
+				// In this case we don't want to GoBack().
+				if (!is_top_window) {
+
+					// Iterate over the stack to search for shared_ptr for "this" because std::find does not work in this case.
```
garymathews: I think we could use ``std::find_if``?

```C++
const auto i = std::find_if(window_stack__.begin(), window_stack__.end(), [=](auto w){
	return w.get() == this;
});
if (i != window_stack__.end()) {
	window_stack__.erase(i);
} else {
	TITANIUM_LOG_WARN("Window.close: Window is not opened or already closed");
}
```

infosia: woot cool, I didn't know this. I'll use this when I get a chance next.

PR #587, opened 2016-03-22
Attempt to fix wonky HTTPClient

sgtcoolguy:
Testing, testing 1...2...3

### DIFF
--- a/Source/Network/src/HTTPClient.cpp
+++ b/Source/Network/src/HTTPClient.cpp
@@ -153,7 +153,7 @@ namespace TitaniumWindows
 		void HTTPClient::send(Windows::Web::Http::IHttpContent^ content)
 		{
 			auto uri = ref new Windows::Foundation::Uri(TitaniumWindows::Utility::ConvertString(location__));
-			
+
 			// Set up the request
 			Windows::Web::Http::HttpRequestMessage^ request;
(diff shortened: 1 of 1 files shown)


Source/Network/src/HTTPClient.cpp
```diff
@@ -339,11 +339,10 @@ namespace TitaniumWindows
...
 			if (timeout__.count() > 0) {
-				const auto timeout = timeout__.count();
-				const auto token_source = cancellationTokenSource__;
-				create_async([timeout, token_source](concurrency::cancellation_token token) {
-					concurrency::wait(static_cast<unsigned int>(timeout));
-					token_source.cancel();
+				create_async([this](concurrency::cancellation_token token) {
+					concurrency::wait(static_cast<unsigned int>(timeout__.count()));
+					cancellationTokenSource__.cancel();
+					cancellationTokenSource__ = concurrency::cancellation_token_source();
```
infosia: Hmm. I was thinking that using member variable (`timeout__` etc) could cause access violation here because certain period before this `timeout` `HTTPClient` instance may have been garbage collected. That's why I didn't use member variables in `create_async` and used copy of `token_source`. I was not pretty sure how it works though...I'll look into more on this.

PR #593, opened 2016-03-24
[TIMOB-20609] ImageView.image from Windows-style path

infosia:
Fix for [TIMOB-20609](https://jira.appcelerator.org/browse/TIMOB-20609)

- ImageView.image from WIndows-style path such as `C:\Data\Users\DefApps\APPDATA\Local\Packages\etc\etc`

- ~~ImageView.image from `Ti.Blob`~~
- ~~ImageView.image from `Ti.Filesystem.File`~~

### DIFF
--- a/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
+++ b/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
@@ -47,6 +47,30 @@ describe("Titanium.UI.ImageView", function () {
         imageView.image = Ti.Filesystem.resourcesDirectory + 'Logo.png';
     });
 
+    (isWindows() ? it : it.skip)("image (local path with separator)", function (finish) {
+        var imageView = Ti.UI.createImageView();
+        imageView.addEventListener('load', function() {
+            should(imageView.image).be.a.String;
+            should(imageView.image).eql(Ti.Filesystem.resourcesDirectory + Ti.Filesystem.separator + 'Logo.png');
+            finish();
+        });
+        // Try appending separator
+        // It's not quite clear if we needs separator, but people often do this
+        imageView.image = Ti.Filesystem.resourcesDirectory + Ti.Filesystem.separator + 'Logo.png';
+    });
+
+    (isWindows() ? it : it.skip)("image (local path with /)", function (finish) {
+        var imageView = Ti.UI.createImageView();
+        imageView.addEventListener('load', function() {
+            should(imageView.image).be.a.String;
+            should(imageView.image).eql(Ti.Filesystem.resourcesDirectory + '/' + 'Logo.png');
+            finish();
+        });
+        // Try appending '/' for the separator
+        // Technically this is not right on Windows, but people often do this
+        imageView.image = Ti.Filesystem.resourcesDirectory + '/' + 'Logo.png';
+    });
+
     (isWindows() ? it : it.skip)("image (nativePath)", function (finish) {
         var fromFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "Logo.png");
         var imageView = Ti.UI.createImageView();
@@ -83,32 +107,6 @@ describe("Titanium.UI.ImageView", function () {
         });
 
         imageView.image = 'ms-appdata:///local/TIMOB-20609.png';
-    });
-
-    (isWindows() ? it : it.skip)("image (File)", function (finish) {
-        var fromFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "Logo.png");
-
-        var imageView = Ti.UI.createImageView();
-        imageView.addEventListener('load', function() {
-            should(imageView.image).be.an.Object;
-            should(imageView.image).eql(fromFile);
-            finish();
-        });
-
-        imageView.image = fromFile;
-    });
-
-    (isWindows() ? it : it.skip)("image (Blob)", function (finish) {
-        var fromFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "Logo.png"),
-            blob = fromFile.read();
-        var imageView = Ti.UI.createImageView();
-        imageView.addEventListener('load', function() {
-            should(imageView.image).be.an.Object;
-            should(imageView.toBlob()).eql(blob);
-            finish();
-        });
-
-        imageView.image = blob;
     });
 
     // TIMOB-18684
--- a/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
+++ b/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
@@ -9,28 +9,106 @@ var should = require('./should');
 describe("Titanium.UI.ImageView", function () {
     this.timeout(5000);
 
+    function isWindows() {
+        return (Ti.Platform.osname == 'windowsphone' || Ti.Platform.osname == 'windowsstore');
+    }
+
     it("apiName", function (finish) {
-        var label = Ti.UI.createImageView({
+        var imageView = Ti.UI.createImageView({
             image: "https://www.google.com/images/srpr/logo11w.png"
         });
-        should(label.apiName).be.a.String;
-        should(label.apiName).be.eql("Titanium.UI.ImageView");
+        should(imageView.apiName).be.a.String;
+        should(imageView.apiName).be.eql("Titanium.UI.ImageView");
         finish();
     });
 
 
-    it("image", function (finish) {
-        var label = Ti.UI.createImageView({
+    it("image (URL)", function (finish) {
+        var imageView = Ti.UI.createImageView({
             image: "https://www.google.com/images/srpr/logo11w.png"
         });
-        should(label.image).be.a.String;
-        should(label.getImage).be.a.Function;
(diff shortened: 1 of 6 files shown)


garymathews:
build win 10

infosia:
build win 10

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
```diff
@@ -9,30 +9,108 @@ var should = require('./should');
...
+        should(imageView.getImage).be.a.Function;
+        should(imageView.image).eql('https://www.google.com/images/srpr/logo11w.png');
+        should(imageView.getImage()).eql('https://www.google.com/images/srpr/logo11w.png');
+        imageView.image = 'path/to/logo.png';
+        should(imageView.image).eql('path/to/logo.png');
+        should(imageView.getImage()).eql('path/to/logo.png');
         finish();
     });
 
+    (isWindows() ? it : it.skip)("image (local path)", function (finish) {
```
sgtcoolguy: any reason this wouldn't work on other platforms?

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
```diff
@@ -9,30 +9,108 @@ var should = require('./should');
...
+        var imageView = Ti.UI.createImageView();
+        imageView.addEventListener('load', function() {
+            should(imageView.image).be.a.String;
+            should(imageView.image).eql(Ti.Filesystem.resourcesDirectory + 'Logo.png');
+            finish();
+        });
+        imageView.image = Ti.Filesystem.resourcesDirectory + 'Logo.png';
+    });
+
+    (isWindows() ? it : it.skip)("image (nativePath)", function (finish) {
```
sgtcoolguy: any reason this wouldn't work on other platforms?

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
```diff
@@ -9,30 +9,108 @@ var should = require('./should');
...
+            should(imageView.image).be.a.String;
+            should(imageView.image).eql('ms-appdata:///local/TIMOB-20609.png');
+            toFile.deleteFile();
+            finish();
+        });
+
+        imageView.image = 'ms-appdata:///local/TIMOB-20609.png';
+    });
+
+    (isWindows() ? it : it.skip)("image (File)", function (finish) {
```
sgtcoolguy: any reason this wouldn't work on other platforms?

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
```diff
@@ -9,30 +9,108 @@ var should = require('./should');
...
+        imageView.addEventListener('load', function() {
+            should(imageView.image).be.an.Object;
+            should(imageView.image).eql(fromFile);
+            finish();
+        });
+
+        imageView.image = fromFile;
+    });
+
+    (isWindows() ? it : it.skip)("image (Blob)", function (finish) {
```
sgtcoolguy: any reason this wouldn't work on other platforms?

Source/TitaniumKit/src/UI/ImageView.cpp
```diff
@@ -256,21 +266,50 @@ namespace Titanium
 
 		TITANIUM_PROPERTY_GETTER(ImageView, image)
 		{
-			return get_context().CreateString(get_image());
+			if (!get_image().empty()) {
```
sgtcoolguy: pull out variables to hold values so we don't call the same method twice (in case it's expensive?)
``` cpp
const auto context = get_context();
const auto image = get_image();
if (!image.empty()) {
  return context.CreateString(image);
}
const auto imageBlob = get_imageAsBlob();
if (imageBlob) {
	return imageBlob->get_object();	
}
const auto imageFile = get_imageAsFile();
if (imageFile) {
	return imageFile->get_object();
}
return context.CreateNull();
```

Source/TitaniumKit/src/UI/ImageView.cpp
```diff
@@ -256,21 +266,50 @@ namespace Titanium
...
 		}
 
 		TITANIUM_PROPERTY_SETTER(ImageView, image)
 		{
-			TITANIUM_ASSERT(argument.IsString());
-			set_image(static_cast<std::string>(argument));
+			if (argument.IsObject()) {
+				auto obj = static_cast<JSObject>(argument);
+				auto blob = obj.GetPrivate<Titanium::Blob>();
+				auto file = obj.GetPrivate<Titanium::Filesystem::File>();
```
sgtcoolguy: move down file decl so we don't execute the conversion attempt unless we know blob failed?
``` cpp
auto obj = static_cast<JSObject>(argument);
auto blob = obj.GetPrivate<Titanium::Blob>();

if (blob) {
	set_imageAsBlob(blob);
	return true;
}
auto file = obj.GetPrivate<Titanium::Filesystem::File>();
if (file) {
 	set_imageAsFile(file);
	return true;
 }
TITANIUM_LOG_WARN("ImageView.image only accepts String, Blob or File");

```

Source/TitaniumKit/src/UI/ImageView.cpp
```diff
@@ -256,21 +266,50 @@ namespace Titanium
...
+			}
 			return true;
 		}
 
 		TITANIUM_PROPERTY_GETTER(ImageView, images)
 		{
 			std::vector<JSValue> images;
-			for (auto image : get_images()) {
-				images.push_back(get_context().CreateString(image));
+			if (get_imagesAsBlob().size() > 0) {
```
sgtcoolguy: same as above, assign to variables in case the method call is expensive

Source/UI/src/ImageView.cpp
```diff
@@ -287,11 +328,31 @@ namespace TitaniumWindows
...
+				set_imageAsFile(images.at(0));
+			}
+		}
+
 		void ImageView::set_images(const std::vector<std::string>& images) TITANIUM_NOEXCEPT
 		{
 			Titanium::UI::ImageView::set_images(images);
-			// HACK For now let's cheat and just set the first image as a static one
-			set_image(images.at(0));
+			if (images.size() > 0) {
```
sgtcoolguy: This whole if can be removed...

Source/UI/src/ImageView.cpp
```diff
@@ -287,11 +328,31 @@ namespace TitaniumWindows
...
+				// HACK For now let's cheat and just set the first image as a static one
+				set_imageAsBlob(images.at(0));
+			}
+		}
+
+		void ImageView::set_imagesAsFile(const std::vector<std::shared_ptr<Titanium::Filesystem::File>>& images) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ImageView::set_imagesAsFile(images);
+			if (images.size() > 0) {
```
sgtcoolguy: This whole if can be removed...

Source/UI/src/ImageView.cpp
```diff
@@ -287,11 +328,31 @@ namespace TitaniumWindows
 			});
 		}
 
+		void ImageView::set_imagesAsBlob(const std::vector<std::shared_ptr<Titanium::Blob>>& images) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ImageView::set_imagesAsBlob(images);
+			if (images.size() > 0) {
```
sgtcoolguy: This whole if can be removed. I added the "hack" before we supported ImageView animations using the images property. Technically we are not supposed to show any image yet until start() is called when using the images property.

sgtcoolguy:
ImageView.start() needs to be updated to support the possible different images as well. Right now it assumes we have std::vector<std::string> images_; It needs to support trying blobs, files and then strings like the other updates you've done here.

sgtcoolguy:
We may want to consider breaking this PR up into the bug fix for the file path issue from the original ticket (for 5.3.0); and the addition of support for File/Blob in a new JIRA ticket for 5.4.0.

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
infosia: Yeah, that's because not all platform has `Ti.Filesystem.resourcesDirectory + 'Logo.png`. I expect only Windows places `Logo.png` under the resources directory.

infosia:
Picked up bug fix for file path with Windows native path.

```javascript
var fromFile = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "Logo.png");
var imageview = Ti.UI.createImageView({width:Ti.UI.FILL, height:Ti.UI.FILL});

imageview.image = fromFile.nativePath;
```

This fix also tries to "normalize" path separator that is not often treated right, such as:

```javascript
// this should just work
imageview.image = Ti.Filesystem.resourcesDirectory + "Logo.png";

// This should work too, right?
imageview.image = Ti.Filesystem.resourcesDirectory + Ti.Filesystem.separator + "Logo.png";

// This is not right on Windows, but people usually do this.
imageview.image = Ti.Filesystem.resourcesDirectory + '/' + "Logo.png";
```

infosia:
Ticket for `Blob` and `File` support for ImageView.image: [TIMOB-23115](https://jira.appcelerator.org/browse/TIMOB-23115)

infosia:
build win 10

PR #597, opened 2016-03-29
[TIMOB-18531] Use default publisher-guid when none is given on build

infosia:
Fix for [TIMOB-18531](https://jira.appcelerator.org/browse/TIMOB-18531)

Ti CLI should prompt to enter a guid when building for `dist-phonestore` or `dist-winstore` if `windows.publisherId` is not configured.

```
ti config -r windows.publisherId
appc run -p windows -T dist-phonestore
```

This also forces `production` build on every `dist-*` build even when `deploy-type` is given explicitly.

### DIFF
--- a/cli/commands/_build/config/winPublisherId.js
+++ b/cli/commands/_build/config/winPublisherId.js
@@ -18,9 +18,16 @@ module.exports = function configOptionWindowsPublisherID(order) {
 		callback(null, value);
 	}
 
+	function shouldPrompt(argv) {
+		// You need to check abbreviated form too here because they are not expanded yet
+		return (argv['T'] === 'dist-phonestore' || argv['target']      === 'dist-phonestore') ||
+		       (argv['T'] === 'dist-winstore'   || argv['target']      === 'dist-winstore') ||
+		       (argv['D'] === 'production'      || argv['deploy-type'] === 'production');
+	}
+
 	return {
 		abbr: 'I',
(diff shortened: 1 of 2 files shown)


cli/commands/_build/config/winPublisherId.js
```diff
@@ -18,9 +18,16 @@ module.exports = function configOptionWindowsPublisherID(order) {
...
+		// You need to check abbreviated form too here because they are not expanded yet
+		return (argv['T'] === 'dist-phonestore' || argv['target']      === 'dist-phonestore') ||
+		       (argv['T'] === 'dist-winstore'   || argv['target']      === 'dist-winstore') ||
+		       (argv['D'] === 'production'      || argv['deploy-type'] === 'production');
+	}
+
 	return {
 		abbr: 'I',
-		default: this.cli.argv['deploy-type'] !== 'production' ? '00000000-0000-1000-8000-000000000000' : this.config.get('windows.publisherId'),
+		default: !shouldPrompt(this.cli.argv) ? '00000000-0000-1000-8000-000000000000' : this.config.get('windows.publisherId'),
```
infosia: Note: We should check arguments here even when we force `production` build in [ cli/commands/_build/config.js ](https://github.com/appcelerator/titanium_mobile_windows/pull/597/files#diff-cf3672fd6786f665356e2032baf5ba90R88) because this line is executed before `cli:pre-validate`.

sgtcoolguy:
LGTM. Merging manually.

PR #599, opened 2016-03-30
[TIMOB-20598] views not animating correctly

infosia:
Fix for [TIMOB-20598](https://jira.appcelerator.org/browse/TIMOB-20598)

```javascript
var win = Ti.UI.createWindow({backgroundColor:'green'});
var box = Ti.UI.createView({
    backgroundColor: 'red',
    left: 100,
    height: '100',
    width: '100'
});

var pin1 = Ti.UI.createView({
    backgroundColor: 'blue',
    left: 150,
    height: '140',
    width: '6'
});
var pin2 = Ti.UI.createView({
    backgroundColor: 'blue',
    left: 200,
    height: '140',
    width: '6'
});
var pin3 = Ti.UI.createView({
    backgroundColor: 'blue',
    left: 250,
    height: '140',
    width: '6'
});
win.add(pin1);
win.add(pin2);
win.add(pin3);

win.add(box);
var pos = 100;
box.addEventListener('click', function () {
    pos += 50;
    var a = Ti.UI.createAnimation({
        left: pos,
        duration: 2000,
    });
    box.animate(a);
});
win.open();
```

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -529,11 +529,11 @@ namespace TitaniumWindows
 				// 
 				const auto top = animation->get_top();
 				if (top) {
-					set_top(std::to_string(*top));
+					setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, std::to_string(*top));
 				}
 				const auto left = animation->get_left();
 				if (left) {
-					set_left(std::to_string(*left));
+					setLayoutProperty(Titanium::LayoutEngine::ValueName::Left, std::to_string(*left));
 				}
 
 				// Make sure to clear the StoryBoard because transform made by StoryBoard remains.
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -520,7 +520,25 @@ namespace TitaniumWindows
 				storyboard->Children->Append(zIndex_anim);
 			}
 
-			storyboard->Completed += ref new Windows::Foundation::EventHandler<Platform::Object ^>([callback, this_object, animation](Platform::Object^ sender, Platform::Object ^ e) mutable {
+			storyboard->Completed += ref new Windows::Foundation::EventHandler<Platform::Object ^>([callback, animation, this_object, this](Platform::Object^ sender, Platform::Object ^ e) mutable {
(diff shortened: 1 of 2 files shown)


Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -520,7 +520,25 @@ namespace TitaniumWindows
...
 
-			storyboard->Completed += ref new Windows::Foundation::EventHandler<Platform::Object ^>([callback, this_object, animation](Platform::Object^ sender, Platform::Object ^ e) mutable {
+			storyboard->Completed += ref new Windows::Foundation::EventHandler<Platform::Object ^>([callback, animation, this_object, this](Platform::Object^ sender, Platform::Object ^ e) mutable {
+				const auto storyboard = dynamic_cast<Media::Animation::Storyboard^>(sender);
+				const auto component = getComponent();
+
+				//
+				// Make sure to update the position of the view because StoryBoard doesn't change actual position of it.
+				// 
+				const auto top = animation->get_top();
```
sgtcoolguy: So I think this actually goes against the stated docs where we _don't_ update the top/left/right/bottom values from what an animation does. Is this necessary for the fix, or does the Storeyboard->Stop() call alone fix the issue?

infosia: Huh I see. Interesting thing I found here is that `StoryBoard` doesn't update `UIElement` position too, which means `Canvas::GetLeft(view)` always returns original position. So what we should do here is to update the `UIElement` poistion and then clear the `StoryBoard`...

```c++
Canvas::SetLeft(component, *left);
Canvas::SetTop(component, *top);
storyboard->Stop();
```

I'll update again later today.

garymathews: http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.View-method-animate

> Note that if you use animate to move a view, the view's actual position is changed, but its layout properties, such as top, left, center and so on are not changed--these reflect the original values set by the user, not the actual position of the view.

infosia: Ah that's interesting idea. I'll  give it a try :+1:

sgtcoolguy:
CR looks good, @garymathews can you please do FR? if it's good you'll need to merge here and cherry-pick to 5_3_X

infosia:
build win 10

garymathews:
FR: **PASS**

PR #605, opened 2016-04-04
Use return value from Collection.IndexOf to check if it's found

infosia:
Fix logical error.

The `boolean` return value from `UIElementCollection.IndexOf` should be used to check if element has been found. `index` parameter for `IndexOf` never be negative because it's unsigned int.


infosia:
build win 10

sgtcoolguy:
@infosia Should this be back ported to 5_3_X?

PR #610, opened 2016-04-07
[TIMOB-23158] Use of native module for UWP 10.0 fails

infosia:
[TIMOB-23158](https://jira.appcelerator.org/browse/TIMOB-23158)

Fix module build error on Windows 10. When you build module on Windows 10, it builds module DLL/winmd for `phone`, `store` and `win10`.  When you build module on Windows 8.1, it builds DLL/winmd for `phone` and `store`. Note that you can't build Windows 10 module on Windows 8.1 because Windows compiler/cmake/sdk does not allow it.

```bash
ti create -p windows -t module
cd MY_MODULE_NAME/windows
# On Windows 10
ti build -p windows -T ws-local -l trace --wp-sdk 10.0
# On Windows 8.1
ti build -p windows -T ws-local -l trace --wp-sdk 8.1
```


infosia:
build win 10

garymathews:
```
appc new -t timodule -n winModule --id com.win.module
cd winModule\windows
appc run -p windows -T ws-local --wp-sdk 10.0 -l trace
```
```
ERROR  | An uncaught exception was thrown!
ENOENT, no such file or directory 'C:\Users\gmathews\projects\winModule\windows\tiapp.xml'
```

It should be looking for ``timodule.xml``, although I think this is a separate issue related to the CLI.

garymathews:
Testing on Windows 10 it seems to skip the ``Windows 10 Win32`` project

```
[INFO]  Skipping Windows 10 Win32 project
```

[windows-module.js#L30](https://github.com/appcelerator/titanium_mobile_windows/pull/610/files#diff-3428e0f5d6228b6ef7975265521bbf91R30) ``os.release()`` returns ``6.3.9600`` and not ``10.0`` like expected

sgtcoolguy:
6.3.9600 corresponds to Windows 8.1...

infosia:
> [windows-module.js#L30](https://github.com/appcelerator/titanium_mobile_windows/pull/610/files#diff-3428e0f5d6228b6ef7975265521bbf91R30) ``os.release()`` returns ``6.3.9600`` and not ``10.0`` like expected

hmm...Do you have any idea to ensure detecting Windows 10 on Node.js?

garymathews:
Turns out I was using an outdated version of node, it works fine on anything ``>0.12.7``

infosia:
> It should be looking for ``timodule.xml``, although I this is a separate issue related to the CLI.

Yes, `appc ti` has issues with building native module: updated [TIMOB-23157](https://jira.appcelerator.org/browse/TIMOB-23157)

infosia:
> Turns out I was using an older version of node, it works fine on anything >0.12.7

Nice! :+1:

garymathews:
We do support ``0.12`` though, so this would be an issue to anyone with ``0.12.7`` or below

garymathews:
Appcelerator Studio downloads and installs node ``0.12.7`` by default on Windows.

infosia:
Sweet. Thank you for following up! So I think it's ok for now to ask upgrade minor version if your uses 0.12. I also saw that Appc portal points to Node.js 4.4.2 for Titanium CLI requirement.

templates/module/default/hooks/windows-module.js
```diff
@@ -69,32 +86,43 @@ exports.init = function (logger, config, cli) {
 
 function runCMake(logger, cmake, projectDir, targetEnv, targetArch, callback){
 	var targetDir = path.join(projectDir, targetEnv+'.'+targetArch),
-		generatorName = 'Visual Studio 12 2013'+(targetArch === 'ARM' ? ' ARM' : ''),
+		generatorName = 'Visual Studio 14 2015'+(targetArch === 'ARM' ? ' ARM' : ''),
```
sgtcoolguy: I don't think e can just hard-code 2015 here, right? Don't we need to support 8.1 only module builds with 2013 generator (if they specify that)?

infosia: Good point. I think we want to select VS version to use, right? Because we run cmake at `ti create -p windows -t module`, we need to add new option for `ti create` to specify VS version...such as `ti create -p windows -t module --wp-sdk 8.1`.

infosia: Or, we just force to use VS 2015 on Windows 10?

```javascript
if (isWindows10()) {
    generatorName = 'Visual Studio 14 2015'+(targetArch === 'ARM' ? ' ARM' : '');
}
```

infosia: I was hoping we can add new CLI option for `ti create` (such as `--wp-sdk 8.1`) but it turns out CLI does not parse additional options. I need to loop deeper into titanium CLI itself then...

infosia:
ok, pushed new option to support 8.1 ... `ti create -p windows -t module --wp-sdk 8.1` that uses VS 2013 generator .

infosia:
Follow-up: Team discussing to support 0.12.8. In case we can't upgrade I can remove 0.12.8 dependency if you want (by removing os.release() but check "--wp-sdk 10.0” CLI argument).

PR #619, opened 2016-04-13
[TIMOB-23205] Background service extension by native module

infosia:
Background service extension by native module: [TIMOB-23205](https://jira.appcelerator.org/browse/TIMOB-23205)

See [Windows Background Service Quick Start](https://wiki.appcelerator.org/display/guides2/Windows+Background+Service+Quick+Start) documentation on how to implement/test background task.

### DIFF
--- a/Examples/NMocha/src/Assets/ti.app.windows.backgroundservice.test.js
+++ b/Examples/NMocha/src/Assets/ti.app.windows.backgroundservice.test.js
@@ -16,27 +16,38 @@ describe('Ti.App.Windows.BackgroundService', function () {
         finish();
     });
 
-    it('locationServiceEnabled', function (finish) {
-        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.Boolean;
-        Ti.App.Windows.BackgroundService.locationServiceEnabled = true;
-        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.true;
-        Ti.App.Windows.BackgroundService.locationServiceEnabled = false;
-        should(Ti.App.Windows.BackgroundService.locationServiceEnabled).be.false;
-        finish();
-    });
-
     it('registerTimerTask', function (finish) {
-        var task = Ti.App.Windows.BackgroundService.registerTimerTask(function() {
-            Ti.API.info('registerTimerTask');
-        }, 15, true);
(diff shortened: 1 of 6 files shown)


infosia:
build win 10

Buder:
Just made a quick review on the file changes, not that I know widows code, but you removed some of the geolocation stuff, will this affect the solution in such way that I am not able to get the location in the callback natively? I guess you ar in top if this but I still have no time to loose here. Also how do I see when it is merged? I have marked this PR as watch so I will get notification when something happens... :)

sgtcoolguy:
Merged since I know @Buder is anxious to test out the latest...

Buder:
Ok, thanks I can see now it is in master, I guess todays nightly build will be the one I can use then...

Thanks

infosia:
We have removed location stuff because e have experienced JavaScript in background doesn't work well.

- Ti.App.Windows.BackgroundService.locationServiceEnabled is removed
- `callback` paraemter for Ti.App.Windows.BackgroundService.registerTimerTask is removed.

Now `registerTimerTask` needs *C++/CX class name* instead of JavaScript callback, let say we have ComExampleTest::BackgroundServiceTask C++/CX class:

```javascript
  task = Ti.App.Windows.BackgroundService.registerTimerTask('ComExampleTest.BackgroundServiceTask', 15, false);
```

For more API details please see [Windows Background Service Quick Start](https://wiki.appcelerator.org/display/guides2/Windows+Background+Service+Quick+Start).

Buder:
Ok, so I will create a function in my module and add it by the above statement. I worry about the location though, I have to be able to access geolocation both in the app (TI context) and in the background (Native context) does the above removed APIs affect this? In any case I will get the nightly build and try this out, my problem will be how to determine if it does in fact run, I have still no prints fro native code, that would be the best way to print something (location) every 15 min... but I could do with just checking if it runs or number of times it did run when I bring the app to front. Also does the backgrund service run during app in foreground? I will get back in I need more info, do you prefer here or in the ticket ?

Buder:
I have now read your guide and I have a few questions, can I use my current native module to add these class for background service OR do I have to create a "new" module based on the latest SDK from nightly build to do this? Also the entry point name is this a fix value or can I use any function?

PR #620, opened 2016-04-13
[TIMOB-20571] Use device or emulator wpsdk when possible

garymathews:
- Use device specified ``wpsdk`` when possible
- List all emulators when both ``-S`` or ``--wp-sdk`` are unspecified

###### TEST CASE
```
appc new -t app -n TIMOB-20571 --id com.win.test --no-services
cd TIMOB-20571
appc run -p windows -T wp-emulator
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-20571)

### DIFF
--- a/cli/commands/_build/config/deviceID.js
+++ b/cli/commands/_build/config/deviceID.js
@@ -43,6 +43,11 @@ module.exports = function configOptionDeviceID(order) {
 			return callback(new Error(__('Invalid device id "%s"', value)));
 		}
 
+		// use wpsdk specified for device
+		if (dev.wpsdk) {
+			cli.argv['wp-sdk'] = dev.wpsdk;
+		}
+
 		// check the device
 		if (cli.argv.target === 'wp-device') {
 			// try connecting to the device to detect no device or more than 1 device
--- a/cli/commands/_build/config/deviceID.js
+++ b/cli/commands/_build/config/deviceID.js
@@ -18,20 +18,22 @@ module.exports = function configOptionDeviceID(order) {
 			return callback(new Error(__('Invalid device id')));
 		}
 
-		var devices = this.getTargetDevices();
+		var devices = this.getTargetDevices(),
+			dev = null;
 
-		if (cli.argv.target === 'wp-emulator' && value === 'xd' && devices[0]) {
-			// pick first emulator
+		// xd: first emulator
+		// de: first device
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
Merged appcelerator/windowslib#37, published windowslib 0.4.7. Opened PRs to update titanium_mobile:
appcelerator/titanium_mobile#7944
appcelerator/titanium_mobile#7943

cli/commands/_build/config/deviceID.js
```diff
@@ -43,6 +43,11 @@ module.exports = function configOptionDeviceID(order) {
 			return callback(new Error(__('Invalid device id "%s"', value)));
 		}
 
+		// use wpsdk specified for device
+		if (dev.wpsdk) {
+			cli.argv['wp-sdk'] = dev.wpsdk;
```
sgtcoolguy: I don't think this handles every case. If the user specifies "wp-device" or "wp-emulator" above, we'll pick the first emulator/device, but the wp-sdk arg won't get set/defaulted to that device/emulator's wpsdk value. Only if the user chose by the name or did will this happen.

cli/commands/_build/config/deviceID.js
```diff
@@ -18,20 +18,22 @@ module.exports = function configOptionDeviceID(order) {
...
-			// pick first emulator
-			return callback(null, devices[0].udid);
-		}
-
-		if (cli.argv.target === 'wp-device' && value === 'de' && devices[0]) {
-			// pick first device
+		// xd: first emulator
+		// de: first device
+		if (cli.argv.target === 'wp-emulator' && value === 'xd' ||
+			cli.argv.target === 'wp-device' && value === 'de' && devices[0]) {
```
infosia: Is this right? I guess not, this could be `true` even when `devices[0]` is undefined.

```
$ node
> (true && true || true && false && false);
true
```

I would prefer grouping conditions in this case...

```
> ((true && true) || (false && false)) && false;
false
```

```diff
+		if (((cli.argv.target === 'wp-emulator' && value === 'xd') ||
+			(cli.argv.target === 'wp-device' && value === 'de')) && devices[0]) {
```

infosia: Added #621

sgtcoolguy:
Closed in favor of #621 with minor logic fix.

PR #622, opened 2016-04-14
[TIMOB-20572] Only list local devices

garymathews:
- Filter out any devices that are not connect by USB

###### TEST CASE
```
# connect phone to USB
# should see device
# disconnect phone from USB
# connect phone to network
# should not see device
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-20572)

### DIFF
--- a/cli/commands/_build/utils.js
+++ b/cli/commands/_build/utils.js
@@ -45,7 +45,15 @@ function getTargetDevices() {
 		}, this);
 		return this.deviceCache;
 	} else if (target === 'wp-device' && Array.isArray(this.windowsInfo.devices)) {
-		return this.deviceCache = this.windowsInfo.devices;
+		var devices = this.windowsInfo.devices;
+		for (var d in devices) {
+			var device = devices[d];
+			// only list local devices
+			if (device.ip && device.ip !== "127.0.0.1") {
+				devices.splice(d, 1);
+			}
+		}
+		return this.deviceCache = devices;
 	}
 
 	return [];


cli/commands/_build/utils.js
```diff
@@ -45,7 +45,15 @@ function getTargetDevices() {
 		}, this);
 		return this.deviceCache;
 	} else if (target === 'wp-device' && Array.isArray(this.windowsInfo.devices)) {
-		return this.deviceCache = this.windowsInfo.devices;
+		var devices = this.windowsInfo.devices;
+		for (var d in devices) {
+			var device = devices[d];
+			// only list local devices
+			if (device.ip && device.ip !== "127.0.0.1") {
```
garymathews: This only applies for Windows 10 devices; local  Windows 10 devices are picked up by ``wptool`` as ``Device`` anyway, should we remove all network devices since we can't determine their ``wpsdk``?

sgtcoolguy:
So have we looked deeper to see if there's no way to truly install apps to a windows 10 device on the same subnet? According to the docs we should be able to. Is it just that the phone needs certain settings or something? https://msdn.microsoft.com/en-us/library/mt203806.aspx

I don't really want to lock down network installs altogether if they are possible, because obviously thats more useful than always having to connect via USB. cc @eharrisappcel @eric34 @wluu

eharrisappcel:
I tried playing with the network deploy a few weeks back using  WinAppDeployCmd but was unsuccessful. I would run the command using what is listed the WinAppDeployCmd help listing but it would always reject the pin, and still is right now
```
Opening connection to device at '192.168.2.10'.

Access denied while connecting to the remote device.
Please retry the command with the "-pin" option and a valid pin as shown in the device settings.
The pin is a one time requirement to establish a pairing relationship with the device.
```
However when using the [device portal](https://ms-iot.github.io/content/en-US/win10/tools/DevicePortal.htm) the same pin is accepted. See also [here](http://embedded101.com/Blogs/David-Jones/entryid/680/Win-10-IoT-Universal-Windows-Apps-Sideloading) for some discussion about it, seems like he never go it to work (although the post is from September)

PR #631, opened 2016-04-18
[TIMOB-23222] Fix crash at Ti.Codec.decodeNumber

infosia:
Fix for [TIMOB-23222](https://jira.appcelerator.org/browse/TIMOB-23222)

```javascript
var win = Ti.UI.createWindow({backgroundColor:'blue'});
var file = Ti.Filesystem.getFile(Ti.Filesystem.resourcesDirectory, "data.bin");

Ti.API.info('File meta:' + file.size);
var blob = file.read();
var length = blob.length;

// Convert Ti.Blob to Ti.Buffer
var stream = Ti.Stream.createStream({ source: blob, mode: Ti.Stream.MODE_READ });
var buffer = Ti.createBuffer({ length: length });
stream.read(buffer);
Ti.API.info("file loaded. size=" + buffer.length);

// Read and parse the data
var index = Ti.Codec.decodeNumber({ source: buffer, position: 0, type: Ti.Codec.TYPE_INT });

    Ti.API.info('Checkpoint');
    var myArray = [];
    for (var i = 0; i < 2048; i++) {
        myArray.push([
            Ti.Codec.decodeNumber({ source: buffer, position: 4 + i * 8 + 0, type: Ti.Codec.TYPE_INT }),
            Ti.Codec.decodeNumber({ source: buffer, position: 4 + i * 8 + 4, type: Ti.Codec.TYPE_INT })
        ]);
    };
    win.backgroundColor = 'green';
    Ti.API.info('Done');
});

win.open();
```

### DIFF
--- a/Source/Ti/src/Codec.cpp
+++ b/Source/Ti/src/Codec.cpp
@@ -107,8 +107,7 @@ namespace TitaniumWindows
 	{
 		using namespace Windows::Storage::Streams;
 
-		const auto data = options.source->get_data();
-		auto source_data = std::vector<std::uint8_t>(data.begin() + options.position, data.end());
+		auto source_data = options.source->get_data(options.position, 8 /* we only need 8 bytes at most */);
 
 		Platform::ArrayReference<std::uint8_t> data_ref(&source_data[0], source_data.size());
 		const auto buffer = CryptographicBuffer::CreateFromByteArray(data_ref);
--- a/Source/TitaniumKit/include/Titanium/Buffer.hpp
+++ b/Source/TitaniumKit/include/Titanium/Buffer.hpp
@@ -72,6 +72,13 @@ namespace Titanium
 		TITANIUM_PROPERTY_IMPL_DEF(std::vector<std::uint8_t>, data);
 
 		/*!
+		@method
+		@abstract get_data
+		@discussion Return data partical representation of this buffer
+		*/
+		virtual std::vector<std::uint8_t> get_data(const std::uint32_t& offset, const std::uint32_t& size) TITANIUM_NOEXCEPT;
+
+		/*!
 		  @method
 		  @abstract append
 		  @discussion Appends `sourceBuffer` to the this buffer.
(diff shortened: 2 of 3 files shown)


infosia:
build win 10

infosia:
build win 10

Source/Ti/src/Codec.cpp
```diff
@@ -107,8 +107,7 @@ namespace TitaniumWindows
 	{
 		using namespace Windows::Storage::Streams;
 
-		const auto data = options.source->get_data();
-		auto source_data = std::vector<std::uint8_t>(data.begin() + options.position, data.end());
+		auto source_data = options.source->get_data(options.position, 8 /* we only need 8 bytes at most */);
```
sgtcoolguy: So I know this is the simplest fix: Just limit to 8 bytes since that's the max we'll need. Shouldn't we vary this based on the type like we do for Java/Android:
https://github.com/appcelerator/titanium_mobile/blob/415bd6c66dcc55b1a59a59574f3babd3c3a84ede/android/titanium/src/java/ti/modules/titanium/codec/CodecModule.java#L128

Basically, we need 1 - 8 bytes, depending on the type. if they're decoding bytes, then using 8 bytes hard-coded here is a big waste.
Looks like:
byte: 1
short: 2
int || float: 4
long || double: 8

sgtcoolguy: Also, should we take a look at encodeNumber to be sure w're not needlessly opening/reading huge buffers unnecessarily? It looks to me like we may be reading in the full destination buffer - maybe we can somehow avoid doing so and just inserting the new data at a particular position/verifying position is valid?

Buder:
Just want to comment that we use both the INT and BYTE length when we call the decodeNumber... not sure if we use the encodeNumber yet though... also supporting a system where we do not load everything into memory is very valuable as to me it seems the men issue is a great issue here, do not not why though as it is not data we load, but yet the system seems ot take up much memory to handle the decoding..

PR #632, opened 2016-04-18
[TIMOB-23222] Fix crash at Ti.Codec.decodeNumber - variable length of bytes to read

sgtcoolguy:
Additional commit on top of #631

### DIFF
--- a/Source/Ti/src/Codec.cpp
+++ b/Source/Ti/src/Codec.cpp
@@ -107,8 +107,7 @@ namespace TitaniumWindows
 	{
 		using namespace Windows::Storage::Streams;
 
-		const auto data = options.source->get_data();
-		auto source_data = std::vector<std::uint8_t>(data.begin() + options.position, data.end());
+		auto source_data = options.source->get_data(options.position, 8 /* we only need 8 bytes at most */);
 
 		Platform::ArrayReference<std::uint8_t> data_ref(&source_data[0], source_data.size());
 		const auto buffer = CryptographicBuffer::CreateFromByteArray(data_ref);
--- a/Source/Ti/src/Codec.cpp
+++ b/Source/Ti/src/Codec.cpp
@@ -107,15 +107,32 @@ namespace TitaniumWindows
 	{
 		using namespace Windows::Storage::Streams;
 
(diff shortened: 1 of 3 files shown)


garymathews:
[Codec.cpp#L65](https://github.com/appcelerator/titanium_mobile_windows/pull/632/files#diff-e631ecfd2736f145d679e5c31ed223d0R65)

I think this needs correcting?

```C++
case Type::Short:
	writer->WriteInt16(static_cast<std::int16_t>(options.source)); break;
case Type::Int:
	writer->WriteInt32(static_cast<std::int32_t>(options.source)); break;
```

Since that is how they are read [Codec.cpp#L141](https://github.com/appcelerator/titanium_mobile_windows/pull/632/files#diff-e631ecfd2736f145d679e5c31ed223d0R141)

Source/Ti/src/Codec.cpp
```diff
@@ -107,16 +107,32 @@ namespace TitaniumWindows
...
+		case Type::Short:
+			length = 2;
+			break;
+		case Type::Int:
+		case Type::Float:
+			length = 4;
+			break;
+		case Type::Double:
+		case Type::Long:
+			length = 8; // Do we always use 8 byte for long? Note that iOS-32bit uses 4 bytes for long.
```
garymathews: It looks like we aren't storing ``Type::Long`` as ``long`` but as a ``long long`` (``std::int64_t``) so it will always be ``8``

However, if we did use ``long`` it would always be ``4``, even on 64-bit architectures since MSVC++ uses ``LLP64``

infosia:
build win 10

infosia:
> [Codec.cpp#L65](https://github.com/appcelerator/titanium_mobile_windows/pull/632/files#diff-e631ecfd2736f145d679e5c31ed223d0R65)
> I think this needs correcting?
> Due to [Codec.cpp#L141](https://github.com/appcelerator/titanium_mobile_windows/pull/632/files#diff-e631ecfd2736f145d679e5c31ed223d0R141)

Hmm sounds like you are right. I will test/fix it after we merge this PR.

PR #634, opened 2016-04-19
Template: Module should extend Titanium::Module

infosia:
Update module template to support Ti event methods, to align with other platforms:

- Module should extend `Titanium::Module` to support event mechanics (`fireEvent`, `addEventListener` etc)


infosia:
build win 10

garymathews:
CR: **PASS**
FR: **PASS**

PR #640, opened 2016-04-20
[TIMOB-23251] Fix module build process for appc run

garymathews:
- Use correct build events for module build process

###### TEST CASE
```
appc new -t timodule -n module --id com.win.module
cd module\windows
appc run -p windows --build-only
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23251)

### DIFF
--- a/cli/commands/_buildModule.js
+++ b/cli/commands/_buildModule.js
@@ -51,7 +51,7 @@ WindowsModuleBuilder.prototype.run = function run(logger, config, cli, finished)
 
 	appc.async.series(this, [
 		function (next) {
-			cli.emit('build.pre.construct', this, next);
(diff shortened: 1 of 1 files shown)


infosia:
Nine catch! :+1: But can we be sure disabling `build.pre.compile` doesn't introduce issue in other area? /cc @sgtcoolguy

garymathews:
The only issue I can see of would be that any Javascript assets of the module would not be encrypted?

garymathews:
On further investigation, it looks like we were using the wrong events, I've updated the PR.

infosia:
build win 10

infosia:
build win 10

infosia:
build win 10

PR #646, opened 2016-04-25
[TIMOB-23253] Windows: Build errors when building to device and selecting second option in device prompt

sgtcoolguy:
Requires appcelerator/windowslib#41 and then we'll need to bump the version and update packaged windowslib on titanium_mobile 5_3_X branch.

- don't connect to device as part of deviceId validation. Seems good enough if the udid can be found in our listing. Connecting slows down the start of the build by a good amount.
- Don't report the build time twice
- Don't try to launch the apps on Windows 10 mobile as this always fails now. Instead spit out a warning message asking user to launch manually. This actually fixes https://jira.appcelerator.org/browse/TIMOB-20611
- log when we start installing dependencies and the app itself to give a little better indicator of progress

### DIFF
--- a/cli/hooks/wp-run.js
+++ b/cli/hooks/wp-run.js
@@ -161,9 +161,6 @@ exports.init = function (logger, config, cli) {
 		post: function (builder, finished) {
 			if (builder.buildOnly || !/^wp-emulator|wp-device$/.test(builder.target)) return finished();
 
-			var delta = appc.time.prettyDiff(cli.startTime, Date.now());
-			logger.info(__('Finished building the application in %s', delta.cyan));
-
 			function install() {
 				if (logRelay) {
 					// start the log relay server
@@ -213,15 +210,22 @@ exports.init = function (logger, config, cli) {
 							skipLaunch: true
 					}, opts),
 					appxExtensions = ['.appx', '.appxbundle'],
-					installs = [];
+					installs = [],
+					isWin10Device = (builder.target == 'wp-device' && builder.wpsdk == '10.0');
 
 				function installApp(deviceId, xapFile, opts, next) {
 					// Now install the real app
 					windowslib.install(deviceId, xapFile, appc.util.mix({
-								appGuid: builder.phoneProductId
+								appGuid: builder.phoneProductId,
+								// launching an app on Win 10 device fails right now!
+								skipLaunch: isWin10Device
 							}, opts))
 						.on('installed', function (handle) {
-							logger.info(__('Finished launching the application'));
+							if (isWin10Device) {
+								logger.warn(__('Due to a bug in Windows tooling, we are unable to launch apps for you on Windows 10 Mobile devices. Please launch the app manually.'));
+							} else {
+								logger.info(__('Finished launching the application'));
+							}
 
 							// watch for when the emulator is quit, if necessary
 							if (builder.target == 'wp-emulator') {
@@ -294,7 +298,7 @@ exports.init = function (logger, config, cli) {
 					});
 				});
 
-				logger.info(__('Installing and launching the application'));
+				logger.info(__('Installing and launching the application. Please wait as this may take some time...'));
 				async.series(installs, function (err, results) {
 					if (err) {
 						logger.error(err);
--- a/cli/hooks/wp-run.js
+++ b/cli/hooks/wp-run.js
@@ -215,6 +215,7 @@ exports.init = function (logger, config, cli) {
 
 				function installApp(deviceId, xapFile, opts, next) {
 					// Now install the real app
+					logger.info(__('Installing the application...'));
 					windowslib.install(deviceId, xapFile, appc.util.mix({
 								appGuid: builder.phoneProductId,
 								// launching an app on Win 10 device fails right now!
@@ -272,6 +273,7 @@ exports.init = function (logger, config, cli) {
 				});
 				possibleDependencies.forEach(function(file) {
 					installs.push(function (next) {
+						logger.info(__('Installing dependency: %s', file));
 						windowslib.install(builder.deviceId, path.resolve(dependenciesDir, file), installOnlyOpts)
 						.on('installed', function (handle) {
 							next();
--- a/cli/hooks/wp-run.js
+++ b/cli/hooks/wp-run.js
@@ -210,23 +210,21 @@ exports.init = function (logger, config, cli) {
 							skipLaunch: true
 					}, opts),
 					appxExtensions = ['.appx', '.appxbundle'],
-					installs = [],
-					isWin10Device = (builder.target == 'wp-device' && builder.wpsdk == '10.0');
+					installs = [];
 
 				function installApp(deviceId, xapFile, opts, next) {
 					// Now install the real app
+					var installed = false;
 					logger.info(__('Installing the application...'));
 					windowslib.install(deviceId, xapFile, appc.util.mix({
-								appGuid: builder.phoneProductId,
-								// launching an app on Win 10 device fails right now!
-								skipLaunch: isWin10Device
+								appGuid: builder.phoneProductId
 							}, opts))
+						.on('launched', function () {
+							logger.info(__('Finished launching the application'));
+						})
 						.on('installed', function (handle) {
-							if (isWin10Device) {
-								logger.warn(__('Due to a bug in Windows tooling, we are unable to launch apps for you on Windows 10 Mobile devices. Please launch the app manually.'));
-							} else {
-								logger.info(__('Finished launching the application'));
-							}
+							installed = true;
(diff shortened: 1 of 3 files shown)


infosia:
build win 10

infosia:
build win 10

garymathews:
When I connect my Windows 10 device and build a Titanium project
```

Which device do you want to install your app on?
   1)  Device                    (udid: 8-0-0)
   2)  Device                    (udid: 8-1-0)
   3)  Nokia Lumia 909 (RM-877)  (udid: 00000011-0e33-8fcc-0000-000000000000)
   4)  Device                    (udid: 10-0-0)
Select by number or name [Device]:
```

Both ``1`` and ``2`` are not valid since my Device isn't a Windows 8 or 8.1 device

sgtcoolguy:
Previously we did not set a wpsdk on any device. I changed it to set an explicit value for each SDK we detected.
If I kept the SDK as null, then choosing the actual device listing didn't end up using the right tooling for a win 10 device. ( I.e. The udid 0 placeholder worked, but the real udid  failed). So I'm not sure the best way to handle this. We don't really know the actual SDK version on the devices. I guess when we go to connect/install maybe we should try the toolchain for 10, then 8.1, then 8? Because as it stands it we check to see if the SDK value is '10.0' to determine if we use wptool or DeployAppCmd to install/launch apps. And for actual devices we end up using AppDeplyCmd instead of wptool, resulting in an "Access denied" error.

On Apr 25, 2016, at 9:22 PM, Gary Mathews <notifications@github.com<mailto:notifications@github.com>> wrote:


When I connect my Windows 10 device and build a Titanium project

Which device do you want to install your app on?
   1)  Device                    (udid: 8-0-0)
   2)  Device                    (udid: 8-1-0)
   3)  Nokia Lumia 909 (RM-877)  (udid: 00000011-0e33-8fcc-0000-000000000000)
   4)  Device                    (udid: 10-0-0)
Select by number or name [Device]:


Both 1 and 2 are not valid since my Device isn't a Windows 8 or 8.1 device

-

garymathews:
The reason the real ``udid`` device (entry ``3``) failed to use the correct tooling is because it was detected by [wptool.js#L150](https://github.com/garymathews/windowslib/blob/ea068a23779f0b3600bb7e0817e54310105dc641/lib/wptool.js#L150) which doesn't return any version information about the device. However, using the ``wptool`` binary does return the device specific ``wpsdk``, but can't determine the real ``udid`` for the device, which is how entry ``4`` is found.

```
3)  Nokia Lumia 909 (RM-877)  (udid: 00000011-0e33-8fcc-0000-000000000000)
4)  Device                    (udid: 10-0-0)
```

Since these are both the same device, detected by different tooling. Why not using the ``wpsdk`` detected in ``4`` and consolidate it with ``3``?

sgtcoolguy:
The udid 0 entries are basically just hard-coded entries that either we create or the native tooling returns. The device may or may not actually exist.
The "real" entry comes from WinAppDeployCmd and there's no guarantees about what version OS it has. Before I upgraded my phone, it showed up in the listing using Win 10 tooling (even though it was an 8.1U1).
So back to the original issue, I didn't specify --wp-SDK, so it's trying to guess at the SDK and the device listing doesn't have that info unless I cheat and hard-code it to assume 10.0 from WinAppDeployCmd listings.

So any ideas what we should do? For a given device we don't know what SDK to use. Sometimes we can cheat because the name says 10 or 8 in its name - but that's not reliable like in the Lumia case here.

I think we'll need to insist on specifying --wp-sdk for device builds or just list a device for each version like I do here.

On Apr 25, 2016, at 10:04 PM, Gary Mathews <notifications@github.com<mailto:notifications@github.com>> wrote:


The reason the real udid device (entry 3) failed to use the correct tooling is because it was detected by wptool.js#L150<https://github.com/garymathews/windowslib/blob/ea068a23779f0b3600bb7e0817e54310105dc641/lib/wptool.js#L150> which doesn't return any version information about the device. However, using the wptool binary does return the device specific wpsdk, but can't determine the real udid for the device, which is how entry 4 is found.

3)  Nokia Lumia 909 (RM-877)  (udid: 00000011-0e33-8fcc-0000-000000000000)
4)  Device                    (udid: 10-0-0)


Since these are both the same device, detected by different tooling. Why not using the wpsdk detected in 4 and consolidate it with 3?

-

garymathews:
The ``wpsdk`` used to be the ``wpsdk`` passed into the ``wptool`` binary, but this PR changed that: [windowslib#37](https://github.com/appcelerator/windowslib/pull/37/files), so it sets the ``wpsdk`` compatible with the device. Which also means, if the ``udid`` entry has a ``wpsdk != null`` then the device must exist since this value is determined from the device itself. Maybe we could use this to set the ``wpsdk`` in the real ``udid`` entry?

sgtcoolguy:
@garymathews You lost me there :)

I was referring to the device entries themselves, they used to keep a value of null for wpsdk. My PR changes it to hold the wpsdk value under which the device is being grouped, or '10.0' if we got it from WinAppDeployCmd.exe.

Your PR set the global wpsdk value to be the one from the device/emulator so we didn't need to set --wp-sdk manually if we knew the sdk version the device was. Which is fine/great for emulators. But for devices we don't really know what OS they actually have. Before I represented that by keeping the device wpsdk value null. Now I'm sort of indirectly forcing the user to choose an --wp-sdk by choosing the device entry with the sdk version they say it's running.

So unless we have some way of knowing if a device is actually running 8.1 or 10.0, I think we either need to keep the device wpsdk property null and force the user to specify --wp-sdk for wp-device builds; or we add a device entry to each idk version, setting it's wpsdk property to that version and have the user choose the device+sdk combo. In that scenario, the "real" devices from the WinAppDeployCmd.exe listing are still an issue in that we do not know what OS they have. So I think we should just keep them all having an wpsdk value of null and force users to specify --wp-sdk for device builds.

Unless you can think of some way to detect device OSes?

sgtcoolguy:
Hmm, ok looking at that other PR, it looks like you are getting the actual wpsdk of the device from wptool binary somehow. So it'd also make sense to "merge" the device listing with udid 0 and the single "actual" device we get from WinAppDeployCmd.exe - then we won't have two entries, just one with the real name/udid and wpsdk value...

sgtcoolguy:
@garymathews Ready to re-review. Updated this and the windowslib PRs.

cli/hooks/wp-run.js
```diff
@@ -213,15 +210,23 @@ exports.init = function (logger, config, cli) {
...
+					isWin10Device = (builder.target == 'wp-device' && builder.wpsdk == '10.0');
 
 				function installApp(deviceId, xapFile, opts, next) {
 					// Now install the real app
+					logger.info(__('Installing the application...'));
 					windowslib.install(deviceId, xapFile, appc.util.mix({
-								appGuid: builder.phoneProductId
+								appGuid: builder.phoneProductId,
+								// launching an app on Win 10 device fails right now!
+								skipLaunch: isWin10Device
```
garymathews: If I set this to ``false`` the app launches fine on my Windows 10 Mobile device, it does take 30-60 seconds from seeing ``Installing the application...`` for it to launch on the phone though

garymathews:
I have merged your windowslib changes. Although, currently a project will still try to build to device even if a device is not connected. I have created a [PR](https://github.com/appcelerator/windowslib/pull/42) to omit any invalid devices, meaning it reports ``No Windows Phone devices found`` when no devices are connected.

I have tried Windows 10 Mobile device app launching on two different machines and devices, it worked fine on both. I don't think we should disable it

sgtcoolguy:
Any time I tested on my device connected via USB, launching failed. I don't understand why it would work for you. The wptool binary explicitly guards against doing a connect for Windows 10 devices, but doesn't for launch for some reason. When I try launch the connection fails with "Access denied". I never paired with the phone so maybe that's the cause of it?

I think it'd be good to break apart installing and launching (or somehow use an error code or something) so we can handle a failed launch by basically asking the user to manually open the app rather than failing altogether (ending the CLI process).

On Apr 26, 2016, at 10:42 PM, Gary Mathews <notifications@github.com<mailto:notifications@github.com>> wrote:


I have merged your windowslib changes. Although, currently a project will still try to build to device even if a device is not connected. I have created a PR<https://github.com/appcelerator/windowslib/pull/42> to omit any invalid devices, meaning it reports No Windows Phone devices found when no devices are connected.

I have tried Windows 10 Mobile device app launching on two different machines and devices, it worked fine on both. I don't think we should disable it

-

sgtcoolguy:
Opened appcelerator/windowslib#43 to fire separate 'installed' and 'launched' events. If we see an 'installed' event and then an 'error' we spit out a warning message to the Cli asking the user to manually open the app - where before we'd just error out of the build entirely.

PR #648, opened 2016-04-27
[TIMOB-23274] Fix: No warn logging from internal TITANIUM_LOG_WARN

infosia:
[TIMOB-23274](https://jira.appcelerator.org/browse/TIMOB-23274)

`TITANIUM_LOG_WARN` has been used only for internal debugging purpose from Visual Studio and we have not been quite aware of logging in Titanium CLI. So replacing entire `TITANIUM_LOG_WARN` to `TITANIUM_MODULE_LOG_WARN` would not fit in this case...we'll need to "choose" what to show manually.

So basic logging policy here is:

- Show all "not supported on Windows" warnings
- Show all "API usage" warnings
- Show other critical errors that devs needs to be notified
- Do not show internal implementation detail (C++ cast error, C++ exception message etc)
- Do not show internal warning message (`fireEvent: No listeners for event xxx` etc)

This fix also includes argument spec changes in C++ macro to align with `TITANIUM_LOG_INFO`: Now all logging macros implement [TiLoggerPolicyInterface](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/include/Titanium/detail/TiLoggerPolicyInterface.hpp) so we can provide consistent logging based on `TiLogger`.

```diff
- TITANIUM_MODULE_LOG_INFO("value: " << value); 
+ TITANIUM_MODULE_LOG_INFO("value: ", value); 
```

Note that you can't use `TITANIUM_MODULE_LOG_WARN` in `TitaniumKit`...because `TITANIUM_MODULE_LOG_WARN` contains Windows-specific implementation (by design, `TitaniumKit` can't contain platform-specific implementation).


infosia:
build win 10

infosia:
build win 10

infosia:
Just pinging, @sgtcoolguy @garymathews Can either of you review/merge this?

Source/Filesystem/src/Filesystem.cpp
```diff
@@ -7,12 +7,13 @@
 #include "TitaniumWindows/Filesystem.hpp"
 #include "Titanium/Filesystem/FileStream.hpp"
 #include "Titanium/detail/TiImpl.hpp"
+#include <windows.h>
```
sgtcoolguy: Do we need windows.h here?

Source/Media/src/Media.cpp
```diff
@@ -10,13 +10,15 @@
 #include "Titanium/Media/Item.hpp"
 #include "Titanium/Media/CameraOptionsType.hpp"
 #include "Titanium/Media/MediaQueryType.hpp"
-#include "TitaniumWindows/Utility.hpp"
 #include "Titanium/detail/TiImpl.hpp"
-#include "TitaniumWindows/AppModule.hpp"
-#include "TitaniumWindows/WindowsMacros.hpp"
+#include <windows.h>
```
sgtcoolguy: do we always need this header when using the log forwarder? Maybe the include should be there?

PR #660, opened 2016-05-03
[TIMOB-23293] Fix layout issue on Corporate Directory App

infosia:
[TIMOB-23293](https://jira.appcelerator.org/browse/TIMOB-23293)

Corporate Directory App had layout issue. Pushing some improvements:

- ScrollView should fill views
- Label.textAlign should accept "center","left","right"

It still doesn't look very beautiful but pushing since that's not a regression for 5.3.x.

capture


garymathews:
CR: **PASS**
FR: **PASS**

sgtcoolguy:
@infosia @garymathews Can you please open a JIAR ticket for addressing the remaining issues? I'm not aware what specifically they are. I'm assuming it's what's shown up above?
if so then:
- The text using the icon font is getting cut off at the bottom
- the background fill for the button is bleeding through over the border

PR #667, opened 2016-05-04
WIP [TIMOB-23276] Windows: Unable to change color Tab/Window titles

sgtcoolguy:
This is a work in progress PR.

This attempts to add the activeColor and titleColor properties to Ti.UI.Tab to enable users to change the tab title FG color for active/inactive tabs.

Seems ot mostly work, though I'm not sure how to revert back to the default Windows colors for a tab if we change to a state where we didn't override the property - i.e. if we set activeColor, but not titleColor and the tab becomes inactive - right now it keeps the activeColor. It should go back to the "default" inactive tab color according to Windows theming.

Some links that may help fix that:
- http://stackoverflow.com/questions/31797875/overriding-pivot-header-foreground-brushes-in-uwp-app-win-10-rtm-sdk
- http://stackoverflow.com/questions/17902064/how-to-change-pivot-header-template-in-windows-phone-8


sgtcoolguy:
Closed in favor of #675

PR #671, opened 2016-05-06
DO NOT MERGE - Check if skipping Contacts.getAllPeople reliably passes the test

infosia:
TIMOB23332.

*DO NOT MERGE THIS PR*


infosia:
So I think Contact.getAllPeople is not actually a source of the crash because Jenkins Windows 10 build still fails. I still can not reproduce this locally.

PR #672, opened 2016-05-07
Fix: "no matching function to call transform" on VS 2015

infosia:
I have experienced `no matching function to call std::transform` compile error for UWP 10.0 app project on freshly-installed Windows 10 + Visual Studio 2015. It looks like we have a workaround (See thread in [stackoverflow](http://stackoverflow.com/questions/7131858/stdtransform-and-toupper-no-matching-function) for details) and see how it goes.


garymathews:
build win 10

PR #675, opened 2016-05-09
[TIMOB-23276] Windows: Unable to change color Tab/Window titles

infosia:
#667 + Reverting back to default color when color is empty


sgtcoolguy:
build win 10

PR #683, opened 2016-05-12
[TIMOB-23283] Fix ProgressBar with FILL should fill the view

infosia:
[TIMOB-23283](https://jira.appcelerator.org/browse/TIMOB-23283)

```javascript
var pb = Ti.UI.createProgressBar({
    min: 0,
    max: 10,
    value: 0,
    color: 'blue',
    backgroundColor: 'green',
    backgroundDisabledColor: 'gray',
    message: 'Downloading 0 of 10',
    width: Ti.UI.FILL
});
win.add(pb);
win.open();
pb.show();
```


sgtcoolguy:
**CR** looks good. Need to do **FR** now...

PR #684, opened 2016-05-12
[TIMOB-23363] Windows: textAlign property of Button does not work in 8.1 or 10

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-23363

Note that this also fixes an incorrect implementation of verticalAlign property.

For certain controls, like Button, they are subclasses of ContentControl - meaning they have a Content property which can typically take a simple String or an actual complex UI as the value. For these controls, to be able to manage their content they have special properties:
- HorizontalContentAlignment
- VerticalContentAlignment

We just need to use those to set textAlign/verticalAlign on the internal text.


infosia:
build win 10

sgtcoolguy:
fixed

PR #686, opened 2016-05-13
[TIMOB-23372]  [TIMOB-23373] Fix resize issue with Label

infosia:
- [[TIMOB-23372]](https://jira.appcelerator.org/browse/TIMOB-23372) Fix: layout with left & right breaks layout
- [[TIMOB-23373]](https://jira.appcelerator.org/browse/TIMOB-23373) Fix: Label with height should render text vertically centered

*TIMOB-23372*

timob23372

```javascript
var win = Ti.UI.createWindow({ backgroundColor: 'green' });

var view = Ti.UI.createView({
    backgroundColor: 'yellow',
    layout: "horizontal",
    top: 10,
    left: 10,
    height: Ti.UI.SIZE,
    width: Ti.UI.SIZE,
});

var label = Ti.UI.createLabel({
    left: 10,
    right: 10,
    color: "blue",
    text: "this is test text"
});

view.add(label);

win.add(view);
win.open();
```

*TIMOB-23373*

timob23373

```javascript
var win = Ti.UI.createWindow({ backgroundColor: 'white' });

var view = Ti.UI.createLabel({
    text: "A",
    height: 100,
    width: 100,
    color: "#C41230",
    backgroundColor: "#F2D0D6",
    textAlign: "center",
});

win.add(view);
win.open();
```

PR #688, opened 2016-05-13
[5_3_X][TIMOB-23374] Windows: Label backgroundColor cut through border with radius

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-23374

This simply applies background brushes (color/image/gradient) to the wrapping Border object when we have a non-null border__ (and don't have an underlying_control__).

I have no idea the ordering of when this should apply: I stuck it before the is_panel and is_control checks, but honestly I don't know under what circumstance we should move the background to the wrapping border and when we shouldn't.

When merged, this needs to be cherry-picked to master.

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -643,6 +643,8 @@ namespace TitaniumWindows
 		{
 			if (underlying_control__) {
 				underlying_control__->Background = brush;
+			} else if (border__) {
+				border__->Background = brush;
 			} else if (is_panel__) {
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
label_border_fixed


We're getting much closer here. The upper content seems to be shifted off-screen a little to the right for some reason now.

infosia:
Awesome!!!

infosia:
The reason why upper content have some offset is because `Label` have a issue with `ScrollView` like I commented in https://github.com/appcelerator/titanium_mobile_windows/pull/686#issuecomment-219025475 . I expect our `LayoutEngine` does something wrong. I will add new ticket for it and will be working on it.

PR #704, opened 2016-05-18
[TIMOB-23392] Analytics: Timestamp for Windows payloads are inconsistent with other platforms

sgtcoolguy:
https://jira.appcelerator.org/browse/TIMOB-23392

Basically it boils down to us using the ISO string format and the other platforms using a "+0000" timezone offset instead of "Z" at the end. So I generate our ISO string and replace "Z" with "+0000".

### DIFF
--- a/Source/TitaniumKit/src/analytics.js
+++ b/Source/TitaniumKit/src/analytics.js
@@ -53,6 +53,16 @@ Analytics.prototype.saveEventQueue = function saveEventQueue() {
 };
 
 /**
+ * Converts an ISO Date string to the format required by our analytics server.
+ * (Basically replace the trailing 'Z' with +0000)
+ * @param  {String} dateString Date.prototype.toISOString()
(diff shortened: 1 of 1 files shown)


infosia:
build win 10

infosia:
build win 8.1

infosia:
Jenkins says `L` is not found, I think this means `Ti.Analytics._start();` throws exception during `Analytics::loadJS` in this case. From Jenkins console:

```javascript
22:09:39 [ERROR]: Failed to initialize Titanium: SyntaxError: Unexpected token ';'. Expected '}' to end a object literal.
```

garymathews:
build win 8.1

garymathews:
CR: **PASS**
FR: **PASS**

```Javascript
{
  ...
  "ts": "2016-05-19T19:22:40.717+0000",
  ...
}
```

PR #690, opened 2016-05-13
[TIMOB-23372] Fix Composite right and bottom margins

garymathews:
- Fix ``right`` and ``bottom`` margins in ``Composite`` ``Horizontal`` and ``Vertical`` of LayoutEngine
- The layout engine did not take into account the ``right`` offset when calculating the new width due to an incorrect ``if`` statement
- If a component is of width ``Ti.UI.SIZE`` and the ``right`` margin is specified, it should calculate the new width taking into account both ``left`` and ``right`` margins
- Made a workaround to match iOS and Android when percentage values of ``left`` and ``right`` are set inside a parent View

###### TEST CASE # 1
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'red' }),
    view = Ti.UI.createView({
        backgroundColor: 'orange',
        layout: "vertical",
        top: 10,
        left: 10,
        right: 10,
        height: Ti.UI.SIZE,
        width: Ti.UI.SIZE,
    }),
    label = Ti.UI.createLabel({
        left: 10,
        right: 10,
        color: "green",
        backgroundColor: 'yellow',
        text: "this is test text"
    }),
    button = Ti.UI.createButton({
        title: 'BUTTON',
        backgroundColor: 'green',
        left: 10,
        width: Ti.UI.SIZE
    });
view.add(label);
view.add(button);
win.add(view);
win.open();
```
###### TEST CASE # 2
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'red' }),
    a = Ti.UI.createView({
        backgroundColor: 'orange',
        top: 10,
        left: 10,
        right: 10,
        bottom: 10,
    }),
    b = Ti.UI.createView({
        backgroundColor: 'yellow',
        top: 10,
        left: 10,
        right: 10,
        bottom: 10,
    });
a.add(b);
win.add(a);
win.open();
```

###### TEST CASE # 3
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'red' }),
    label = Ti.UI.createLabel({
        left: 10,
        right: 10,
        backgroundColor:'yellow',
        color: "green",
        text: "this is test text"
    });
win.add(label);
win.open();
```

###### TEST CASE # 4
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'green' }),
    view = Ti.UI.createView({
        backgroundColor: 'yellow',
        layout: "horizontal",
        top: 10,
        left: 10,
        right: 10,
        height: Ti.UI.SIZE,
[...]

PR #709, opened 2016-05-20
[TIMOB-23372] Fix: right margin breaks layout

infosia:
[TIMOB-23372](https://jira.appcelerator.org/browse/TIMOB-23372)

#690 + unit tests + fixes on Label

- Fix ``right`` and ``bottom`` margins in ``Composite`` ``Horizontal`` and ``Vertical`` of LayoutEngine
- The layout engine did not take into account the ``right`` offset when calculating the new width due to an incorrect ``if`` statement
- If a component is of width ``Ti.UI.SIZE`` and the ``right`` margin is specified, it should calculate the new width taking into account both ``left`` and ``right`` margins
- Made a workaround to match iOS and Android when percentage values of ``left`` and ``right`` are set inside a parent View

###### TEST CASE # 1

```javascript
    view = Ti.UI.createView({
        backgroundColor: 'yellow',
        top: 10,
        left: 10,
        right: 10,
        height: Ti.UI.SIZE,
        width: Ti.UI.SIZE,
    }),
    label = Ti.UI.createLabel({
        left: 10,
        right: 10,
        color: "blue",
        backgroundColor: 'orange',
        text: "this is test text"
    });
view.add(label);
win.add(view);
win.open();
```

###### TEST CASE # 2

Note: On iOS, view is not scaled correctly due to [TIMOB-23385](https://jira.appcelerator.org/browse/TIMOB-23385).

```javascript
    view = Ti.UI.createView({
        backgroundColor: 'yellow',
        top: 10,
        left: 10,
        height: Ti.UI.SIZE,
        width: Ti.UI.SIZE,
    }),
    label = Ti.UI.createLabel({
        left: 10,
        right: 10,
        color: "blue",
        backgroundColor: 'orange',
        text: "this is test text"
    });
view.add(label);
win.add(view);
win.open();
```

### DIFF
--- a/Source/LayoutEngine/src/Horizontal.cpp
+++ b/Source/LayoutEngine/src/Horizontal.cpp
@@ -49,6 +49,7 @@ namespace Titanium
 			unsigned int len = children.size();
 			unsigned int rowLen = 0;
 			unsigned int rowsLen = (len > 0) ? 1 : 0;
+			bool percentageFix = false;
 
 			// Calculate horizontal size and position for the children
 			for (i = 0; i < len; i++) {
@@ -60,6 +61,14 @@ namespace Titanium
 				heightLayoutCoefficients = layoutCoefficients.height;
 				sandboxWidthLayoutCoefficients = layoutCoefficients.sandboxWidth;
 				leftLayoutCoefficients = layoutCoefficients.left;
+
+				// For some reason, iOS and Android append 10% right padding to horizontal
+				// views of width Ti.UI.SIZE when an element inside uses % padding this is
+				// a workaround to offset the padding and keep parity with iOS and Android
+				if (isWidthSize && !percentageFix && leftLayoutCoefficients.x1 > 0) {
+					runningWidth -= 0.1* width;
+					percentageFix = true;
+				}
 
 				// Pedro
 				measuredWidth = widthLayoutCoefficients.x1 * width + widthLayoutCoefficients.x2 * (width - (isWidthSize ? 0 : runningWidth)) + widthLayoutCoefficients.x3;
@@ -102,7 +111,7 @@ namespace Titanium
 
 				(*child).measuredLeft = measuredLeft;
 
-				if (rowsLen >= rows.size()) {
+				while (rowsLen >= rows.size()) {
 					rows.push_back(std::vector<Element*>());
 				}
 				if (rowLen >= rows[rowsLen - 1].size()) {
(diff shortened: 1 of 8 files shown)


Source/LayoutEngine/src/Horizontal.cpp
```diff
@@ -102,7 +111,7 @@ namespace Titanium
 
 				(*child).measuredLeft = measuredLeft;
 
-				if (rowsLen >= rows.size()) {
+				while (rowsLen >= rows.size()) {
```
infosia: Note: Changed `if` to `while` because when `rowLen - rows.size() > 1` the line `rows[rowsLen - 1]` below should crash. I actually saw it crashes on WP 8.1. (I'm not sure why it didn't crash on my Win10 phone though)

Source/UI/src/Label.cpp
```diff
@@ -45,27 +44,6 @@ namespace TitaniumWindows
 			parent__ = ref new Controls::Grid();
 			label__ = ref new Windows::UI::Xaml::Controls::TextBlock();
 
-			// In case width/height is Ti.UI.SIZE, grid needs to resize when label text is changed
-			label_sizechanged_event__ = label__->SizeChanged += ref new SizeChangedEventHandler([this](Platform::Object^ sender, SizeChangedEventArgs^ e) {
```
infosia: Note: Stopped handling `SizeChangedEventHandler` here and let `WindowsLayoutDelegate` to handle this by setting `use_own_size__` flag using `WindowsLayoutDelegate::useOwnSize()`.

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -1312,10 +1312,33 @@ namespace TitaniumWindows
...
 		}
 
+
+		bool WindowsViewLayoutDelegate::shouldUseOwnWidth() const TITANIUM_NOEXCEPT
+		{
+			// Check if parent can't decide its width and this component doesn't have fixed width property & handles its width
+			if (use_own_size__ && layout_node__->parent != nullptr) {
+				return layout_node__->properties.width.valueType != Titanium::LayoutEngine::ValueType::Fixed
+					&& layout_node__->parent->properties.right.valueType == Titanium::LayoutEngine::ValueType::None
+					&& layout_node__->parent->properties.width.valueType == Titanium::LayoutEngine::ValueType::Size;
```
infosia: Note: As we saw in our test cases, when parent width is `Ti.UI.SIZE`, parent view should respect its child view size.

PR #714, opened 2016-05-24
[TIMOB-23305] Ti.UI.Label width is not updated when setting new text

infosia:
[TIMOB-23305](https://jira.appcelerator.org/browse/TIMOB-23305)

Attempt to update width when setting new text. Currently it's blocked because Xaml `TextBlock` does not shrink its size with shorter text.

* `TextBlock` does not change its size when `TextWrapping` is `Wrap`.
* `TextBlock` expands its size when `TextWrapping` is `NoWrap`.
* `TextBlock` does not shrink its size in every cases of text wrapping.
* ~~We might need to find a way to calculate its size for the given text and font manually.~~ <- `Measure()` works good for it

```javascript
    backgroundColor: 'green',
    layout: 'vertical'
});

var update = true;

var topView = Ti.UI.createView({ backgroundColor: 'blue', width: Ti.UI.FILL, height: '10%' });
var button = Ti.UI.createButton({ title: 'Change Text' });
topView.add(button);

    text: update ? 'Lorem ipsum' : 'Lorem ipsum dolor sit amet.',
    backgroundColor: 'orange'
});

button.addEventListener('click', function () {
    label.text = update ? 'Lorem ipsum dolor sit amet.' : 'Lorem ipsum';
    update = !update;
});

var bottomView = Ti.UI.createView({backgroundColor:'blue', width:Ti.UI.FILL, height: '10%'});

win.add(topView);
win.add(label);
win.add(bottomView);

win.open();
```

### DIFF
--- a/Source/UI/src/Label.cpp
+++ b/Source/UI/src/Label.cpp
@@ -89,10 +89,34 @@ namespace TitaniumWindows
 
 		void Label::set_text(const std::string& text) TITANIUM_NOEXCEPT
 		{
+			set_text(text, true);
+		}
+
+		void Label::set_text(const std::string& text, const bool resize) TITANIUM_NOEXCEPT
+		{
 			Titanium::UI::Label::set_text(text);
 			const auto new_text = TitaniumWindows::Utility::ConvertUTF8String(text);
-			if (label__->Text != new_text) {
-				label__->Text = new_text;
+
+			label__->Text = new_text;
+
+			if (propertiesSet__) {
+				measureDesiredSize();
+			}
+		}
+
+		void Label::measureDesiredSize() TITANIUM_NOEXCEPT
+		{
+			// TIMOB-23305: Measure desired size for this text and resize parent container accordingly.
+			Windows::Foundation::Size desiredSize{ static_cast<float>(label__->MaxWidth), static_cast<float>(label__->MaxHeight) };
+			label__->Measure(desiredSize);
+
+			const auto layout = getViewLayoutDelegate<WindowsViewLayoutDelegate>();
+
+			if (layout->get_width().empty() || layout->get_right().empty()) {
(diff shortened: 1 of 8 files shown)


sgtcoolguy:
It appears that the general solution people are proposing is to create a TextBlock, set the text/font, then call Measure and Arrange on it and grab the ActualWidth/ActualHeigth properties afterwards.

See:
- http://stackoverflow.com/questions/9126076/how-can-i-calculate-the-width-of-a-string-in-metro-without-displaying-it
- https://invokeit.wordpress.com/2012/09/19/how-to-measure-rendered-string-dimensions-in-win8dev/

It may be that we just need to call something like UIElement.UpdateLayout for the parent or something?

infosia:
`Measure()` is working good for TextBlock 👍 Now I'm trying to ensure it doesn't break anything...

infosia:
build win 10

infosia:
build win 10

garymathews:
build win 10

garymathews:
build win 10

Source/UI/src/Label.cpp
```diff
@@ -64,6 +64,18 @@ namespace TitaniumWindows
 
 			const auto layout = getViewLayoutDelegate<WindowsViewLayoutDelegate>();
 
+			layout->postRequestLayoutHook__ = [this, layout]() {
```
garymathews: Is this needed, the label seem to scale correctly first time without this?

infosia: Oops right it does work without this...I think I was confused. I saw `rect` property is not update so I was to make sure LayoutEngine calculate it again. I'll update to better one.

garymathews:
CR: **PASS**
FR: **PASS**

:+1:

infosia:
build win 10

PR #716, opened 2016-05-24
Tool to convert Titanium app to Example

sgtcoolguy:
Add a script to convert a Titanium project that has been run with {{ti build -p windows}} into an example in our Examples dir (which we can then use another script to generate a VS sln file to run that app against debugger/source)

Effectively we do this:
```
cd C:/path/to/projectproject
ti build -p windows #...
cd C:/path/to/titanium_mobile_windows/Tools/Scripts
npm install
node convert_project_to_example.js C:/path/to/project
node generate_project.js ProjectName -s 10.0
```

This _should_ get us from a user app to a VS solution that we can use to do much deeper debugging and try out fixes to our SDK against their app. Beats having to hand-craft copies of apps into our Examples folder like we have in the past.


garymathews:
Looks good! But would it be better to have an .ejs template for the CMakeLists.txt? Just in case the NG CMakeLists.txt changes?

PR #718, opened 2016-05-25
[TIMOB-23411] Catch InvalidArgumentException in requiredModuleExists()

garymathews:
- Catch an uncaught ``InvalidArgumentException`` in ``requiredModuleExists()``
- This can happen when the ``StorageFile`` path is invalid
- An attempt to fix [TIMOB-23411](https://jira.appcelerator.org/browse/TIMOB-23411)

### DIFF
--- a/Source/Global/src/GlobalObject.cpp
+++ b/Source/Global/src/GlobalObject.cpp
@@ -91,6 +91,8 @@ namespace TitaniumWindows
 			try {
 				task.get();
 				exists = true;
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
InvalidArgumentException is a subclass of COMException, so I don't think this should matter: https://msdn.microsoft.com/en-us/library/windows/apps/xaml/hh755794.aspx

garymathews:
Yep, you're right. I've just never seen this exception thrown before and thought we weren't handling it. I'll close this PR.

PR #719, opened 2016-05-25
[TIMOB-19351] Support unit values for Ti.UI.createAnimation()

garymathews:
- Add support for ``% cm mm cm em pt pc in px dp dip`` in ``width`` ``height`` ``top`` ``left`` (``bottom`` and ``right`` are not implemented)
- Fixed crash when animating ``width`` or ``height`` from ``0``
- Fixed ``width`` and ``height`` not being retained after animation

###### TEST CASE # 1
``Scale height from 0 to 100%``
```Javascript
var win = Ti.UI.createWindow({backgroundColor: 'black'}),
    view = Ti.UI.createView({backgroundColor: 'red', top: 0, left: 0, width:'10%', height: 0}),
    animation = Ti.UI.createAnimation({height: "100%", curve: Ti.UI.ANIMATION_CURVE_EASE_OUT, duration: 2000});
view.animate(animation);
win.add(view);
win.open();
```

###### TEST CASE # 2
``Scale width from 0 to 100%``
```Javascript
var win = Ti.UI.createWindow({backgroundColor: 'black'}),
    view = Ti.UI.createView({backgroundColor: 'red', top: 0, left: 0, width:'0', height: '10%'}),
    animation = Ti.UI.createAnimation({width: '100%', curve: Ti.UI.ANIMATION_CURVE_EASE_OUT, duration: 2000});
view.animate(animation);
win.add(view);
win.open();
```

###### TEST CASE # 3
``Transform left from 0 to 90%``
```Javascript
var win = Ti.UI.createWindow({backgroundColor: 'black'}),
    view = Ti.UI.createView({backgroundColor: 'red', top: 0, left: 0, width:'10%', height: '10%'}),
    animation = Ti.UI.createAnimation({left: '90%', curve: Ti.UI.ANIMATION_CURVE_EASE_OUT, duration: 2000});
view.animate(animation);
win.add(view);
win.open();
```

###### TEST CASE # 4
``Transform top from 0 to 90%``
```Javascript
var win = Ti.UI.createWindow({backgroundColor: 'black'}),
    view = Ti.UI.createView({backgroundColor: 'red', top: 0, left: 0, width:'10%', height: '10%'}),
    animation = Ti.UI.createAnimation({top: '90%', curve: Ti.UI.ANIMATION_CURVE_EASE_OUT, duration: 2000});
view.animate(animation);
win.add(view);
win.open();
```

[TIMOB-19351](https://jira.appcelerator.org/browse/TIMOB-19351)

### DIFF
--- a/Source/TitaniumKit/include/Titanium/UI/Animation.hpp
+++ b/Source/TitaniumKit/include/Titanium/UI/Animation.hpp
@@ -60,7 +60,7 @@ namespace Titanium
 			  @abstract bottom
 			  @discussion Value of the `bottom` property at the end of the animation.
 			*/
-			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<double>, bottom);
+			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<std::string>, bottom);
 
 			/*!
 			  @property
@@ -102,14 +102,14 @@ namespace Titanium
 			  @abstract height
 			  @discussion Value of the `height` property at the end of the animation.
 			*/
-			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<double>, height);
+			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<std::string>, height);
 
 			/*!
 			  @property
 			  @abstract left
 			  @discussion Value of the `left` property at the end of the animation.
 			*/
-			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<double>, left);
+			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<std::string>, left);
 
 			/*!
 			  @property
@@ -137,14 +137,14 @@ namespace Titanium
 			  @abstract right
 			  @discussion Value of the `right` property at the end of the animation.
 			*/
-			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<double>, right);
+			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<std::string>, right);
 
 			/*!
 			  @property
 			  @abstract top
 			  @discussion Value of the `top` property at the end of the animation.
 			*/
-			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<double>, top);
+			TITANIUM_PROPERTY_IMPL_DEF(boost::optional<std::string>, top);
 
 			/*!
 			  @property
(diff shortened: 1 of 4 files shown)


garymathews:
I will update the PR with Mocha test cases soon

infosia:
I see some test cases are failing for Windows 10.  http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_10_prs/616/ @garymathews Could you look into this?

garymathews:
Yes, I enabled the previously failing animation test cases since they worked locally for me. For some reason they don't work on the Jenkins server? I'll take a look, but there's no error description

sgtcoolguy:
@garymathews The lack of an error message is related to the errors getting "squashed" that I mentioned in a previous PR and opened up https://jira.appcelerator.org/browse/TIMOB-23439 for. Basically if the assertion is in an event callback or some async code and fails, the error just kind of gets ignored and the test times out - so the error is the timeout (though, we should show a better error message for that case. The 'message' property of some errors doesn't get sent along in the JSON.stringify call in the mocha app's code, probably because it's not an "own property" of the error subclass thrown).

sgtcoolguy:
build win 10

garymathews:
I will revert to skipping the tests on Windows 10 Store

garymathews:
build win 10

sgtcoolguy:
For the first two tests, I don't see the view height/width actually animating. I just see the start and end state.

Because I'm working on a common test suite, the tests are causing a merge conflict.  Here's the resolved ti.ui.view.test.js: https://gist.github.com/sgtcoolguy/528a9690b6c862e521047329d9750a2e

sgtcoolguy:
Playing around with this some, I think the issue stems from Windows having serious issues with the easing function applied against such a drastic change in scale. Effectively if we're scaling from 0 to 100% of the parent (the full height of the phone), I think the easing function just goes nuts and we basically go from start to end state immediately. If I change the first test to start from a height of 100 to 100% it looks fine. I'm not really sure what to make of this. Maybe just document it?

Actually, even setting it to a height of 1 first fixes it to work fine. So maybe when we correct for 0 we also need to do the same for the component, because the easing function can't handle starting at 0?

sgtcoolguy:
@garymathews So, I think we need to check for a height of 0 and try setting the component's height to 1 before starting the animation, or maybe explicitly setting the From property on the DoubleAnimation to 1 for that case, so that the animation actually happens and doesn't just "appear" in the end state.

garymathews:
@sgtcoolguy I did have those checks in place. It seems to have broke. I've updated the PR, for some reason I need to set the width/height through ``setLayoutProperty`` in order for it to work again

PR #721, opened 2016-05-26
PoC: [TIMOB-23439] Unified way to catch exception

infosia:
*PROOF OF CONCEPT*

[TIMOB-23439](https://jira.appcelerator.org/browse/TIMOB-23439)

Introduce new unified way to catch exception inspired by #715 

* `TITANIUM_EXCEPTION_CATCH_START` indicates you starts to catch the exception 
* `TITANIUM_EXCEPTION_CATCH_END` indicates you catches the exception and show "red screen of death". Used in TitaniumKit.
* `TITANIUMWINDOWS_EXCEPTION_CATCH_START` is Windows-specific version of `TITANIUM_EXCEPTION_CATCH_END`. It catches `Platform::COMException` and show "red screen of death" with exception detail. 

```c++
TITANIUM_EXCEPTION_CATCH_START {
    throw ref new Platform::COMException(0, "TEST EXCEPTION");
}  TITANIUMWINDOWS_EXCEPTION_CATCH_END
```

In asyc functions, you need to use `concurrency::task` along with `concurrency::task::get()` to catch exception, because `get()` gives a chances to catch exception like below.

```c++
concurrency::create_task(Geolocator::RequestAccessAsync()).then([this](concurrency::task<GeolocationAccessStatus> task) {
	TITANIUM_EXCEPTION_CATCH_START {
		const auto status = task.get(); // this could throw new exception
		//
		// do something useful with the status then...
		//
	}  TITANIUMWINDOWS_EXCEPTION_CATCH_END
});
```

For more details [Geolocation.cpp](https://github.com/infosia/titanium_mobile_windows/blob/41c13085c7c7fd3b8a4f988ccf6cb0d872736284/Source/Sensors/src/Geolocation.cpp) would be a good example to see how to migrate existing codes.

### DIFF
--- a/Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
+++ b/Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
@@ -435,4 +435,14 @@ TITANIUM_PROPERTY_SETTER(MODULE, NAME) { \
 	return false; \
 }
 
+#define TITANIUM_EXCEPTION_CATCH_START try
+#define TITANIUM_EXCEPTION_CATCH_END \
+catch (const HAL::detail::js_runtime_error& ex) { \
+    std::ostringstream os; \
+    os << "Runtime Error: " << ex.js_message(); \
+    Titanium::Module::ShowRedScreenOfDeath(get_context(), os.str()); \
+} catch (...) { \
+    Titanium::Module::ShowRedScreenOfDeath(get_context(), "Unknown Exception"); \
+}
+
 #endif  // _TITANIUM_DETAIL_TIIMPL_HPP_
--- a/Source/Sensors/src/Geolocation.cpp
+++ b/Source/Sensors/src/Geolocation.cpp
@@ -7,7 +7,7 @@
  */
 
 #include "TitaniumWindows/Geolocation.hpp"
-#include "Titanium/detail/TiImpl.hpp"
+#include "TitaniumWindows/WindowsTiImpl.hpp"
 
 #include <boost/algorithm/string.hpp>
 #include <ppltasks.h>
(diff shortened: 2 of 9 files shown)


infosia:
@sgtcoolguy Please review, this is mainly for PoC and need review before proceeding further.

Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
```diff
@@ -435,4 +435,14 @@ TITANIUM_PROPERTY_SETTER(MODULE, NAME) { \
 	return false; \
 }
 
+#define TITANIUM_EXCEPTION_CATCH_START try
+#define TITANIUM_EXCEPTION_CATCH_END \
+catch (const HAL::detail::js_runtime_error& ex) { \
```
sgtcoolguy: It feels like in every case we should attempt to log the exception via Ti.API.error, then do red screen of death if we can.

Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
```diff
@@ -435,4 +435,14 @@ TITANIUM_PROPERTY_SETTER(MODULE, NAME) { \
 	return false; \
 }
 
+#define TITANIUM_EXCEPTION_CATCH_START try
```
sgtcoolguy: So I get that we likely want a macro pair for exception handling, but I'm having a hard time thinking of a case where the start is ever going to be more than a ``try {``

sgtcoolguy:
This looks Ok, though it feels a little clunky syntax-wise.

We might want to consider allowing the user to specify some custom exception handling in some cases (primarily on Windows exceptions, I'd assume)? I mean we have the catch-all COMException, so I guess we could catch the more specific subclass and add custom handling for "expected" exceptions (like AccessDenied or file not existing type stuff).

Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
infosia: You mean you would prefer `try { } TITANIUMWINDOWS_EXCEPTION_CATCH_END` ?

```c++
try {
    throw ref new Platform::COMException(0, "TEST EXCEPTION");
}  TITANIUMWINDOWS_EXCEPTION_CATCH_END
```

Source/TitaniumKit/include/Titanium/detail/TiImpl.hpp
infosia: It looks like...we are already doing so.

https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/src/Application.cpp#L35

```javascript
function Titanium_RedScreenOfDeath(e) {
...
    try {
        Ti.API.error("Application Error: "+JSON.stringify(e, null, 2));
...
```

infosia:
So when we need specific exception handling, we could do like this:

```c++
TITANIUM_EXCEPTION_CATCH_START {
    throw ref new Platform::COMException(0, "TEST EXCEPTION");
} catch (SpecificException^ e) {
    // do something specific to this exception
}  TITANIUMWINDOWS_EXCEPTION_CATCH_END
```

Yeah, I agree this is not a beautiful code at all... @sgtcoolguy @garymathews  Do you have something better in your mind for this?

PR #733, opened 2016-06-02
Copy over latest test suite

sgtcoolguy:
This is a work in progress to keep our tests suite for Windows up to date with the common suite I'm extracting from it. I'm actively trying to beef up the tests, add custom assertions to trim down test wordiness for common companions, etc.

Eventually we will need to combine the test driver script from node, and the app.js as well. Ideally we'd simply source the common test suite as a sub-module with some way to add tests locally in our repo for PRs that would get built up and then could eventually be copied into the common suite (otherwise we'd have to write the test and push it to the suite repo first?). I'm open to suggestions for how we can control this so that can continue to refine our test suite and use it across platforms without being an undue burden for adding unit tests for each new PR as we go.

### DIFF
--- a/Examples/NMocha/src/Assets/app.js
+++ b/Examples/NMocha/src/Assets/app.js
@@ -1,22 +1,23 @@
 /*
  * Appcelerator Titanium Mobile
- * Copyright (c) 2011-2015 by Appcelerator, Inc. All Rights Reserved.
+ * Copyright (c) 2011-2016 by Appcelerator, Inc. All Rights Reserved.
  * Licensed under the terms of the Apache Public License
  * Please see the LICENSE included with this distribution for details.
  */
 
 var win = Ti.UI.createWindow({
-    backgroundColor: 'lightyellow'
+	backgroundColor: 'yellow'
 });
 win.open();
 
 require('./ti-mocha');
 var $results = [],
-    failed = false;
+	failed = false;
 
 // ============================================================================
 // Add the tests here using "require"
 require('./ti.accelerometer.test');
+require('./ti.api.test');
 require('./ti.app.test');
 require('./ti.app.properties.test');
 require('./ti.app.windows.backgroundservice.test');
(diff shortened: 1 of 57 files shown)


sgtcoolguy:
@infosia I'm wondering if maybe the way I'm testing read-only properties is wrong. On both Android and Windows, I'm seeing what I _think_ is incorrect behavior. For a number of the properties marked as read-only, I seem to be able to assign new values to them.

For example: http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_10_prs/655/testReport/junit/Windows10.Store.Titanium/App/EVENT_ACCESSIBILITY_CHANGED/

Or any number of other failing tests (for read-only properties that aren't declared as constants, like Ti.App.version). My test verifies the type (String or Number, typically) and then assigns the existing value to a new variable, sets the property to some string and then checks to see if it changed:

```javascript
var value = Ti.UI.FILL;
Ti.UI.FILL = 'something else';
should(Ti.UI.FILL).eql(value);
```

That seems to be failing, and I'm not sure why. On Android, it seems like I can actually assign new values, I haven't manually tested on Windows, but that's my guess there as well.

Should I be testing the property descriptor writable value?
```javascript
Object.getOwnPropertyDescriptor(Ti.UI, "FILL").writable;
```

It's late here, so I'll have to investigate further tomorrow, but it looks like on Andorid/V* and Windows/JSC we're setting the ReadOnly property in C++ properly, and my understanding is that it should mean we **can't** assign new values...

infosia:
Manually tested, on Windows + JSC, it works as expected.

```javascript
Ti.API.info("before: " + Ti.UI.FILL);
Ti.UI.FILL = "test value";
Ti.API.info("after: " + Ti.UI.FILL);
```

goes to:

```
[INFO] before: LAYOUT_FILL
[INFO] after: LAYOUT_FILL
```

I'm not sure how V8 handles read-only string though.

sgtcoolguy:
build win 10

sgtcoolguy:
build win 8.1

sgtcoolguy:
build win 10

sgtcoolguy:
build win 10

infosia:
Closing as this is obsolete. We still need to try to keep up to date with our common suite though.

PR #737, opened 2016-06-06
[TIMOB-23296] Include language resources on packaging

infosia:
[TIMOB-23296](https://jira.appcelerator.org/browse/TIMOB-23296)
[TIMOB-23404](https://jira.appcelerator.org/browse/TIMOB-23404)

- Include language resources on packaging
- Add languages resources to samples

```javascript
    Ti.Locale.setLanguage('en');
    Ti.API.info('current language is: ' + Ti.Locale.getCurrentLanguage());
    Ti.API.info('before change ' + L('this is my key'));
    Ti.Locale.setLanguage('en-GB');
    Ti.API.info('new language is: ' + Ti.Locale.getCurrentLanguage());
    Ti.API.info('after change ' + L('this is my key'));
});
win.open();
```

i18n/en/strings.xml
```xml
<resources>
  <string name="this is my key">this is my value</string>
</resources>
```

i18n/en/app.xml
```xml
<resources>
    <string name="appname">app name en</string>
</resources>
```

i18n/en-GB/strings.xml
```xml
<resources>
  <string name="this is my key">this is my en-GB value</string>
</resources>
```

i18n/en-GB/app.xml
```xml
<resources>
    <string name="appname">app name en-GB</string>
</resources>
```


infosia:
Hmm, I don't know why `NMocha` doesn't handle i18n resources...basically I think `i18n` resources should have been included in Visual Studio project with specific path (`Strings/en/Resources.resw` etc), and `Ti.Locale.getString` just handles it accordingly based on current language. It does work when you manually create VS project (see `NG` project), Titanium CLI doesn't handle that accordingly now.

vs_project

infosia:
I'm still struggling to make this work. Basic idea here is to let CMake handle languages resources (see [Examples/NG/CMakeLists.txt](https://github.com/appcelerator/titanium_mobile_windows/pull/737/files#diff-62eaa993604ca92e37cb12c85bf05cfc). It works good on `NG` and `NMocha` examples so far, but when it comes to Titanium CLI it doesn't, `Ti.Locale.getString` doesn't work as expected even when app package seems to be correctly created. Any help would be appreciated! cc @sgtcoolguy @garymathews

sgtcoolguy:
build win 10

sgtcoolguy:
build win 8.1

Tools/Scripts/build/test.js
```diff
@@ -201,7 +201,7 @@ function addTiAppProperties(sdkVersion, next) {
 function copyMochaAssets(next) {
 	// Copy over Resources
 	wrench.copyDirSyncRecursive(MOCHA_ASSETS_DIR, path.join(projectDir, 'Resources'), {
-		forceDelete: true
+		forceDelete: true, exclude: /Resources.resw$/
 	});
 	// Copy over modules
```
sgtcoolguy: I think the issue here is that you need to add a line copying over the i18n files explicitly here, something along the lines of:
```javascript
// Copy over i18n
wrench.copyDirSyncRecursive(path.join(MOCHA_ASSETS_DIR, 'Strings'), path.join(projectDir, 'i18n'), {
forceDelete: true
});
```

infosia:
Thanks! I found it was a part of the reason why it didn't work 👍  Also, I think I got another reason why it doesn't work on emulator. Looks like it was because `ApplicationLanguages::PrimaryLanguageOverride` actually needs language packages installed on the emulator.  It needs language package downloaded/installed on the emulator, and until it doesn't `ApplicationLanguages::PrimaryLanguageOverride` takes no effect.

setting

PR #738, opened 2016-06-06
On Hold - [TIMOB-23405] Translate 8.1 client version

garymathews:
- Translate Windows 8.1 client version `6.3.-` into `8.1`
###### NOTES
- Do we want to append the build number to the end? e.g `8.1.9600`
- If we don't then we should remove the build number from Windows 10 `10.0.10586.0`
###### TEST CASE

``` Javascript
Ti.API.info(Ti.Platform.version);
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23405)

### DIFF
--- a/Source/Ti/src/Platform.cpp
+++ b/Source/Ti/src/Platform.cpp
@@ -16,6 +16,7 @@
 #include <collection.h>
 #include <regex>
 #include <concrt.h>
+#include <boost/algorithm/string.hpp>
 
 namespace TitaniumWindows
 {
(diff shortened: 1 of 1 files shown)


garymathews:
build win 10

infosia:
Can we have unit test for this?

infosia:
> Do we want to append the build number to the end? e.g 8.1.9600

IMO yes.

cc @sgtcoolguy

sgtcoolguy:
So first, I'm not even sure we should be doing this. I understand it's confusing that Windows reports the version as 6.3.x for Windows 8.1 - but that's what they do. I think we should surface the version as they report it, even though their "marketing" version differs.

But, if the consensus generally is to hack the version, then yeah we should keep the build number at the end. But, I'd argue we should get more feedback from people internally/externally about this before we go hacking the version string.

sgtcoolguy:
Closing outdated PR until we're sure we really want to mangle the version number we get from windows (even though it's not intuitive).

PR #739, opened 2016-06-06
[TIMOB-23399] Implement getter for Ti.Network.networkTypeName

garymathews:
- Implement ``Titanium.Network.networkTypeName`` property getter

###### TEST CASE
```Javascript
Ti.API.info(Ti.Network.networkTypeName);
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23399)

### DIFF
--- a/Examples/NMocha/src/Assets/ti.network.test.js
+++ b/Examples/NMocha/src/Assets/ti.network.test.js
@@ -35,4 +35,23 @@ describe('Titanium.Network', function () {
 		should(Ti.Network.createHTTPClient).be.a.Function;
 		finish();
 	});
+
+	it('networkType', function () {
+	    should(Ti.Network.networkType).not.be.null;
+	    should(Ti.Network.networkType).be.a.Number;
+	});
+
+	it('networkTypeName', function () {
+	    should(Ti.Network.networkTypeName).not.be.null;
+	    should(Ti.Network.networkTypeName).be.a.String;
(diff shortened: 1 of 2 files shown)


infosia:
Can we have unit test for this?

garymathews:
Yes, but I'm unsure how it would be a reliable test since the value changes depending on the connection type? In some cases ``UNKNOWN`` would be valid too.

infosia:
ok, how about something like this: check if it returns a string without any exception. better than nothing :)

```javascript
it('networkTypeName', function (finish) {
	should(Ti.Network.networkTypeName).not.be.null;
	should(Ti.Network.networkTypeName).be.a.String;
	finish();
});
```

[ti.network.test.js](https://github.com/garymathews/titanium_mobile_windows/blob/cf3a845f7599ab8736c9da56d413f70e87c94c76/Examples/NMocha/src/Assets/ti.network.test.js)

sgtcoolguy:
That could be shortened to:
```javascript
it('networkTypeName', function () {
    should(Ti.Network.networkTypeName).not.be.null;
    should(Ti.Network.networkTypeName).be.a.String;
});
```

Presumably you could do a comparison of the networkType int value and ensure the expected text for the int value was used...

sgtcoolguy:
build win 8.1

sgtcoolguy:
build win 10

PR #742, opened 2016-06-14
[5_4_X] [TIMOB-23125] Fix: Getting photo from photo gallery fails

infosia:
[TIMOB-23125](https://jira.appcelerator.org/browse/TIMOB-23125)

```javascript
var window = Ti.UI.createWindow({
    exitOnClose: true,
});
var viewParent = Ti.UI.createView({
    width: '60%',
    height: '41%',
    top: '1%'
});

var imageTaken = Ti.UI.createImageView({
    width: Ti.UI.FILL,
    height: Ti.UI.FILL,
});

var cameraImage = Ti.UI.createButton({
    title: "Choose Photos",
    bottom: '0%',
    height: '20%',
    font: {
        fontSize: 15,
        fontWeight: "bold"
    },
    width: Ti.UI.FILL,
});

viewParent.add(imageTaken);
viewParent.add(cameraImage);
window.add(viewParent);

cameraImage.addEventListener('click', function (e) {

    Ti.Media.openPhotoGallery({
        autoHide: true,
        mediaTypes: [Ti.Media.MEDIA_TYPE_PHOTO],
        success: function (e) {
            if (e.mediaType === Ti.Media.MEDIA_TYPE_PHOTO) {
                imageTaken.image = e.media;
            }
        },
        cancel: function () {
        },
        error: function (err) {
            Ti.API.error(err);
        }
    });

});
window.open();
```

### DIFF
--- a/Source/Media/src/Media.cpp
+++ b/Source/Media/src/Media.cpp
@@ -271,13 +271,13 @@ namespace TitaniumWindows
 		}
 
 		if (files.size() > 0) {
-			const auto blob = TitaniumWindows::Utility::GetTiBlobForFile(get_context(), files.at(0));
-			if (blob.IsObject()) {
+			const auto filename = files.at(0);
+			if (!filename.empty()) {
 				Titanium::Media::CameraMediaItemType item;
 				item.code = 0;
 				item.success = true;
 				item.mediaType = Titanium::Media::MediaType::Photo;
-				item.media = static_cast<JSObject>(blob).GetPrivate<Titanium::Blob>();
+				item.media_filename = filename;
 				openPhotoGalleryOptionsState__.callbacks.onsuccess(item);
 			} else {
 				Titanium::ErrorResponse error;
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
@infosia When testing this, I keep getting access denied when trying to access the photo from the Camera Roll directory of the photo gallery (with and without the fix). testing on Windows 10 Mobile emulator. I used our NG example app and modified the Package.appxmanifest to add the "picturesLibrary" capability. I'm guessing the photo chooser/picturesLibrary permissions stuff behaves differently than 8.1? Looks like we basically save the filename chosen and then try to grab the file later? I think maybe Windows 10 only gives you the handle to the file as the result of opening the gallery and may then deny trying to access that exact same file by path after the fact?

infosia:
Hmm...I can't reproduce it on my end for both Windows 10 Mobile and WP 8.1. Tested on both emulator and device. I believe Windows gives permission to access not only when opening gallery but also app process.

PR #758, opened 2016-06-29
[5_4_X] [TIMOB-23560] Fix: Setting images on ImageView will crash the app

infosia:
[TIMOB-23560](https://jira.appcelerator.org/browse/TIMOB-23560)

```javascript
var win = Ti.UI.createWindow({backgroundColor:'green'});

var images = [];
for (var i = 0; i < 31; i++) {
    images.push('Roamler_app_loader_' + ( '0' + i ).slice( -2 ) + '.png');
}

var image = Ti.UI.createImageView({
    images: images,
    width: 75,
    height: 75,
    duration: 67
});

    image.start();
});

win.add(image);
win.open();
```

Images: [loader.zip](https://jira.appcelerator.org/secure/attachment/59655/loader.zip)

### DIFF
--- a/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
+++ b/Examples/NMocha/src/Assets/ti.ui.imageview.test.js
@@ -8,7 +8,7 @@ var should = require('./should'),
 	utilities = require('./utilities/utilities');
 
 describe('Titanium.UI.ImageView', function () {
-	this.timeout(5000);
+	this.timeout(10000);
 
 	it('apiName', function (finish) {
 		should(Ti.UI.ImageView.apiName).be.eql('Ti.UI.ImageView');
@@ -212,7 +212,6 @@ describe('Titanium.UI.ImageView', function () {
 			height: Ti.UI.SIZE
 		});
 		var innerView = Ti.UI.createImageView({
-			image: 'http://api.randomuser.me/portraits/women/0.jpg',
 			width: 100,
 			height: Ti.UI.SIZE,
 			top: 0,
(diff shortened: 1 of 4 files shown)


Source/UI/src/ImageView.cpp
```diff
@@ -234,23 +234,11 @@ namespace TitaniumWindows
...
-					if (bitmaps_loaded_count__ >= images_count) {
-						bitmaps_loaded__ = true;
-						fireEvent("load");
-						// start animation when we held it off
-						if (bitmaps_waiting__) {
-							start();
-						}
-					}
-				});
+				bitmaps__->Append(ref new BitmapImage(TitaniumWindows::Utility::GetUriFromPath(images.at(i))));
```
infosia: Note: it turns out we just can use `BitmapImage(Uri)` here to load images from Uri, and it's much lightweight and safer.

Examples/NMocha/src/Assets/ti.ui.imageview.test.js
```diff
@@ -228,6 +227,7 @@ describe('Titanium.UI.ImageView', function () {
 				finish();
 			}, 1000);
 		});
+	    innerView.image = 'http://api.randomuser.me/portraits/women/0.jpg';
```
garymathews: formatting?

garymathews:
CR: **PASS**
FR: **PASS**

NOTE: That the sequence may flicker if the ``duration`` is set too high, this is a platform issue see: [Flickering Issue with Image Animations](https://social.msdn.microsoft.com/Forums/windows/en-US/1b3f556c-07e2-40e4-b895-1d78bf829a13/flickering-issue-with-image-animations?forum=wpdevelop)

PR #761, opened 2016-06-29
[TIMOB-23156] Generate capabilities from API usage

garymathews:
- Generate ``Package.appxmanifest`` ``Capabilities`` and ``DeviceCapabilities`` from API usage
- Only generate capabilities when none are specified in ``tiapp.xml``

|               API               |
|:-------------------------------:|
| Titanium.Contacts.* |
| Titanium.Filesystem.externalStorageDirectory |
| Titanium.Geolocation.* |
| Titanium.Media.openPhotoGallery |
| Titanium.Media.startVideoCapture |
| Titanium.Media.takeScreenshotToFile |
| Titanium.Media.showCamera |
| Titanium.Media.AudioRecorder.* |
| Titanium.Network.* |

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'black' }),
    lbl = Ti.UI.createLabel({ top: '5%', left: 5, color: 'white' }),
    btn_a = Ti.UI.createButton({ top: '25%',color: 'white', title: 'showCamera()' }),
    btn_b = Ti.UI.createButton({ top: '35%', color: 'white', title: 'openPhotoGallery()' });

Ti.Geolocation.getCurrentPosition(function (data) {
    if (data.success) {
        lbl.text += 'getCurrentPosition()' + data.coords.latitude + ',' + data.coords.longitude;
    }
});
btn_a.addEventListener('click', function () {
    Titanium.Media.showCamera({});
});
btn_b.addEventListener('click', function () {
    Titanium.Media.openPhotoGallery({mediaTypes: [Ti.Media.MEDIA_TYPE_PHOTO]});
});

win.add(lbl);
win.add(btn_a);
win.add(btn_b);
win.open();
```
```
appc run -p windows --build-only
```
```XML
<Capability Name="picturesLibrary" />
<DeviceCapability Name="location" />
<DeviceCapability Name="microphone" />
<DeviceCapability Name="webcam" />
```
[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23156)

### DIFF
--- a/cli/commands/_build/generate.js
+++ b/cli/commands/_build/generate.js
@@ -24,6 +24,7 @@ function mixin(WindowsBuilder) {
 	WindowsBuilder.prototype.generateNativeWrappers = generateNativeWrappers;
 	WindowsBuilder.prototype.generateModuleFinder = generateModuleFinder;
 	WindowsBuilder.prototype.generateCmakeList = generateCmakeList;
+	WindowsBuilder.prototype.generateCapabilityList = generateCapabilityList;
 	WindowsBuilder.prototype.generateAppxManifestForPlatform = generateAppxManifestForPlatform;
 	WindowsBuilder.prototype.generateAppxManifest = generateAppxManifest;
 }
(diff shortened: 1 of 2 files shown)


infosia:
I got this when I put empty no `<windows>` tag in `tiapp.xml`, and then building with `appc ti build -p windows --wp-sdk 10.0 --target wp-device`.

```
[ERROR] .titanium\vsbuild\win10_classic\win10.ARM\Debug\AppxManifest.xml : 
error APPX0501: Validation error. error C00CE169: 
App manifest validation error: The app manifest must be valid as per schema:
 Line 35, Column 17, Reason: 'picturesLibrary' violates enumeration constraint of 'internetClient internetClientServer privateNetworkClientServer allJoyn codeGeneration'. 
The attribute 'Name' with value 'picturesLibrary' failed to parse.
 [.titanium\vsbuild\win10_classic\win10.ARM\Win10Classic.vcxproj]
```

That is because `generateAppxManifestForPlatform` skips `uap:Capability` conversion for the generated tags.

### tiapp.xml

```xml
<!--
  <windows>
    <manifest>
      <Capabilities>
        <Capability Name="internetClient"/>
          <Capability Name="picturesLibrary"/>
        </Capabilities>
    </manifest>
  </windows>
-->
```

garymathews:
But we shouldn't need to worry about that right?

https://github.com/appcelerator/titanium_mobile_windows/blob/master/templates/build/Package.win10.appxmanifest.in.ejs#L7

garymathews:
Updated to handle ``uap:Capability`` permissions

cli/commands/_build/generate.js
```diff
@@ -328,7 +407,7 @@ function generateAppxManifestForPlatform(target, properties) {
 		});
 		properties.Capabilities = capabilities.concat(deviceCapabilities);
 	} else {
-		properties.Capabilities = ['<Capability Name=\"internetClient\" />'];
+		this.generateCapabilityList(target, properties);
 	}
```
infosia: When there's `windows/manifest` entry in tiapp.xml, `generateCapabilityList` never be executed.

```xml
  <windows>
    <manifest>
      <Capabilities>
        <Capability Name="internetClient"/>
          <Capability Name="picturesLibrary"/>
        </Capabilities>
    </manifest>
  </windows>
```

I think we should always do `generateCapabilityList` here so we can check properties even when `properties.Capabilities` exists.

```diff
if (properties.Capabilities) {
...
    properties.Capabilities = capabilities.concat(deviceCapabilities); 
- } else {
-		properties.Capabilities = ['<Capability Name=\"internetClient\" />']; 
}
+ this.generateCapabilityList(target, properties); 
```

infosia:
Following up just in case, the reason because we need to add `uap:` explicitly even when we have ` IgnorableNamespaces="uap mp"` is that we want to avoid issue when you have entries that has both `Capability` and `uap:Capability` tags like below.

```xml
<Capability Name="internetClient"/>
<Capability Name="picturesLibrary"/>
```

In this case we should end up with adding `uap:` to `picturesLibrary` because without it it is ambiguous and then causes `error APPX0501: Validation error. error C00CE169`.

```xml
<Capability Name="internetClient"/>
<uap:Capability Name="picturesLibrary"/>
```

cli/commands/_build/generate.js
garymathews: @infosia  ``generateCapabilityList`` would need to be changed to include any existing properties (I think this is the reason why the Jenkins server is failing). The reason the capabilities are only generated when none are defined in the ``tiapp.xml`` is so users have the ability to define their own.

infosia: ok. I would prefer always generate the required capabilities and merge them into existing ones. Could you push that soon? Thanks.

PR #763, opened 2016-06-30
Skip forward/reverse geocoder for Windows 10 Store

infosia:
Currently I see geocoder is unstable for Windows 10 Store. ([titanium_mobile_windows_10_prs/786](http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_10_prs/786/))
Skipping it for now so no one blocked by that.

### DIFF
--- a/Examples/NMocha/src/Assets/ti.geolocation.test.js
+++ b/Examples/NMocha/src/Assets/ti.geolocation.test.js
@@ -156,7 +156,7 @@ describe('Titanium.Geolocation', function () {
 		finish();
 	});
 
-	it('forwardGeocoder', function (finish) {
+	((utilities.isWindows10() && utilities.isWindowsDesktop()) ? it.skip : it)('forwardGeocoder', function (finish) {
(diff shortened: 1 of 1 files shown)


infosia:
@sgtcoolguy Do you think this is temporary issue (network or something?), or because Win10 Store doesn't support it?

garymathews:
Windows 10 Store should support it, it's only a HTTP request. I'm unsure why the tests time out?

sgtcoolguy:
The error message gives no details, unfortunately. I'm guessing it's an intermittent failure due to a timeout. I desperately need to fold in the current test suite changes I made for iOS/Android because it'll capture stderr/stdout from test runs and give us a better idea.

Given that the test took exactly 2 seconds and failed, it's probably just a timeout. The best course would be to set a longer timeout for a network test like this (i.e. 30 seconds) - not skip it.

infosia:
I see, 2 seconds sounds too short, will update it fails again. Closing this.

PR #750, opened 2016-06-24
[TIMOB-23512] [TIMOB-23543] Relative require() from sub-directory fails

infosia:
[TIMOB-23512](https://jira.appcelerator.org/browse/TIMOB-23512) and [TIMOB-23543](https://jira.appcelerator.org/browse/TIMOB-23543)

### app.js
```javascript
var test = require('subdir/index');
should(test.filename).eql('/subdir/index.js');
should(test.name).eql('index.js');
should(test.sub).eql('sub.js');
```

### subdir/index.js

```javascript
var sub = require('./sub')
module.exports = {
    filename: __filename,
    name: 'index.js',
    sub: sub.name
};
```

### subdir/sub.js

```javascript
module.exports = {
    name: 'sub.js',
};
```

Note: In  order to retrieve current file name from requiring object, I ended up adding new hidden variable `__TI_REQUIRED__` in global object, which is only available while evaluating required module. This is tricky but needed because `JSObject parent` in `GlobalObject::requestResolveModule(parent, moduleId)` is always global object. (See also [Source/TitaniumKit/src/GlobalObject.cpp ](https://github.com/appcelerator/titanium_mobile_windows/pull/750/files#diff-3b6d0c60899ff30352e5cc386b90fa56R270))

PR #764, opened 2016-06-30
Fix another broken require case where multiple relative requires in same file failed.

sgtcoolguy:
Follow-on to #750 

The common mocha suite was still broken because it had two relative requires in the same file. The use of the __TI_REQUIRED__ hack would only work for a single require because the property lived off the global, so after the first require it got removed and we lost state. It needed to be able to be dealt with recursively. I instead move it to a currentDir__ field on GlobalObject in the C++ code, mimicking what we do for iOS.

Note that this is a hack that works for now, but a real proper Node-style require implementation requires more than what we're doing now. They generate a Module object after the path is resolved holding the resolved filename and a pointer to the parent Module. They then generate a require function that is hooked within the context of that module. When they load the module from source, they wrap it in a way that they can pass in more arguments than we currently provide: filename, dirname, parent, the custom require for that context.

Basically it allows them to recursively do requires where you take the parent context to know how to resolve. Their cache also holds module objects and returns back module.exports from require calls (where we just hold the exports object itself in the cache).

We should consider fixing our implementation to be closer to theirs. Instead of doing that magic eval string we do, we need to generate a function we can call with actual args and pass in the various pieces. Our Android implementation follows much closer to theirs.

Most of the magic for Node is here: 
- https://github.com/nodejs/node/blob/master/lib/module.js
- https://github.com/nodejs/node/blob/master/lib/internal/module.js
- https://github.com/nodejs/node/blob/1a21524b6908892e2d0a79cbb4241aee78530c09/lib/internal/bootstrap_node.js

See also: https://nodejs.org/dist/latest-v6.x/docs/api/modules.html#modules_the_module_wrapper

### DIFF
--- a/Examples/NMocha/src/Assets/ti.require.test.js
+++ b/Examples/NMocha/src/Assets/ti.require.test.js
@@ -139,14 +139,16 @@ describe('requireJS', function () {
 		should(with_index_js.name).be.eql('index.js');
 		finish();
 	});
-    
-    // TIMOB-23512
+
+	// TIMOB-23512
 	it('relative require() from sub directory', function (finish) {
-	    var with_index_js = require('./with_index_js/sub1');
-	    should(with_index_js).have.property('name');
-	    should(with_index_js.name).be.eql('sub1.js');
-	    should(with_index_js.sub).be.eql('sub2.js');
-	    finish();
+		var with_index_js = require('./with_index_js/sub1');
+		should(with_index_js).have.property('name');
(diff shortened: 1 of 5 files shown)


sgtcoolguy:
build win 10

sgtcoolguy:
@infosia ready for review/merge. A single UI test failed on Windows 10, looks like a random timeout - shouldn't be related to this PR.

sgtcoolguy:
blocks #733

PR #766, opened 2016-07-04
[TIMOB-23607] Fix Ti.UI.animate() for right and bottom properties

garymathews:
- Fix ``Titanium.UI.animate()`` where ``right`` and ``bottom`` properties would not be retained after the animation

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'red' }),
    view = Ti.UI.createView({
        backgroundColor: 'orange',
        top: 0,
        left: 0,
        height: '100',
        width: '100'
    });

view.addEventListener('click', function () {
    view.animate(Ti.UI.createAnimation({
        right: '50%',
        bottom: '50%',
        duration: 2000
    }));
});

win.add(view);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23607)

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -732,20 +732,20 @@ namespace TitaniumWindows
 				// Make sure to update the position of the view because StoryBoard doesn't change actual position of it.
 				//
 				const auto top = animation->get_top();
+				const auto bottom = animation->get_bottom();
 				if (top) {
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, *top);
-				}
-				const auto bottom = animation->get_bottom();
-				if (bottom) {
+				} else if (bottom) {
+					setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, "NONE");
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Bottom, *bottom);
 				}
 
 				const auto left = animation->get_left();
+				const auto right = animation->get_right();
 				if (left) {
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Left, *left);
-				}
-				const auto right = animation->get_right();
-				if (right) {
+				} else if (right) {
+					setLayoutProperty(Titanium::LayoutEngine::ValueName::Left, "NONE");
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Right, *right);
 				}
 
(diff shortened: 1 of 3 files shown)


infosia:
@garymathews Can you add unit test for this?

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -732,20 +732,20 @@ namespace TitaniumWindows
 				// Make sure to update the position of the view because StoryBoard doesn't change actual position of it.
 				//
 				const auto top = animation->get_top();
+				const auto bottom = animation->get_bottom();
 				if (top) {
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, *top);
-				}
```
infosia: I'm wondering why we don't update the bottom here?

```c++
if (top) {
    setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, *top);
    setLayoutProperty(Titanium::LayoutEngine::ValueName::Bottom, "NONE");
}
```

Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -732,20 +732,20 @@ namespace TitaniumWindows
...
+				} else if (bottom) {
+					setLayoutProperty(Titanium::LayoutEngine::ValueName::Top, "NONE");
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Bottom, *bottom);
 				}
 
 				const auto left = animation->get_left();
+				const auto right = animation->get_right();
 				if (left) {
 					setLayoutProperty(Titanium::LayoutEngine::ValueName::Left, *left);
-				}
```
infosia: Don't we need to update the `Right` too?

Source/UI/src/WindowsViewLayoutDelegate.cpp
garymathews: We could, but the ``LayoutEngine`` takes care of that; so it's not really needed.

https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/LayoutEngine/src/Composite.cpp#L384

- ``left`` has priority over ``right`` and ``top`` has priority over ``bottom``

###### TEST CASE - animate from right, bottom to left, top
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'red' }),
    view = Ti.UI.createView({
        backgroundColor: 'orange',
        right: '50%',
        bottom: '50%',
        height: '100',
        width: '100'
    });

view.addEventListener('click', function () {
    view.animate(Ti.UI.createAnimation({
        top: 0,
        left: 0,
        duration: 2000
    }));
});

win.add(view);
win.open();
```

infosia:
Looks good. ok to merge once unit tests pass.

garymathews:
build win 8.1

PR #770, opened 2016-07-06
[TIMOB-23563] Suppress 'possible EventEmitter memory leak detected'

infosia:
[TIMOB-23563](https://jira.appcelerator.org/browse/TIMOB-23563)

Suppress `possible EventEmitter memory leak detected` warnings from on packaging module by using [EventEmitter.defaultMaxListeners](https://nodejs.org/api/events.html#events_eventemitter_defaultmaxlisteners). I assume this is safe because these warnings are from Node.js Streams and they properly removes listeners on finish.


garymathews:
FR: **PASS**
CR: **PASS**

PR #779, opened 2016-07-15
TitaniumKit performance optimization

infosia:
The [single context mode](https://github.com/appcelerator/HAL/pull/63) and [retain count management](https://github.com/appcelerator/HAL/pull/64) I introduced in the latest HAL provides huge performance improvement in TitaniumKit. I  have observed it cuts down over ~~50%~~ 70% of time.

On Lumia 630 (Windows Phone 8.1) using Ti SDK 6.0.0.v20160714010528

- Ti SDK 6.0.0.v20160714010528 ... 5.5 sec
- Ti SDK 6.0.0.v20160714010528 with this update ... ~~2.7 sec~~ -> 1.7 sec

```javascript
var win = Ti.UI.createWindow({ backgroundColor: 'green' });
var data = [];

var start = +new Date();
for (var i = 0; i < 100; i++) {
    var task = {
        Id: i,
        Title: 'Row ' + i,
        Data: 'Data ' + i
    };
    var tvr = Ti.UI.createTableViewRow({
        height: 70,
        taskId: task.Id
    });
    tvr.add(Ti.UI.createLabel({
        text: task.Title,
        color: "#000",
        top: 5,
        left: 20
    }));
    tvr.add(Ti.UI.createLabel({
        text: task.Data,
        color: "#000",
        top: 30,
        left: 20,
        font: {
            fontSize: 12
        }
    }));
    data.push(tvr);
}

var table = Ti.UI.createTableView({data: data});

table.addEventListener('click', function (e) {
    alert(e.row.taskId)
});

alert('Total: ' + (+new Date() - start));

win.add(table);
win.open();
```

I have also added performance improvement for TableView.

PR #783, opened 2016-07-19
FIXME: Skip some tests on Windows 10 desktop

infosia:
According to the latest mocha tests, some tests on Windows 10 desktop start failing. We might want to just skip it for now because we've been seeing similar cases already before. Need to revisit latest on.


infosia:
This PR may be obsolete. Will wait for a while, as just wanted to ensure if restarting server fixes this. See https://github.com/appcelerator/titanium_mobile_windows/pull/779#issuecomment-233799440

PR #794, opened 2016-07-29
[TIMOB-23696] Fix Ti.App._restart() functionality

garymathews:
- Remove `_restart()` from `TitaniumWindows_Ti`
- Implement `Ti.App._restart()` in `TitaniumKit`

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow({ backgroundColor: 'blue' }),
	btn = Ti.UI.createButton({ title: 'Ti.App._restart()' });

setTimeout(function () {
    win.backgroundColor = 'red';
}, 2000);

    Ti.App._restart();
});

win.add(btn);
win.open();
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23696)

### DIFF
--- a/Source/TitaniumKit/src/App.cpp
+++ b/Source/TitaniumKit/src/App.cpp
@@ -257,7 +257,7 @@ namespace Titanium
 
 	void AppModule::_restart() TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_WARN("AppModule::_restart: Unimplemented");
+		get_context().JSEvaluateScript("eval(Ti.Filesystem.getFile('app.js').read().text)");
 	}
 
 	AppModule& AppModule::PropertiesClass(const JSClass& Properties) TITANIUM_NOEXCEPT
--- a/Source/Ti/include/TitaniumWindows/AppModule.hpp
+++ b/Source/Ti/include/TitaniumWindows/AppModule.hpp
@@ -55,8 +55,6 @@ namespace TitaniumWindows
 		virtual void setProximityDetection(const bool&) TITANIUM_NOEXCEPT override;
 		virtual bool proximityState() const TITANIUM_NOEXCEPT override;
 
-		virtual void _restart() TITANIUM_NOEXCEPT override;
-
 	private:
 
 		static JSObject RectToJS(const JSContext& js_context, const Windows::Foundation::Rect& rect);
--- a/Source/Ti/src/AppModule.cpp
+++ b/Source/Ti/src/AppModule.cpp
@@ -154,9 +154,4 @@ namespace TitaniumWindows
 		return keyboardVisible__;
 	}
 
-	void AppModule::_restart() TITANIUM_NOEXCEPT
-	{
-		Windows::ApplicationModel::Core::CoreApplication::Exit();
-	}
-
 }  // namespace TitaniumWindows


Source/TitaniumKit/src/App.cpp
```diff
@@ -257,7 +257,7 @@ namespace Titanium
 
 	void AppModule::_restart() TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_WARN("AppModule::_restart: Unimplemented");
+		get_context().JSEvaluateScript("require('app.js')");
```
garymathews: Maybe unload any previously loaded components?

infosia: I'm guessing this is not enough. We might want to re-create JS context, unload all existing UI components and un-registers from LayoutEngine. Hopefully want to restart the app by calling `Windows::UI::Xaml::Application::Start` again. I'll look into whether this is even possible.

infosia:
After carefully auditing our designs,  I've almost came to a conclusion that our `HAL` and `TitaniumKit` framework doesn't really support "restart" situation unfortunately, except for the situation where the application is completely restarted. We need to completely clean up the application state when it's restarted, but I can see that we are now using `static` variables in our framework. In order to clean the app state, not only delete all instances but all of the static variables needs to be cleaned up, otherwise application state will be messed up.

For instance,

- Constants cache in HAL- [JSExportClass.hpp#L136](https://github.com/appcelerator/HAL/blob/624d60a4e1ae362bb15626e0d6fa5723998ff5b5/include/HAL/detail/JSExportClass.hpp#L136)
- JSObject pointer mapping in HAL- [JSObject.hpp#L449](https://github.com/appcelerator/HAL/blob/624d60a4e1ae362bb15626e0d6fa5723998ff5b5/include/HAL/JSObject.hpp#L449)
- Window stack in TitaniumWindows - [Window.hpp#L108](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/UI/include/TitaniumWindows/UI/Window.hpp#L108)

infosia:
I think I came up with an idea. Let me try.

Source/TitaniumKit/src/App.cpp
```diff
@@ -257,7 +257,7 @@ namespace Titanium
 
 	void AppModule::_restart() TITANIUM_NOEXCEPT
 	{
-		TITANIUM_LOG_WARN("AppModule::_restart: Unimplemented");
+		get_context().JSEvaluateScript("eval(Ti.Filesystem.getFile('app.js').read().text)");
```
infosia: @garymathews, Is there any reason you didn't call `require('app')` but reading text content here? I assume this does not work on `Release` build because `app.js` is encrypted.

garymathews: @infosia Yes, ``require('app.js')`` is cached [GlobalObject.cpp#L296](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/src/GlobalObject.cpp#L296) and would not reload ``app.js``

infosia: Makes sense!

infosia:
Closing as taken over by #796

PR #804, opened 2016-08-04
[TIMOB-23723] Top and bottom, left and right should not override a set width and height

garymathews:
- When `top`, `bottom` and `height` or `left`, `right` and `width` are set. The `height` or `width` should have priority

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow({backgroundColor: 'purple'}),
    img = Ti.UI.createImageView({
        image: 'Logo.png',
        backgroundColor: 'red',
        left: 0,
        top: 0,
        bottom: 0,
        width: Ti.UI.SIZE,
        height: Ti.UI.SIZE
    });
win.add(img);
win.open();
```
```
Image should be visible
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23723)


infosia:
👍

PR #791, opened 2016-07-27
[TIMOB-23678] Fix Ti.Network.HTTPClient.send() concurrency

garymathews:
- Fixed ``RequestState`` enum order
- ``Titanium.Network.HTTPClient.send()`` is now fully asynchronous

###### TEST CASE
```Javascript
var win = Ti.UI.createWindow(),
    btn = Ti.UI.createButton({
        title: 'HTTP GET'
    });

btn.addEventListener('click', function () {
    var request = Ti.Network.createHTTPClient({
            onload: function () {
                alert('onload:\n\treadyState: ' + request.readyState);
            },
            onerror: function (e) {
                alert('onerror:\n\treadyState: ' + request.readyState + '\n\terror: ' + e.error);
            }
        }),
        timeout = new Date().getTime() + 5000,
        host = 'http://www.google.com/',
        done = false;

    request.cache = false;
    request.open('GET', host);
    request.send();

    while (!done) {
        if (request.readyState === 4 || request.status === 404) {
            done = true;
            alert('success');
        } else if ((timeout - (new Date()).getTime()) <= 0) {
            done = true;
            alert('failed');
        }
    }
});

win.add(btn);
win.open();
```
```
onload:
    readyState: 4
```
```
success
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23678)

PR #810, opened 2016-08-10
[TIMOB-23678] Fix Ti.Network.HTTPClient.send() concurrency

infosia:
#791 + Workaround for Windows 10 desktop. We've been seeing intermittent crash on Windows 10 desktop with async HTTPClient.


infosia:
ok, merging this for now, as this works good on mobile environments. We should revisit Windows 10 desktop. I'll create new ticket for it.

PR #818, opened 2016-08-12
[TIMOB-23760] --wp-sdk option should handle actual Windows SDK versions

infosia:
Note: ~~This needs [windowslib 0.4.17](https://github.com/appcelerator/windowslib/pull/52) merged for [titanium_mobile](https://github.com/appcelerator/titanium_mobile)~~ -> *Merged*.

[TIMOB-23760](https://jira.appcelerator.org/browse/TIMOB-23760)
[TIMOB-23758](https://jira.appcelerator.org/browse/TIMOB-23758)
[TIMOB-23759](https://jira.appcelerator.org/browse/TIMOB-23759)

```
appc run -p windows --wp-sdk 10.0.10586.0 --target wp-device -l trace
```

PR #819, opened 2016-08-12
[TIMOB-23759] Add ability to set Windows 10 TargetPlatformVersion

garymathews:
- Implement ability to set `TargetPlatformVersion` and `TargetPlatformMinVersion` from `tiapp.xml`
- Ability to select a target SDK with `--wp-sdk`

###### tiapp.xml
```
...
<windows>
    ...
    <TargetPlatformVersion>10.0.10586.0</TargetPlatformVersion>
    <TargetPlatformMinVersion>10.0.10240.0</TargetPlatformMinVersion>
    ...
</windows>
...
```
```
appc run -p windows -T wp-device -S 10.0.10586.0 --build-only
```

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-23759)

### DIFF
--- a/cli/commands/_build/generate.js
+++ b/cli/commands/_build/generate.js
@@ -28,6 +28,7 @@ function mixin(WindowsBuilder) {
 	WindowsBuilder.prototype.generateAppxManifestForPlatform = generateAppxManifestForPlatform;
 	WindowsBuilder.prototype.generateAppxManifest = generateAppxManifest;
 	WindowsBuilder.prototype.addI18nVSResources = addI18nVSResources;
+	WindowsBuilder.prototype.setTargetPlatformVersion = setTargetPlatformVersion;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


cli/commands/_build/generate.js
```diff
@@ -115,6 +116,31 @@ function addI18nVSResources(next) {
...
+		var projectName = this.sanitizeProjectName(this.cli.tiapp.name),
+			projectPath = path.resolve(this.cmakeTargetDir, projectName + '.vcxproj'),
+			project = fs.readFileSync(projectPath, 'utf8');
+
+		this.logger.info(__('Setting TargetPlatformVersion: %s', target.cyan));
+		this.logger.info(__('Setting TargetPlatformMinVersion: %s', min.cyan));
+
+		fs.existsSync(projectPath) && fs.renameSync(projectPath, projectPath + '.bak');
+		project = project.replace(/<WindowsTargetPlatformVersion>(10\.0)(\.\d+)*<\/WindowsTargetPlatformVersion>/, '<WindowsTargetPlatformVersion>' + target + '</WindowsTargetPlatformVersion>');
+		project = project.replace(/<WindowsTargetPlatformMinVersion>(10\.0)(\.\d+)*<\/WindowsTargetPlatformMinVersion>/, '<WindowsTargetPlatformMinVersion>' + min + '</WindowsTargetPlatformMinVersion>');
```
infosia: Yes this is a simplest way to do this, but I think we want to let cmake to generate this because we don't want to maintain its internal XML representation when possible. For the target version we can use `CMAKE_SYSTEM_VERSION` or [CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION](https://cmake.org/cmake/help/v3.4/variable/CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.html#variable:CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION), and for the min version we can use [VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION](https://cmake.org/cmake/help/v3.4/prop_tgt/VS_WINDOWS_TARGET_PLATFORM_MIN_VERSION.html). I'll amend this in my PR #818.

infosia:
Will amend this in PR #818.

PR #836, opened 2016-08-25
[6_0_X] Skip reverse/forward geocoder on desktop 

infosia:
Skip reverse/forward geocoder on desktop because of intermittent network issue


infosia:
So basically we've been seeing intermittent crash/freeze on Jenkins on Windows desktop. we've been trying to fix it for a while but had no luck, so I would skip it for now because it blocks our nightly build.

PR #846, opened 2016-08-31
[TIMOB-23752] Async HTTPClient crashes on Win10 Desktop

infosia:
[TIMOB-23752](https://jira.appcelerator.org/browse/TIMOB-23752)

Async HTTPClient introduced crash at NMocha test on Windows 10 Desktop. Reducing temporary memory for JavaScriptCore seems to be fixing this.


garymathews:
Nice! :+1:

PR #856, opened 2016-09-07
Workaround to make nightly build pass

infosia:
Cherry-pick #855 for master

### DIFF
--- a/Examples/NMocha/src/Assets/app.js
+++ b/Examples/NMocha/src/Assets/app.js
@@ -33,6 +33,10 @@ require('./ti.contacts.person.test');
 require('./ti.database.test');
 require('./ti.filestream.test');
 require('./ti.filesystem.test');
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
Ideally we shouldn't be skipping whole suites of tests to avoid crashes - we should be fixing the root cause.

infosia:
Yes, root cause is described in comments at [TIMOB-23776](https://jira.appcelerator.org/browse/TIMOB-23776). Basically it's because of missing  [GetThreadContext](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679362(v=vs.85).aspx) call in [JSC::MachineThreads::Thread::getRegisters](https://github.com/appcelerator/webkit-private/commit/e0fc3fd783646f8abcf950bd77a32da805f89743#diff-627ab4f57b17be3f362eb6906157f78bL418) and the problem is it's not available on Windows 8.1 Store app.

PR #872, opened 2016-09-30
[TIMOB-23965] Add eventName after feature event validation

infosia:
[TIMOB-23965](https://jira.appcelerator.org/browse/TIMOB-23965)

Exclude `eventName` property from feature event validation.


garymathews:
CR: **PASS**
FR: **PASS**

👍

PR #880, opened 2016-10-13
[TIMOB-23998] Implement TableView scroll/scrollend event

infosia:
[TIMOB-23998](https://jira.appcelerator.org/browse/TIMOB-23998)

Note that some callback properties are not available on Windows due to platform component spec.

- `scroll` callback supports `type`, `source`, `size` and `totalItemCount` property
- `scrollend` callback supports `type`, `source`  and `size` property

```javascript
var tableData = [];
var data = [{ title: 'Apples' }, { title: 'Bananas' }, { title: 'Carrots' }, { title: 'Potatoes' }];

for (var i = 0; i < 5; i++) {
    for (var j = 0; j < data.length; j++) {
        tableData.push(data[j]);
    }
}

    data: tableData
});

table.addEventListener('scroll', function (e) {
    Ti.API.info('---- ' + e.type + '----');
    Ti.API.info(JSON.stringify(e.size));
    Ti.API.info(JSON.stringify(e.totalItemCount));
});

table.addEventListener('scrollend', function (e) {
    Ti.API.info('---- ' + e.type + '----');
    Ti.API.info(JSON.stringify(e.size));
});


win.add(table);
win.open();
```

### DIFF
--- a/Source/TitaniumKit/include/Titanium/Module.hpp
+++ b/Source/TitaniumKit/include/Titanium/Module.hpp
@@ -164,6 +164,8 @@ namespace Titanium
 		TITANIUM_FUNCTION_DEF(applyProperties);
 		TITANIUM_FUNCTION_DEF(fireEvent);
 
+		virtual bool hasEventListener(const std::string& event_name) TITANIUM_NOEXCEPT;
+
 		/*
 		* Stop firing all events, especially used when module is closed/hidden.
 		*/
--- a/Source/TitaniumKit/src/Module.cpp
+++ b/Source/TitaniumKit/src/Module.cpp
@@ -148,6 +148,11 @@ namespace Titanium
 		}
 	}
 
+	bool Module::hasEventListener(const std::string& event_name) TITANIUM_NOEXCEPT
+	{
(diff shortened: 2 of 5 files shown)


infosia:
build win 10

garymathews:
When I scroll to the very end of the list (top or bottom), the `scrollend` event is received twice. Scrolling through the middle of the list is fine, it's only when I reach either end stop of the list.

infosia:
Looks like `scrollend` is fired twice because `ViewChanged` is fired right after `ViewChainging` (`scroll`) event too. We need to do something to prevent firing after `scroll` event.

infosia:
On the desktop, interestingly `scrollend` event is fired twice only when you drag the scroll bar to the end.  When you click the "arrow" button (up/down), it's working fine. On the phone deivce, it is always fired twice when you scroll down to the end of the list. This is interesting, but I would think this is a platform component spec and I didn't come up with good idea to prevent it.

infosia:
ok I was able to prevent it fired twice by checking previous ScrollView position. Pushed a fix for it.

garymathews:
CR: **PASS**
FR: **PASS**

Looks good! 👍

PR #881, opened 2016-10-17
[6_0_X] [TIMOB-24019] ListView should keep "original" data even when filtered

infosia:
[TIMOB-24019](https://jira.appcelerator.org/browse/TIMOB-24019)

- `ListView` `sections` and `sectionCount` should keep its original data even when it's filtered
- `ListView` `itemclick` event should return "original" position even when it's filtered

For instance when you search "p" on following app and then click on the second item (`Potatoes`), it should return its original position (`sectionIndex=1`, `itemIndex=1`).

```javascript
var win = Ti.UI.createWindow();
var win = Ti.UI.createWindow({ backgroundColor: 'green', layout: 'vertical' });
var searchText = Ti.UI.createTextField({ width: Ti.UI.FILL, height: '10%' });

var listView = Ti.UI.createListView({ width: Ti.UI.FILL, height: '90%', caseInsensitiveSearch: true });
var sections = [];

var fruitSection = Ti.UI.createListSection({ headerTitle: 'Fruits' });
var fruitDataSet = [
    { properties: { title: 'Apple', searchableText: 'Apple' } },
    { properties: { title: 'Banana', searchableText: 'Banana' } },
];
fruitSection.setItems(fruitDataSet);
sections.push(fruitSection);

var vegSection = Ti.UI.createListSection({ headerTitle: 'Vegetables' });
var vegDataSet = [
    { properties: { title: 'Carrots', searchableText: 'Carrots' } },
    { properties: { title: 'Potatoes', searchableText: 'Potatoes' } },
];
vegSection.setItems(vegDataSet);
sections.push(vegSection);

listView.sections = sections;

searchText.addEventListener('change', function () {
    listView.searchText = searchText.value;
});

listView.addEventListener('itemclick', function (evt) {
    var clickItemIndex = evt.itemIndex;
    var clickSectionIndex = evt.sectionIndex;

    alert('sectionCount=' + listView.sectionCount + ' section=' + clickSectionIndex + ' item=' + clickItemIndex + '\r\n' + listView.sections[clickSectionIndex].items[clickItemIndex].properties.title);
});

win.add(searchText);
win.add(listView);
win.open();
```

PR #882, opened 2016-10-18
[TIMOB-24019] ListView should keep "original" data even when filtered

infosia:
Cherry-pick #881 for master


infosia:
build win 10

garymathews:
build win 10

PR #888, opened 2016-10-27
[TIMOB-24093] Hyperloop for Windows (Reflection based)

infosia:
Ready for review.

[TIMOB-24093](https://jira.appcelerator.org/browse/TIMOB-24093)

Hyperloop for Windows. Basically removes previous Hyperloop (static version) and introduces reflection-based Hyperloop.

This needs Appcelerator Platform Pro account as well as `hyperloop` module for Windows.

# How to Test

- Download Titanium SDK build artifacts from here: [titanium_mobile_windows_10_prs/1249/artifact/dist/windows](http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_10_prs/1249/artifact/dist/windows/) as well as [titanium_mobile_windows_8.1_prs/991/artifact/dist/windows](http://studio-jenkins.appcelerator.org/job/titanium_mobile_windows_8.1_prs/991/artifact/dist/windows/), and extract/replace them with your Titanium SDK `windows` directory.
- Make sure you have installed `hyperloop` module and you have Appcelerator Platform Pro account

```javascript
var win = Ti.UI.createWindow({ backgroundColor: 'green' });
win.addEventListener('open', function () {

    // Test instance method call
    var System_Object = require('System.Object');
    var object = new System_Object();
    alert(object.GetHashCode());
});
win.open();
```

In addition to that you'll need to set hyperloop module in your `tiapp.xml`.

```xml
<modules>
  <module platform="windows">hyperloop</module>
</modules>
```

PR #891, opened 2016-11-04
[TIMOB-23931] Stop enumerating parent & children for View

infosia:
[TIMOB-23931](https://jira.appcelerator.org/browse/TIMOB-23931)

`Ti.UI.View` and its subclasses has `parent` and `children` property, and they are enumerable. That makes it impossible to log them through `JSON.stringify` because of circular references. We can easily workaround it by configuring them non-enumerable.

```
    label = Ti.UI.createLabel();

    Ti.API.info(JSON.stringify(win,null,2));
});

win.add(label);
win.open();
```


garymathews:
I receive a `nullptr` error when running the test case

```
Unhandled exception thrown: read access violation.
this->layoutDelegate__._Ptr was nullptr.
```
`L757 in TitaniumKit View.cpp`

infosia:
`nullptr` in `this->layoutDelegate__` should be fixed in #888, so let's merge this PR after #888 is pushed to the master.

infosia:
Should be ready for merge.

garymathews:
CR: **PASS**
FT: **PASS**

👍

PR #898, opened 2016-11-28
[TIMOB-24111] Parity issue in TableView touch events

infosia:
*NOT READY FOR MERGE*

This PR is not ready for merge. 

[TIMOB-24111](https://jira.appcelerator.org/browse/TIMOB-24111)

This is parity issue in `TableView` touch related events.

- `click` event should contain `detail`, `index`, `row`, `rowData`, `searchMode` and `section`.
- `touchstart`, `touchend`, `singletap`, `doubletap`, `longpress` should return same properties with `click` event.
- `x` and `y` may be considered optional.

But it turns out that touch events other than `click` does not return any row data, just because platform events (`Tapped`, `Holding` etc) does not return any row-related data such as row index. I tried to get row index by calculating event position but had no luck. Pushing this for review to see if we can get better way to do this from community.

### DIFF
--- a/Source/UI/src/TableView.cpp
+++ b/Source/UI/src/TableView.cpp
@@ -9,6 +9,7 @@
 #include "TitaniumWindows/UI/TableView.hpp"
 #include "TitaniumWindows/UI/TableViewRow.hpp"
 #include "Titanium/UI/TableViewSection.hpp"
+#include "Titanium/UI/SearchBar.hpp"
 #include <collection.h>
 #include <windows.h>
 #include "TitaniumWindows/UI/View.hpp"
@@ -25,6 +26,7 @@ namespace TitaniumWindows
 
 		using namespace Platform::Collections;
 		using namespace Windows::UI::Xaml;
+		using namespace Windows::UI::Xaml::Input;
 		using namespace Windows::Foundation;
 
 		TableView::TableView(const JSContext& js_context) TITANIUM_NOEXCEPT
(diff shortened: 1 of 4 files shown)


Source/UI/src/TableView.cpp
```diff
@@ -427,45 +429,108 @@ namespace TitaniumWindows
...
 
-						fireEvent("click", eventArgs);
-					}
+			if (event_name == "click") {
+				click_event__ = tableview__->ItemClick += ref new Controls::ItemClickEventHandler(
+					[this, ctx](Platform::Object^ sender, Controls::ItemClickEventArgs^ e) {
+					auto result = searchRowBySelectedRow(e->ClickedItem);
+					const auto component = safe_cast<FrameworkElement^>(e->ClickedItem);
+					result.x = Controls::Canvas::GetLeft(component);
+					result.y = Controls::Canvas::GetTop(component);
```
infosia: Note: `Controls::Canvas::GetLeft` and `GetTop` always return 0 here. 
We need a way to get absolute position from clicked item.

Source/UI/src/TableView.cpp
```diff
@@ -427,45 +429,108 @@ namespace TitaniumWindows
...
+					fireTableViewRowEvent("click", result);
 				});
+			} else if (event_name == "touchstart") {
+				tableview__->PointerPressed += ref new PointerEventHandler([this](Platform::Object^ sender, PointerRoutedEventArgs^ e) {
+					const auto component = safe_cast<Controls::ListView^>(sender);
+					const auto point = Windows::UI::Input::PointerPoint::GetCurrentPoint(e->Pointer->PointerId);
+					auto result = searchRowBySelectedRow(component->SelectedItem);
+					result.x = point->Position.X;
+					result.y = point->Position.Y;
```
infosia: Note: We can get the position correctly here, but we can't get row info because `component->SelectedItem` goes nullptr and `component->SelectedIndex` goes -1 here.

Source/UI/src/TableView.cpp
```diff
@@ -427,45 +429,108 @@ namespace TitaniumWindows
...
+					result.y = Controls::Canvas::GetTop(component);
+					fireTableViewRowEvent("singletap", result);
+				});
+			} else if (event_name == "doubletap") {
+				doubletap_event__ = tableview__->DoubleTapped += ref new DoubleTappedEventHandler([this](Platform::Object^ sender, DoubleTappedRoutedEventArgs^ e) {
+					const auto selectedItem = safe_cast<FrameworkElement^>(e->OriginalSource);
+					const auto position = e->GetPosition(tableview__);
+					auto result = searchRowBySelectedRow(selectedItem);
+					result.x = position.X;
+					result.y = position.Y;
```
infosia: Note: We can get the position correctly here, but we can't get row info because `component->SelectedItem` goes nullptr and `component->SelectedIndex` goes -1 here.

PR #909, opened 2016-12-19
[TIMOB-20544] Align method name for Ti.Media.AudioRecorder

infosia:
[TIMOB-20544](https://jira.appcelerator.org/browse/TIMOB-20544)

- This defines `Ti.Media.requestRecorderPermission()` as alias of `Ti.Media.requestAuthorization()`
- On Windows, `requestRecorderPermission` basically does nothing because there's no equivalent function in WinRT and recorder permission should be defined in `tiapp.xml` instead.

```js
// Should show "function"
alert(typeof Ti.Media.requestRecorderPermission);
```


garymathews:
CR: **PASS**
FT: **PASS**

PR #910, opened 2016-12-21
See which test causes the build issue

infosia:
Relates to #908

We've been seeing that some tests in ListView hangs when we enable this HttpClient test. Trying to see what's going on after HttpClient runs.


infosia:
So this means every single call from HttpClient could be cause of the crash. Closing.

PR #912, opened 2016-12-22
[TIMOB-23613] Attempt to split encryption process

infosia:
Relates to [TIMOB-23613](https://jira.appcelerator.org/browse/TIMOB-23613)

Attempt to split encryption process due to the issue around `spawn ENAMETOOLONG`.
This needs new `titanium_prep` because we need a way to specify encryption key for following commands. https://github.com/appcelerator/titanium_prep/pull/26

### DIFF
--- a/cli/commands/_build/encrypt.js
+++ b/cli/commands/_build/encrypt.js
@@ -20,15 +20,41 @@ exports.mixin = mixin;
  */
 function mixin(WindowsBuilder) {
 	WindowsBuilder.prototype.processEncryption = processEncryption;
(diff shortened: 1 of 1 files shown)


infosia:
So just pushed a quick fix, this may look dirty (I was using unused argument to specify encryption seed for titanium_prep) https://github.com/appcelerator/titanium_prep/pull/26 I was able to make it work locally by this.

infosia:
This workaround may work good, but I'm afraid how splitting Java process affects build time. Obviously it will be bad. I'll come up with `--file-listing` solution that @sgtcoolguy suggests.

infosia:
#913

PR #914, opened 2016-12-28
[TIMOB-24263] Investigate crash on HttpClient

infosia:
DO NOT MERGE

[TIMOB-24263](https://jira.appcelerator.org/browse/TIMOB-24263)

We've been seeing that Windows 10 node on Jenkins tends to crash around UI tests (ListView etc) when we enable HttpClient test. I suspect something is leaking/blocking after use of HttpClient.


infosia:
Interesting...this patch makes tests pass! :1st_place_medal:

infosia:
Closing, as taken over by #916

PR #916, opened 2017-01-05
[TIMOB-24268] Fix crash/freeze in HttpClient

infosia:
[TIMOB-24268](https://jira.appcelerator.org/browse/TIMOB-24268)

Attempt to fix crash/freeze in HttpClient.

It may be because we used `concurrency::cancellation_token_source` wrong, we did cancel it twice at HttpClient destructor. We might just want to use it as local variable so it doesn't bother.


infosia:
#917

PR #927, opened 2017-01-18
[TIMOB-24318] Stop use of Sleep()

infosia:
[TIMOB-24318](https://jira.appcelerator.org/browse/TIMOB-24318)

It seems Microsoft starts to deprecate some threading functions that makes Windows App Certification Kit fails even when they are marked "*Supported on Windows Phone 8.1*"  :anguished:

* [SuspendThread](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686345(v=vs.85).aspx) (used in JavaScriptCore)
* [ResumeThread](https://msdn.microsoft.com/en-us/library/windows/desktop/ms685086(v=vs.85).aspx)  (used in JavaScriptCore)
* [Sleep](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686298(v=vs.85).aspx)

Fortunately we have `std::this_thread::sleep_for` for alternative of `Sleep`.


infosia:
test this please

garymathews:
CR: **PASS**

PR #940, opened 2017-01-31
[TIMOB-24134] TableView.data should return TableViewSection[]

infosia:
[TIMOB-24134](https://jira.appcelerator.org/browse/TIMOB-24134)

`TableView.data` should return `TableViewSection[]`. As far as I can see from the source, `data` always returns `TableViewSection[]` on [iOS](https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/TiUITableViewProxy.m#L782) and [Android](https://github.com/appcelerator/titanium_mobile/blob/master/android/modules/ui/src/java/ti/modules/titanium/ui/TableViewProxy.java#L697). The reason why type of `Ti.UI.TableView.data` is marked both `Titanium.UI.TableViewRow[]/Titanium.UI.TableViewSection[]` in [API doc](http://docs.appcelerator.com/platform/latest/#!/api/Titanium.UI.TableView-property-data) I think is that it _accepts_ `Titanium.UI.TableViewRow[]` as well as `Titanium.UI.TableViewSection[]` whereas `data` _always returns_ `Titanium.UI.TableViewSection[]`. So Windows implementation was wrong in this case.

```js
var _window = Ti.UI.createWindow();
    top: 0,
    bottom: 0
});

var tableData = [];

for (var i = 0; i < 10; i++) {
    tableData.push(Ti.UI.createTableViewRow({
        title: "Row " + (i + 1)
    }));
}

table.setData(tableData);

_window.add(table);

table.addEventListener("click", function (e) {
    alert(table.data[0].rows[e.index].title);
});
_window.open();
```

### DIFF
--- a/Source/TitaniumKit/include/Titanium/UI/TableView.hpp
+++ b/Source/TitaniumKit/include/Titanium/UI/TableView.hpp
@@ -56,7 +56,7 @@ namespace Titanium
 			  @abstract data
 			  @discussion Rows of the table view.
 			*/
-  			TITANIUM_PROPERTY_IMPL_DEF(std::vector<JSObject>, data);
+			virtual void set_data(const std::vector<JSObject>& data) TITANIUM_NOEXCEPT;
 
 			/*!
 			  @method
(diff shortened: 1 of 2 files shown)


garymathews:
CR: **PASS**
FT: **PASS**

👍

PR #941, opened 2017-01-31
Move body of builds from batch scripts into Jenkinsfile, to hopefully…

sgtcoolguy:
… set up for better parallelization of build eventually and maybe more stability.


sgtcoolguy:
Squashed and merged manually.

PR #947, opened 2017-02-08
[TIMOB-24385] Fix Ti.Network.Socket.TCP.readAsync() to be continuous

garymathews:
- Fixed continuous output from `readAsync()`
- Improved error/exception handling for both `read` and `write` callbacks

##### app.js
```Javascript
var socket = Ti.Network.Socket.createTCP({
    host: '127.0.0.1',
    port: 6262,
    timeout: 3000,

    connected: function (e) {
        console.log('connected to server');

        // recieve data
        Ti.Stream.pump(e.socket, function (e) {
            if (!e.success) {
                console.error('recv: ' + e.error);
                return;
            }
            console.log('recv: ' + e.buffer.toString());
        }, 1024, true);

        // send data
        var i = setInterval(function () {
            if (!socket) {
                clearInterval(i);
                return;
            }

            console.log('sending data...');
            Ti.Stream.write(socket, Ti.createBuffer({ value: 'test' }), function (e) {
                if (!e.success) {
                    socket = null;
                    console.error('send: ' + e.error);
                    return;
                }
                console.log('sent');
            });
        }, 5000);
    },
    error: function (e) {
        clearInterval(i);
        socket = null;
        console.error(e);
    }
});

console.log('connecting...');
socket.connect();
```

##### index.js
```Javascript
var net = require('net'),
	port = 6262;

net.createServer(function (socket) {

	console.log('client ' + socket.remoteAddress + ' connected');

	socket.on('data', function (data) {
		console.log('recv: ' + data);
		socket.write(data);
	});

	socket.on('end', function () {
		console.log('client ' + socket.remoteAddress + ' disconnected');
	});

}).listen(port);

console.log('server listening on port ' + port);
```
- Run the TCP server (*index.js*) and run the test case application
- The `Ti.Stream.pump` callback should continuously be receiving data
- Terminate the server while the test case application is still running
- The test case application should error elegantly

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24385)

### DIFF
--- a/Source/Network/src/Socket/TCP.cpp
+++ b/Source/Network/src/Socket/TCP.cpp
@@ -169,6 +169,7 @@ namespace TitaniumWindows
 			{
 				if (state__ == Titanium::Network::Socket::State::Connected ||
 					state__ == Titanium::Network::Socket::State::Listening) {
+					state__ = Titanium::Network::Socket::State::Closed;
 					delete socket__;
 				} else {
 					error("socket is not in connected or listening state");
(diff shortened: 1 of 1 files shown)


infosia:
I saw a crash when server is down _before_ app is launched. In this case C++ exception is happening while `connect`, and then `error` callback is called accordingly. But in this case `error()` callback throws JavaScript exception because `undefined` is specified in `clearInterval()`.  Note that `i` goes `undefined` when it is called _before_ `connect()` in your example.

```js
    error: function (e) {
        clearInterval(i); // i is undefined!
        socket = null;
        console.error(e);
    }
```

So in this case we found an issue in clearInterval actually, `clearInterval(undefined)` should never throws exception :( But also, what we need to do here is that we should be making sure that invoking JavaScript callback never throws exception even in this case.

```c++
void TCP::error(const std::string& message)
{
	try {
		if (error__.IsObject()) {
			auto error_obj = static_cast<JSObject>(error__);
			if (error_obj.IsFunction()) {
...
				error_obj({ args }, get_object());
			}
		}
	} catch (...) {
                // last resort
		TITANIUM_LOG_WARN("oops");
	}
}
```

infosia:
**CR** & **FR** Looks good :)

PR #948, opened 2017-02-09
[TIMOB-24382] Adding same component should not cause error

infosia:
[TIMOB-24382](https://jira.appcelerator.org/browse/TIMOB-24382)

Adding same component multiple times should not cause error to align with iOS behavior.

```js

var label = Ti.UI.createLabel({ text: 'Test' });

win.add(label);
win.add(label);

win.open();
```


garymathews:
CR: **PASS**
FT: **PASS**

👍

PR #954, opened 2017-02-15
[TIMOB-24132] Implement TableView.touchmove event

infosia:
[TIMOB-24132](https://jira.appcelerator.org/browse/TIMOB-24132)

`touchmove` event for `TableView` did not work. We need to use Xaml `PointerMoved` event for TableView here because Xaml `ListView` doesn't fire `ManipulationDelta` event which is widely used in TitaniumKit, more lightweight version of `PointerMoved`.

```js
var _window = Ti.UI.createWindow();

var data = [];
for (var i = 1; i <= 10; i++) {
    data.push({
        title: 'Row ' + i
    });
}
var MyListbox = Ti.UI.createTableView({ data: data, top: 0, left: 0, width: Ti.UI.FILL, height: Ti.UI.FILL });

MyListbox.addEventListener('touchmove', TchMove);

function TchMove(evt) {
    Ti.API.info('TchMove');
}

_window.add(MyListbox);
_window.open();
```


garymathews:
CR: **PASS**
FT: **PASS**

PR #958, opened 2017-02-23
[TIMOB-24422] Skip installing dependencies

infosia:
[TIMOB-24422](https://jira.appcelerator.org/browse/TIMOB-24422)

Skip installing app dependencies manually. WinAppDeployCmd should be taking care of that for now. I also added `--installDependencies` flag just to keep backward compatibility.

```
> appc run -p windows --target wp-emulator --wp-sdk 10.0 -l trace
```

### DIFF
--- a/cli/hooks/wp-run.js
+++ b/cli/hooks/wp-run.js
@@ -268,28 +268,30 @@ exports.init = function (logger, config, cli) {
 						});
 				}
 
-				// Install dependencies
-				var possibleDependencies = fs.readdirSync(dependenciesDir);
-				possibleDependencies = possibleDependencies.filter(function(file) {
-					return appxExtensions.indexOf(path.extname(file)) !== -1;
-				});
-				possibleDependencies.forEach(function(file) {
-					installs.push(function (next) {
-						logger.info(__('Installing dependency: %s', file));
-						windowslib.install(builder.deviceId, path.resolve(dependenciesDir, file), installOnlyOpts)
-						.on('installed', function (handle) {
-							next();
-						})
-						.on('timeout', function (err) {
-							logRelay && logRelay.stop();
-							next(err.message);
-						})
-						.on('error', function (err) {
-							logRelay && logRelay.stop();
-							next(err.message);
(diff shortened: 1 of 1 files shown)


sgtcoolguy:
So, I finally nailed this down to invalid windows Firewall rules as the cause. I had set up inbound TCP, but not UDP - which actually resulted in this error happening intermittently. Yuck.

sgtcoolguy:
So, I've been looking at this issue occasionally (on build machines). I strongly suspect something is up with the new hyperloop stuff and the app dependencies/min version OS. From what I've found from Googling, some people have had similar issues that were magically fixed by bumping the min OS version in the AppPackage manifest xml file, which really related to the dependency libraries that got installed.
Additionally, if I install an app from VS 2015 onto the emulator first and keep it open and then build, it seems to work most of the time.
I think we're not installing some dependency library we should or something (I recall seeing a new .NET core dependency getting installed from a branch new universal app in VS2015 when I run it on the emulator)

infosia:
Thanks for update. That sounds this error may be related to something I got while integrating Visual Studio 2017 #959 too.

infosia:
I ended up finding a way to explicitly running Visual Studio IDE as a compiler (using [Devenv](https://msdn.microsoft.com/en-us/library/b20w810z.aspx)) behind the scenes and I hope this works for Jenkins dependency issue too. I will update this PR soon to see how it works on Jenkins.

infosia:
So it turns out that `devenv` does not quite work well for this situation because it does not generate application packages (appx). I found interesting behavior while working on #959, that is *some batch tools (`.bat`) does not work well when it is executed by [8.3 filenames](https://en.wikipedia.org/wiki/8.3_filename). We are usually using 8.3 filename when requiring `vcvarsall.bat` but it seems it does not work correctly with 8.3 filename. I'm also guessing that maybe we should not use `vcvarsall.bat` but should use `VsDevCmd.bat` instead because it's the file that Visual Studio Developer Console is using. I was able to make it work locally with Visual Studio 2017 with long filename (I mean not 8.3 filename).

infosia:
Closing, I am pushing #962 instead.

PR #959, opened 2017-02-27
[TIMOB-24189] Support Visual Studio 2017

infosia:
[TIMOB-24189](https://jira.appcelerator.org/browse/TIMOB-24189)

~~_Requires https://github.com/appcelerator/windowslib/pull/65 and  https://github.com/appcelerator/windowslib/pull/66_~~
~~_Requires new windowslib_ - https://github.com/appcelerator/titanium_mobile/pull/8893~~

- Add support for Visual Studio 2017 installation

As of Visual Studio 2017, you can install multiple editions of Visual Studio (Community, Professional etc) and it requires new way to detect its instances. (See also  [this](https://github.com/Microsoft/vs-setup-samples) and [this](http://www.visualstudioextensibility.com/2016/11/23/some-implications-of-the-new-modular-setup-of-visual-studio-2017-for-vsx-developers/) ). 

So the problem for `windowslib` is that the result from `windowslib.visualstudio.detect` actually did not quite consider this situation and it uses MSBuild version number as a key to select Visual Studio such as `12.0` and `14.0`. You might think we could use `15.0` for VS 2017, but this does not work because we could have multiple installation of MSBuild that have same version number `15.0`. So for now we should have a way to choose one instance from multiple installation of Visual Studio 2017 _by its name_ such as `Visual Studio Community 2017 RC` and `Visual Studio Enterprise 2017 RC` etc.

So for now `visualstudio` output from `appc ti info -p windows -o json` is going to look like this:

```json
"visualstudio": {
        "Visual Studio Community 2017 RC": {
                "version": "Visual Studio Community 2017 RC",
                "supported": "true",
                "msbuildVersion": "15.0.26206.0",
                "path": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community",
                "wpsdk": null,
                "registryKey": null,
                "clrVersion": null,
                "selected": true,
                "vcvarsall": "C:\\PROGRA~2\\MIB055~1\\2017\\COMMUN~1\\VC\\AUXILI~1\\Build\\VCVARS~1.BAT"
        },
        "Visual Studio Professional 2017 RC": {
                "version": "Visual Studio Professional 2017 RC",
                "supported": "true",
                "msbuildVersion": "15.0.26206.0",
                "path": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Professional",
                "wpsdk": null,
                "registryKey": null,
                "clrVersion": null,
                "selected": false,
[...]

PR #962, opened 2017-03-14
Use VsDevCmd instead of vcvarsall

infosia:
I found interesting behavior while working on https://github.com/appcelerator/titanium_mobile_windows/pull/959, that is some VS batch tools (including `vcvarsall.bat`) does not work correctly when it is executed by [8.3 filenames](https://en.wikipedia.org/wiki/8.3_filename). We are usually using 8.3 filename when requiring `vcvarsall.bat` but it seems it does not work correctly under the situation. I'm also thinking that we should use `VsDevCmd.bat` instead because it's the file that Visual Studio Developer Console is using. We can safely replace it because  `VsDevCmd.bat` is available as of Visual Studio 2012.


infosia:
Merging changes in #963 to see how it works with this.

infosia:
Closing, fixed by #963

PR #961, opened 2017-03-08
[TIMOB-24454] Improve library integration with module

infosia:
[TIMOB-24454](https://jira.appcelerator.org/browse/TIMOB-24454)

Package libraries just like what Titanium Android module does for use of 3rd party JARs. 

- Every \*.lib, \*.dll and \*.winmd files in *lib* folder will be packaged into module and also copied/linked during Ti app build.
- Every files in *platform* folder are packaged into module and copied into Ti app so module can overwrite files during Ti app build.

For steps to test and sample code, refer to [TIMOB-24454](https://jira.appcelerator.org/browse/TIMOB-24454)

PR #967, opened 2017-04-07
[TIMOB-24101] Ability to use 3rd-party native classes

infosia:
[TIMOB-24101](https://jira.appcelerator.org/browse/TIMOB-24101)

Ability to use 3rd-party Windows Runtime component with hyperloop.

**This requires #961 to be merged.** This PR is actually created on top of #961 so this includes these changes too. Checkout individual commit https://github.com/appcelerator/titanium_mobile_windows/commit/404f3a4355f43308af2508400284ecda9b0e768b for code review. This requires latest Hyperloop module.

**Steps to test**

1. Create Windows Runtime Component

Create Windows Runtime Component named `WindowsRuntimeComponent1` from Visual Studio. Compile it and you should get `WindowsRuntimeComponent1.winmd` somewhere under in your Visual Studio project folder. You can skip this step by using prebuilt [WindowsRuntimeComponent1.winmd](https://jira.appcelerator.org/secure/attachment/61987/61987_WindowsRuntimeComponent1.winmd).

```csharp
using System;

namespace WindowsRuntimeComponent1
{
    public sealed class Test
    {
        public int SampleNumber { get; set; }

        public static string SayHello(string str)
        {
            return String.Format("Hello, {0}!", str);
        }

        public string sayNumber()
        {
            return String.Format("SampleNumber is {0}.", SampleNumber);
        }
    }
}
```

2. Copy your `WindowsRuntimeComponent1.winmd` file into `lib/windows/win10/x86/` under  your Titanium app project folder.

**app.js**

```javascript
    Test = require('WindowsRuntimeComponent1.Test'),
    test = new Test();

win.addEventListener('open', function() {
    alert(test.SayHello('World!'));
});

win.open();
```

3. Place `appc.windows.js` directly under your Titanium project folder. Make sure it contains names of your Windows Runtime component under `hyperloop.windows.thirdparty` key. It should look like below.

**appc.windows.js**

```js
/**
 * Hyperloop configuration
 */
module.exports = {
	type: 'app',
	group: 'titanium',
	dependencies: {},
	hyperloop: {
		windows:
		{
			/**
			 * Optionally, you can bring in third-party or first-party Windows Runtime components.
			 * The 'key' is the root namespace of the component that will be used in the require.
			 * There's no values for now but we reserve them for future release.
			 *
			 * Place the library files into the lib/windows/(phone|store|win10)/(x86|ARM)/ folder under your app project.
			 * Hyperloop will pick up the winmd files and will generate necessary bindings and include the winmd in your app. 
			 */
			thirdparty: {
				'WindowsRuntimeComponent1': {}
			}
		}
	}
};
```
4. Make sure to enable hyperloop, and then run it using `appc run -p windows --wp-sdk 10.0 --target ws-local -l trace`.
5. Then you should see "Hello, World!!" dialog box in your Titanium app.

### DIFF
--- a/cli/commands/_build/initialize.js
+++ b/cli/commands/_build/initialize.js
@@ -85,5 +85,10 @@ function initialize(next) {
 	this.cmakeListFile = path.join(this.buildDir, 'CMakeLists.txt'); // lives above the buildSrcDir
 	this.cmakeFinderDir = path.join(this.buildDir, 'cmake');
 
+	// Hyperloop configuration
+	var hyperloopAppcJs = path.join(this.projectDir, 'appc.js');
+	this.hyperloopConfig = fs.existsSync(hyperloopAppcJs) && require(hyperloopAppcJs).hyperloop || {};
+	this.hyperloopConfig.windows || (this.hyperloopConfig.windows = {});
+
 	next();
 }
--- a/cli/commands/_build/initialize.js
+++ b/cli/commands/_build/initialize.js
@@ -86,7 +86,7 @@ function initialize(next) {
 	this.cmakeFinderDir = path.join(this.buildDir, 'cmake');
 
 	// Hyperloop configuration
-	var hyperloopAppcJs = path.join(this.projectDir, 'appc.js');
+	var hyperloopAppcJs = path.join(this.projectDir, 'appc.windows.js');
 	this.hyperloopConfig = fs.existsSync(hyperloopAppcJs) && require(hyperloopAppcJs).hyperloop || {};
 	this.hyperloopConfig.windows || (this.hyperloopConfig.windows = {});
 
--- a/cli/lib/stub.js
+++ b/cli/lib/stub.js
@@ -170,7 +170,7 @@ function generateNativeProject(dest, platform, builder, options, callback) {
 	for (var i = 0; i < thirdpartyLibraries.length; i++) {
 		externalReferences.push({
 			Include: thirdpartyLibraries[i],
-			HintPath: '..\\..\\lib\\' + platform + '\\' + builder.arch + '\\' + thirdpartyLibraries[i] + '.winmd'
+			HintPath: path.join('..', '..', 'lib', platform, builder.arch, thirdpartyLibraries[i] + '.winmd')
 		});
 	}
 
--- a/cli/lib/stub.js
+++ b/cli/lib/stub.js
@@ -15,17 +15,7 @@ var fs = require('fs'),
 	path = require('path'),
 	async = require('async'),
 	spawn = require('child_process').spawn,
-	logger, windowsInfo,
-	targetPlatformSdkVersion, targetPlatformSdkMinVersion,
-	cmakePlatform;
-
-exports.init = function(thiz) {
-	logger = thiz.logger;
-	windowsInfo = thiz.windowsInfo;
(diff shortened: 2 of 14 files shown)


cli/lib/stub.js
```diff
@@ -171,17 +161,25 @@ function buildNativeTypeHelper(dest, platform, buildConfiguration, callback) {
...
-
-	logger.trace("Updating target version for " + platform);
+		csproj   = path.join(dest, platform, 'TitaniumWindows_Hyperloop.csproj'),
+		externalReferences = [];
+		thirdpartyLibraries = builder.hyperloopConfig.windows.thirdparty && Object.keys(builder.hyperloopConfig.windows.thirdparty) || [];
+
+	for (var i = 0; i < thirdpartyLibraries.length; i++) {
+		externalReferences.push({
+			Include: thirdpartyLibraries[i],
+			HintPath: '..\\..\\lib\\' + platform + '\\' + builder.arch + '\\' + thirdpartyLibraries[i] + '.winmd'
```
garymathews:
```JS
path.join('..', '..', 'lib', platform, builder.arch, thirdpartyLibraries[i]) + '.winmd'
```

cli/commands/_build/initialize.js
```diff
@@ -85,5 +85,10 @@ function initialize(next) {
 	this.cmakeListFile = path.join(this.buildDir, 'CMakeLists.txt'); // lives above the buildSrcDir
 	this.cmakeFinderDir = path.join(this.buildDir, 'cmake');
 
+	// Hyperloop configuration
+	var hyperloopAppcJs = path.join(this.projectDir, 'appc.js');
```
garymathews: Could we rename `appc.js` to something else, when executing `appc` commands in the directory it will open the `appc.js` file instead.

infosia: Yeah, I'm not sure if it's a good name but the reason why I use `appc.js` is that I want to align with how iOS handles 3rd party libraries. See also [iOS Hyperloop Programming Guide](http://docs.appcelerator.com/platform/latest/#!/guide/iOS_Hyperloop_Programming_Guide-section-src-46253491_iOSHyperloopProgrammingGuide-Usingthird-partylibraries) and [appcelerator/hyperloop-examples/appc.js](https://github.com/appcelerator/hyperloop-examples/blob/master/appc.js)

cli/lib/stub.js
infosia: Makes sense! 👍

cli/commands/_build/initialize.js
garymathews: it's unfortunate, but on Windows the command line will open any local file `appc.*` first instead of executing the `appc` command :(

ewanharris: @garymathews See https://jira.appcelerator.org/browse/CLI-1038

hansemannn: We should rename the appc.js soon, agree. But for now, it's the main Hyperloop config file and that won't change until 3.0.0 I guess.

infosia: hmm...I think it's ok to rename (only for Hypeloop Windows) for now because we haven't released Hyperloop Windows yet. For now I would use something like "hyperloop.json".

infosia: ah maybe "hyperloop.js" would be better because it's not a valid JSON...If you guys have something to suggest that would be nice.

hansemannn: Can you use "appc.windows.js" ? We will definitely migrate to something unified in the future, but just so it's more unified.

infosia: ok 👍

PR #991, opened 2017-05-05
[TIMOB-24618] Windows: Replace uglify usage in build process

sgtcoolguy:
**JIRA**: https://jira.appcelerator.org/browse/TIMOB-24618

**Description**
uglify-js **still** does not support ES6/7+, which has the effect of breaking our CLI handling non-ES5 JS code. We want to be able to support users writing JS code in ES6+ because our JS engines often support newer features.

To do so, we're trying to move away from using uglify-js for any parsing/walking AST/transforming of JS code; and moving to the babel ecosystem. This PR attempts to replace usage of uglify-js with the babel equivalents in the Windows CLI code.

This code appears to be very specific to Windows hyperloop where we're detecting use of native requires/types. As such, I think the real test here is:

1. Can the user make use of ES6 syntax in their app.js (or elsewhere in their app) and the CLI won't barf on it?
2. Does this still handle hyperloop usage and properly detect reference native types?

### DIFF
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -8,7 +8,10 @@ var appc = require('node-appc'),
 	path = require('path'),
 	ti = require('node-titanium-sdk'),
 	wrench = require('wrench'),
-	UglifyJS = require('uglify-js'),
+	babel = require('babel-core'),
+	babylon = require('babylon'),
+	types = require('babel-types'),
+	traverse = require('babel-traverse').default,
 	__ = appc.i18n(__dirname).__;
 
 /*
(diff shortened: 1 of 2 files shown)


sgtcoolguy:
@infosia So I have to admit, I did this PR "blind" - meaning I did not build and test it. I modified the parsing/AST code by hand and using astexplorer.net 
I can push it to a branch on the main repo for you to extend if you'd like, since I'm not currently set up for the Windows SDK development env. I just wanted to give you a starting point, because the ES6 uglify-js -> babel work has already been merged to master and 6_1_X, so I think until this PR works and is merged, the windows CLI may be broken (since uglify-js may not be in the SDK's node_modules folder anymore?)

infosia:
@sgtcoolguy Yes, we needed `uglify-js` to make Hyperloop work. I will take a look. Is it possible to keep `uglify-js` in node_modules until we resolve TIMOB-24618 just to make sure it doesn't break the CLI?

infosia:
Pushing #992

PR #996, opened 2017-05-11
[TIMOB-24644] Decouple Hyperloop logic and Ti SDK

infosia:
[TIMOB-24644](https://jira.appcelerator.org/browse/TIMOB-24644)

Move Hyperloop-related code & logic to hyperloop plugin.

This PR doesn't actually remove existing Hyperloop logic and template files, but provides "compatibility mode" so we can run older version of Hyperloop module (< 2.2.0) in the newer version of Titanium SDK (>= 6.2.X).

### DIFF
--- a/Source/UI/include/TitaniumWindows/UI/WebView.hpp
+++ b/Source/UI/include/TitaniumWindows/UI/WebView.hpp
@@ -47,7 +47,7 @@ namespace TitaniumWindows
 			static std::string InjectLocalScript(const std::string& content);
 
 			virtual void postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) override;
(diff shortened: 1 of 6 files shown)


Source/UI/include/TitaniumWindows/UI/WebView.hpp
```diff
@@ -47,7 +47,7 @@ namespace TitaniumWindows
 			static std::string InjectLocalScript(const std::string& content);
 
 			virtual void postCallAsConstructor(const JSContext& js_context, const std::vector<JSValue>& arguments) override;
-			virtual std::string evalJS(const std::string& code, JSObject& callback) override TITANIUM_NOEXCEPT;
+			virtual std::string evalJS(const std::string& code, JSObject& callback) TITANIUM_NOEXCEPT override;
```
infosia: pushing these because this is syntax issue and could cause compile error.

PR #998, opened 2017-05-12
[TIMOB-24674] Windows: require call in node_modules directory fails to find module

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-24674

Notes

- Tried to cleanup inconsistent formatting around the file while I was there


Source/TitaniumKit/src/GlobalObject.cpp
```diff
@@ -175,12 +179,12 @@ namespace Titanium
 			resolvedPath = COMMONJS_SEPARATOR__ + resolvedPath;
 		}
 		for (size_t i = 0, len = reqPaths.size(); i < len; i++) {
-			auto newResolvedPath = reqPaths[i] + resolvedPath;
-			modulePath = resolvePathAsFile(parent,newResolvedPath);
+			auto newResolvedPath = reqPaths[i] + COMMONJS_SEPARATOR__ + moduleId;
```
ewanharris: Before, this would give us a path like `"/node_modules/bar/node_modules/node_modules/bar/baz"`, now this gives "/node_modules/bar/node_modules/baz"

Source/TitaniumKit/src/GlobalObject.cpp
```diff
@@ -329,7 +337,7 @@ namespace Titanium
 					const std::string app_module_js = "try {__dirname='/',__filename='app.js'; " + module_js + "} catch (E) { E.fileName='app.js'; Titanium_RedScreenOfDeath(E);}";
 					result = js_context.JSEvaluateScript(app_module_js, js_context.get_global_object());
 				} else {
-					const std::string require_module_js = "(function(global) { var exports={},__OXP=exports,module={'exports':exports},__dirname='" + currentDir__ + "',__filename='/"
+					const std::string require_module_js = "(function(global) { var exports={},__OXP=exports,module={'exports':exports},__dirname='" + currentDir__ + "',__filename='"
```
ewanharris: The leading slash here was removed because of the change on line [232](https://github.com/appcelerator/titanium_mobile_windows/pull/998/files#diff-3b6d0c60899ff30352e5cc386b90fa56R232) to append it to modulePath before returning from the function. Before this change we would get filenames like //foo.js rather than the expected /foo.js

Source/TitaniumKit/src/GlobalObject.cpp
```diff
@@ -53,20 +53,24 @@ namespace Titanium
 	{
 		std::vector<std::string> paths;
 		if (dirname == COMMONJS_SEPARATOR__) {
-			paths.push_back("node_modules");
+			paths.push_back("/node_modules");
+			return paths;
 		} else {
```
garymathews: This else looks like it can be replaced with this:
```C++
if (dirname != COMMONJS_SEPARATOR__) {
```

ewanharris:
Closing in favour of #1009

PR #1002, opened 2017-05-23
[TIMOB-24714] ListView's scroll breaks if search bar added

infosia:
[TIMOB-24714](https://jira.appcelerator.org/browse/TIMOB-24714)

`ListView` doesn't scroll when use with SearchBar.

```js
var _ = require('underscore');

var win = Ti.UI.createWindow({
    backgroundColor: 'white'
});

var searchBar = Ti.UI.createSearchBar({
    backgroundColor: 'red'
});

var usersListView = Ti.UI.createListView({
    height: Ti.UI.FILL,
    width: Ti.UI.FILL,
    backgroundColor: 'purple',
    searchView: searchBar
});

var sections = [];
var userSection = Ti.UI.createListSection();
sections.push(userSection);
usersListView.sections = sections;

var wrapper = Ti.UI.createView({
    top: 0,
    left: 0,
    height: Ti.UI.FILL,
    width: Ti.UI.FILL
});

wrapper.add(usersListView);
win.add(wrapper);

var url = 'https://randomuser.me/api/?results=150&nat=us';
var client = Ti.Network.createHTTPClient({
    // function called when the response data is available
    onload: function (e) {
        var response = JSON.parse(this.responseText);
        var userItems = _.map(response.results, function (_userObject) {
            return {
                properties: {
                    height: 40,
                    title: _userObject.name.first + ' ' + _userObject.name.last,
                    accessoryType: Ti.UI.LIST_ACCESSORY_TYPE_DISCLOSURE,
                    color: 'black',
                    font: {
                        fontSize: 16
                    }
                }
            };
        });

        userSection.setItems(userItems);
    },
    // function called when an error occurs, including a timeout
    onerror: function (e) {
        alert('error');
    },
    timeout: 15000  // in milliseconds
});
// Prepare the connection.
client.open("GET", url);
// Send the request.
client.send();

win.open();
```

You need [underscore.js for Titanium](https://github.com/appcelerator-developer-relations/Forging-Titanium/blob/master/ep-011/underscore/Resources/underscore.js) to make this demo work.

PR #1003, opened 2017-05-23
[6_1_X] [TIMOB-24714] ListView's scroll breaks if search bar added

infosia:
Cherry-pick #1002 for 6_1_X.

I would like to backport this into `6.1.x` if QA time permits.


ewanharris:
Adding do not merge label until we are after the 6.1.0 release

ssjsamir:
@ewanharris, @infosia This is safe to merge now right? and the "Do not merge" label can be removed?

infosia:
I believe yes, since we've already released 6.1.0.GA. It's safe to merge now. Just removed the label.

PR #1009, opened 2017-06-06
[TIMOB-24674] Windows: require call in node_modules directory fails to find module

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-24674

This is  a retry of #998, I believe I had messed up the jenkins workspace somehow as the failures were unrelated to my changes, so retrying in a fresh PR

Notes

- Tried to cleanup inconsistent formatting around the file while I was there


ewanharris:
Everything works locally when using NMocha via Visual Studio. It appears my hack around node_modules being gitignored causes an issue with the script used to copy the assets as they do not get copied to node_modules, I'm probably going to do what I should have done in the first place and add a rule to .gitignore to whitelist this files rather than workd around it

ewanharris:
Updated the gitignore and moved the files, hopefully CI should pass now 🙏

PR #1017, opened 2017-06-20
[TIMOB-24847] borderRadius does not work with text values

infosia:
[TIMOB-24847](https://jira.appcelerator.org/browse/TIMOB-24847)

On Windows UWP, the border values in [Windows.UI.Xaml.Thickness](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.thickness) are `double` and it does not actually accepts unit strings such as `dp`.  ~In this case we just want to make sure it does not throw exception, and just convert strings to double using `std::stod` then.~

```js
var win = Ti.UI.createWindow({backgroundColor:'green'}),
    lbl = Ti.UI.createView({
        borderRadius: '3dp',
        borderWidth: '1dp',
        borderColor: 'red',
        backgroundColor:'blue'
    }); 

win.add(lbl);
win.open();
```

Expected: Should not throw exception.


infosia:
Nice catch❗️ 👍

infosia:
> Shouldn't this be passed through _computeValue() to scale correctly?

@garymathews Addressed by c48fd8a8bb34057a833e0e572a9361c0b505a0ef

sgtcoolguy:
The Windows 8.1 phone emulator unit tests are failing because the app is crashing mid-run:
```
0:58:09 [INFO] :   Started: image (File)
10:58:09 
10:58:16 [INFO] :   Disconnected from app
```

PR #1018, opened 2017-06-21
[TIMOB-24867] borderRadius for View not working

infosia:
[TIMOB-24867](https://jira.appcelerator.org/browse/TIMOB-24867)

```js
    view = Ti.UI.createView({
        backgroundColor: 'blue',
        borderRadius: 50,
        borderColor: 'red',
        borderWidth: 3,
        width: '30%',
        height: '30%'
    });

win.add(view);
win.open();
```

**Expected**

view

### DIFF
--- a/Source/UI/src/WindowsViewLayoutDelegate.cpp
+++ b/Source/UI/src/WindowsViewLayoutDelegate.cpp
@@ -921,6 +921,8 @@ namespace TitaniumWindows
 				underlying_control__->Background = brush;
 			} else if ((is_grid__ || is_border__) && border__) {
 				border__->Background = brush;
+			} else if (is_panel__ && border__) {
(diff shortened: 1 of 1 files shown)


Source/UI/src/WindowsViewLayoutDelegate.cpp
```diff
@@ -921,6 +921,8 @@ namespace TitaniumWindows
 				underlying_control__->Background = brush;
 			} else if ((is_grid__ || is_border__) && border__) {
```
garymathews: How about this?

```Java
...
} else if ((is_grid__ || is_panel__ || is_border__) && border__) {
    border__->Background = brush;
...
```

infosia:
@garymathews Makes sense, pushed 👍

PR #1026, opened 2017-06-28
Remove hyperloop usage from NG app

ewanharris:
Building NG currently errors out, which I'm pretty sure is down to the hyperloop usage. So, I removed it, replacing with the default template code.

Is there any other parts I should remove?


infosia:
Looks good, merging.

PR #1030, opened 2017-06-30
[TIMOB-24915] Fix issue where ti cli could not use run cmake hook

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-24915

When run through titanium CLI this hook would be unable to resolve async and wrench and silently fail, to fix this

- Use the async that's internal to node-appc
- Replace usage of wrench with the same rmdir function in [alloy deep-clean hook](https://github.com/appcelerator/alloy/blob/master/hooks/deepclean.js#L45-L65)

@infosia  out of interest what was the reason for splitting this out into a hook? Could we not do this during the normal module build process?

Verification

1. Grab https://github.com/appcelerator-archive/ti.paint
2. Build the module using `ti build -p windows --build-only --run-cmake` (not `appc ti build`)
    - Should build successfully rerunning CMake
3. Build the module using `ti build -p windows --build-only --run-cmake` (not `appc ti build`)
    - Should build successfully removing the existing CMake dirs before rerunning CMake

### DIFF
--- a/templates/module/default/template/windows/plugins/hooks/windows.js
+++ b/templates/module/default/template/windows/plugins/hooks/windows.js
@@ -1,10 +1,6 @@
 var spawn = require('child_process').spawn,
-    async = require('async'),
     path = require('path'),
-    fs   = require('fs'),
-    ejs  = require('ejs'),
-    appc = require('node-appc'),
-    wrench = require('wrench');
+    fs   = require('fs');
 
 exports.cliVersion = ">=3.2";
 exports.init = function(logger, config, cli, nodeappc) {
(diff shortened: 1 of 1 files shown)


infosia:
@ewanharris 

> out of interest what was the reason for splitting this out into a hook? Could we not do this during the normal module build process?

Yes, that totally makes sense. We should be able to do that during normal module build process (I guess it's `cli/_buildModule.js`) As I remember at that time I was thinking use of plugin could give module developer a flexibility to customize which VS-project to re-build by editing the plugin, such as executing additional commands (like `nuget`) and copying resources etc because developer needs varies. But maybe I'm thinking too much.

infosia:
Verification updated,

1. Grab https://github.com/appcelerator-archive/ti.paint
2. Overwrite `ti.paint/windows/plugins/hooks/windows.js` with the `windows.js` in this PR
3. Change path to the preferred Titanium SDK here: ti.paint/blob/master/windows/CMakeLists.txt#L24, change it to the 6.2.x SDK path.
4. Build the module using `ti build -p windows --build-only --run-cmake` (not `appc ti build`)
    - Should build successfully rerunning CMake
5. Build the module using `ti build -p windows --build-only --run-cmake` (not `appc ti build`)
    - Should build successfully removing the existing CMake dirs before rerunning CMake

PR #1048, opened 2017-07-27
[6_2_X] Merge master changes into 6_2_X

infosia:
Merge all changes in master into 6_2_X.


ewanharris:
Thanks for that @infosia!

PR #1053, opened 2017-07-31
SearchBar missing events and property

infosia:
- [TIMOB-24969](https://jira.appcelerator.org/browse/TIMOB-24969) SearchBar.cancel event is not fired
- [TIMOB-24970](https://jira.appcelerator.org/browse/TIMOB-24970) Implement SearchBar.return event
- [TIMOB-24971](https://jira.appcelerator.org/browse/TIMOB-24971) Implement Ti.UI.SearchBar.barColor

```js

var view = Ti.UI.createView({
    width: '80%',
    height: '80%',
    backgroundColor: 'green'
});
var search = Titanium.UI.createSearchBar({
    width: 350,
    height: 143,
    barColor: 'red',
    hintText: 'Search for something!',
    top: 0
});

search.addEventListener('cancel', function () {
    alert('cancel');
});

search.addEventListener('return', function () {
    alert('return');
});

view.add(search);
win.add(view);
win.open();
```

Expected:

- `cancel` event should be fired when you click 'X' button that appears after input 
- `return` event should be fired when there's `return` keyboard input

searchbar


ewanharris:
@infosia When attempting to log the event to the console using JSON.stringify my app crashes. Could that be due to a similar problem like [TIMOB-23931](https://jira.appcelerator.org/browse/TIMOB-23931), it looks to happen due to the event.source value which makes me think it's the same cause.

ssjsamir:
Confirmed to be a separate issue by @infosia

PR #1066, opened 2017-08-07
[TIMOB-24984] Support XAML style templates

infosia:
**I accidentally closed orignal PR #1044. Pushing this again.**

[TIMOB-24984](https://jira.appcelerator.org/browse/TIMOB-24984)

Add flexible way to directly deal with [XAML style templates](https://docs.microsoft.com/en-us/windows/uwp/controls-and-patterns/xaml-styles). Introduced **Ti.UI.Windows.Style**.

**Resource/Styles.xaml**
```xml
<ResourceDictionary
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" 
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">

    <ResourceDictionary.ThemeDictionaries>
        <ResourceDictionary x:Key="Light">
            <SolidColorBrush x:Key="ButtonBorderBrush" Color="Blue"/>
            <SolidColorBrush x:Key="ButtonBorderBrushPointerOver" Color="Pink"/>
            <SolidColorBrush x:Key="ButtonBorderBrushPressed" Color="Yellow"/>
        </ResourceDictionary>
    </ResourceDictionary.ThemeDictionaries>
    
    <Style x:Key="ButtonStyle" TargetType="Button">
        <Setter Property="BorderThickness" Value="2" />
    </Style>

</ResourceDictionary>
```

app.js

```js

    title: 'BUTTON'
});

var style = Ti.UI.Windows.createStyle({ source: 'ms-appx:///Styles.xaml' });
style.apply(button, 'ButtonStyle');

win.add(button);
win.open();
```

Expected: 

- Button border should be blue by default
- Button border should become pink when pointer hovers it
- Button border should become yellow when you click the button


Source/UI/src/Windows/Style.cpp
```diff
@@ -0,0 +1,93 @@
...
+			}
+
+			TITANIUM_PROPERTY_GETTER(Style, source)
+			{
+				return get_context().CreateString(TitaniumWindows::Utility::ConvertUTF8String(source__));
+			}
+
+			TITANIUM_PROPERTY_SETTER(Style, source)
+			{
+				source__   = TitaniumWindows::Utility::ConvertUTF8String(static_cast<std::string>(argument));
```
ewanharris: I think we should do some validation on the source file passed in, I made a typo in a filename and the error I got was the below, which pointed me to the apply call but it took me a little while to figure out I had made a typo.

```text
[ERROR] ----- Titanium Javascript Runtime Error -----
[ERROR] In app.js: 11,38
[ERROR] Message: Uncaught Error: unknown exception
````

File: apidoc/WindowsOnly/Titanium.UI.Windows.Style.yml [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+description: |
+    Styles let you reuse control properties for a consistent appearance across multiple controls.
+    Use [Titanium.UI.Windows.createStyle()](Titanium.UI.Windows.createStyle) method
+    and use `source` property to indicate where to load XAML template.
+
+        var style = Ti.UI.Windows.createStyle({
+            source: 'ms-appx:///Styles.xaml'
+        });
+
+    In `Styles.xaml`, you can define a theme using `ResourceDictionary.ThemeDictionaries`
```
ewanharris: We might need to also explicitly call out that these should only be referenced by filename, some people might wish to group things under `/styles`, but if you set source to to `styles/LabelStyle.xaml`, we'll error out as we can't find it, using `LabelStyle.xaml` however will load it just fine

infosia: Updated, now `Style` should be able to handle multiple styles across multiple components. For the files under sub directory issue, I'm guessing that the root cause might be that we might not be properly packaging folder structure under `Resources`. I'll investigate it and then create new ticket for it.

Source/UI/src/Windows/Style.cpp
infosia: Updated, now it throws JavaScript exception with property message.

ewanharris:
@infosia 

I've had success with Button and View, but am hitting problems with quite a few others such as Label, SearchBar, ScrollView. I'm writing my xaml in VS and getting autocomplete of the properties so I'm pretty sure they're valid, I've got my code for label attached below. I ran it through VS and I got the error below [here](https://github.com/appcelerator/titanium_mobile_windows/pull/1066/files#diff-49729a9ee6c50bdd6d573c9ab387210fR84) (when running normally I just get unknown exception), it looks like component is pulling the border element we create and not the TextBox element?

```
WinRT information: Cannot apply a Style with TargetType 'Windows.UI.Xaml.Controls.TextBox' to an object of type 'Windows.UI.Xaml.Controls.Border'.
```

```js
var lbl = Ti.UI.createLabel({
    text: 'label'
})
var style = Ti.UI.Windows.createStyle({ source: 'ms-appx:///Styles.xaml' });
style.apply(lbl, 'LabelStyle');
win.open();

```

```xml
<ResourceDictionary
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" 
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
    <Style x:Key="LabelStyle" TargetType="TextBox">
        <Setter Property="Foreground" Value="Plum"></Setter>
    </Style>
</ResourceDictionary>
```

infosia:
@ewanharris Nice, it was a bit tricky on Titanium...Should be addressed now;

```js
var lbl = Ti.UI.createLabel({
    text: 'label'
})
var style = Ti.UI.Windows.createStyle({ source: 'ms-appx:///Styles.xaml' });
style.apply(lbl, 'LabelStyle');

win.add(lbl);
win.open();
``` 

```xml
<ResourceDictionary
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" 
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
    <Style x:Key="LabelStyle" TargetType="TextBlock">
        <Setter Property="Foreground" Value="Plum"></Setter>
    </Style>
</ResourceDictionary>
```

ewanharris:
For testing check out https://gist.github.com/ewanharris/21b051884a6483db770514c498c19c92

@infosia  I just had a thought about the usage of this in alloy, currently I think it would be required for users to be calling apply themselves correct? Would it be possible to remove the apply call and set it directly on the element at creation? Then I think a user could do something like `<Button style='MyButtonStyle'/>`?

infosia:
@ewanharris I think `<Button style='MyButtonStyle'/>` is a good use case. Would you create new ticket for it? I think it's possible by adding some glue code, it would be something like...

(pseudo code)

```js
View.prototype.setStyle = function(name) {
    var style = Ti.UI.Windows.createStyle(); // This loads "ms-appx:///Styles.xaml" by default
    style.apply(this, name);
};
```

We might need to think about if `style` is a good property name for this new feature though. Isn't it too generic? My first PoC was actually just like above, but I did not want to introduce new feature for common property name.

PR #1073, opened 2017-08-14
[TIMOB-25166] Fix issue where VS2017 could not use 15063 sdk

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-25166

Allow VS 2017 to build using 10.0.15063

Verification

1. Using VS 2017 only install 10.0.15063 (or select it using --win-sdk)
    - It should build fine


ewanharris:
@infosia  sorry I should have marked this as do not merge originally, I was not seeing any issues from a quick smoke on my dirtier machine so made this PR, I've just tried this again on the machine I've just setup and I'm hitting some issues. I'll try and take a further look

ewanharris:
Turns out my error was just because I didn't install everything we require 😊 For VS2017 I think we're good to support that SDK version, app building, module building and hyperloop usage are all working fine with only 10.0.15063

PR #1075, opened 2017-08-15
[TIMOB-25072] Add waitForResponse() to wait for a response from server

infosia:
[TIMOB-25072](https://jira.appcelerator.org/browse/TIMOB-25072)
[TIMOB-24893](https://jira.appcelerator.org/browse/TIMOB-24893)

Introduce new API only for Windows: `Ti.Network.HTTPClient._waitForResponse` to make sure to wait for server response in order to get `responseText` and `readyState` in sync. This API is designed for **LiveView**, which uses `while` loop to wait for response, that actually screwed networking operation on Windows.

```js
var win = Ti.UI.createWindow({ backgroundColor: 'gray' }),
    btn = Ti.UI.createButton({
        title: 'HTTP GET'
    });

    var request = Ti.Network.createHTTPClient({
        onload: function () {
            alert('onload:\n\treadyState: ' + request.readyState);
        },
        onerror: function (e) {
            alert('onerror:\n\treadyState: ' + request.readyState + '\n\terror: ' + e.error);
        }
    });

    request.open('GET', 'http://www.axway.com/');
    request.send();

    try {
        request._waitForResponse();
        console.log(request.responseText);
        win.backgroundColor = 'green';
    } catch (E) {
        alert(JSON.stringify(E));
    }

});
win.add(btn);
win.open();
```

So with this, LiveView now able to wait for server response _without waiting for it in `while` loop_.

```js
//
// Windows only API: _waitForResponse() waits for the response from the server.
//
if (this.platform == 'windows' && request._waitForResponse) {
	request._waitForResponse();
	if (request.readyState === 4 || request.status === 404) {
		rsp = request.status === 200 ? request.responseText : false;
	} else {
		throw new Error('[LiveView] File Server unavailable. Host Unreachable @ ' + Module._url + ':' + Module._port + '\n[LiveView] Please ensure your device and computer are on the same network and the port is not blocked.');
	}
	done = true;	
}
```


garymathews:
I thought this was fixed by https://github.com/appcelerator/titanium_mobile_windows/pull/791, is this a regression?

Is there a way to not have to implement a separate method? I think the behaviour should be the same as iOS and Android for this. I've seen customers use `while` loops like this (although a bad practice)

infosia:
@garymathews #791 looked like it fixed the issue on the classic app, but on alloy app it still crashed ([TIMOB-25072](https://jira.appcelerator.org/browse/TIMOB-25072)). As far as I can see Windows app tends to crash when application does not response some time, and waiting for network response in `while` loop in UI thread is the case that crashes app. Yes I agree that we hope to keep same behaviour between platforms but I think this is the safest way to avoid crash in LiveView.

ewanharris:
@infosia Will this be an undocumented API solely intended for usage in LiveView similar to `Ti.App._restart`, or will we support this being used by developers?

infosia:
@ewanharris I would think this is a private API especially designed for LiveView like `Ti.App._restart`.

ewanharris:
@infosia Grand, checking the yml produced from the build this is being documented. Could you possible add a TITANIUM_FUNCTION_UNIMPLEMENTED macro for this? Or is there a way we can hide it from the docs?

garymathews:
@infosia We should make it `Ti.Network.HTTPClient._waitForResponse`

infosia:
@garymathews @ewanharris  ok, renamed and also marked "unimplemented". updated [liveview pr](https://github.com/appcelerator/liveview/pull/106) too.

PR #1076, opened 2017-08-16
[TIMOB-25064] Ability to use 3rd party .NET managed DLL with Hyperloop

infosia:
[TIMOB-25064](https://jira.appcelerator.org/browse/TIMOB-25064)

Ability to use 3rd party .NET managed DLL with Hyperloop.

**Steps to Test**

1. Create .NET managed DLL 

Create _Universal Windows Class Library_ with Visual C#. It should output only `.dll` other than `.winmd`. You can skip this step by using prebuilt [ManagedClassLibrary1.dll](https://jira.appcelerator.org/secure/attachment/63087/ManagedClassLibrary1.dll).

```cs
using System;

namespace ManagedClassLibrary1
{
    public class Test
    {
        public Test()
        {

        }
        public string Hello()
        {
            return "Hello, World!";
        }
    }
}
```

2. Copy your `ManagedClassLibrary1.dll` file into `lib/windows/win10/x86/` under  your Titanium app project folder.

**app.js**

```js
    Test = require('ManagedClassLibrary1.Test'),
    test = new Test();

win.addEventListener('open', function() {
    alert(test.Hello());
});

win.open();
```

3. Place `appc.windows.js` directly under your Titanium project folder. Make sure it contains names of your class library component under `hyperloop.windows.thirdparty` key. It should look like below.

**appc.windows.js**

```js
/**
 * Hyperloop configuration
 */
module.exports = {
	type: 'app',
	group: 'titanium',
	dependencies: {},
	hyperloop: {
		windows:
		{
			/**
			 * Optionally, you can bring in third-party or first-party Windows Runtime components.
			 * The 'key' is the root namespace of the component that will be used in the require.
			 * There's no values for now but we reserve them for future release.
			 *
			 * Place the library files into the lib/windows/(phone|store|win10)/(x86|ARM)/ folder under your app project.
			 * Hyperloop will pick up the winmd files and will generate necessary bindings and include the winmd in your app. 
			 */
			thirdparty: {
				'ManagedClassLibrary1': {}
			}
		}
	}
};
```
4. Make sure to enable hyperloop module and hyperloop plugin. Don't forget to enable `deployment-targets` for windows.

```xml
  <modules>
    <module platform="windows" version="2.2.0">hyperloop</module>
  </modules>
  <deployment-targets>
    <target device="windows">true</target>
  </deployment-targets>
    <plugins>
    <plugin version="2.2.0">hyperloop</plugin>
  </plugins>
```

5.Run it using `appc run -p windows --wp-sdk 10.0 --target ws-local -l trace`.
6. Then you should see "Hello, World!" dialog box in your Titanium app.

### DIFF
--- a/cli/commands/_build/generate.js
+++ b/cli/commands/_build/generate.js
@@ -280,6 +280,7 @@ function generateCmakeList(next) {
 			defaultLanguage: this.defaultLanguage,
 			native_modules: native_modules,
 			targetPlatformSdkMinVersion: this.targetPlatformSdkMinVersion,
+			thirdPartyReferences: this.hyperloopConfig.windows.thirdPartyReferences || [],
 			vsSdkReferences: this.vsSdkReferences || []
 		}
 	), next);
--- a/templates/build/CMakeLists.txt.ejs
+++ b/templates/build/CMakeLists.txt.ejs
@@ -85,6 +85,11 @@ set(SOURCE_Assets
   ${CMAKE_CURRENT_BINARY_DIR}/${APP_MANIFEST_NAME}
   lib/TitaniumKit.dll
   lib/HAL.dll
+<% for(var i = 0; i < thirdPartyReferences.length; i++) { -%>
+<% if (thirdPartyReferences[i].ContentPath) { -%>
+  "<%= thirdPartyReferences[i].ContentPath.replace(/\\/g, '/') %>"
+<% } -%>
+<% } -%>
   # Include the "." assets here!
 <% for(var i = 0; i < sourceGroups['.'].length; i++) { -%>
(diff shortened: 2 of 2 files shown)


ewanharris:
@infosia Shouldn't this belong in the Hyperloop plugin? Apologies if I'm misunderstanding here as I'm not massively familiar with the Hyperloop integration

infosia:
>  Shouldn't this belong in the Hyperloop plugin?

@ewanharris That's nice catch, yes it should wherever possible. I found that some of the logic needs to remain on the Titanium SDK template still, but I think most of the code can be moved to hyperloop plugin. I'll work on it.

PR #1077, opened 2017-08-17
[TIMOB-25031] Respect targetSdkVersion in timodule.xml

infosia:
[TIMOB-25031](https://jira.appcelerator.org/browse/TIMOB-25031)

Since Visual Studio 2015 doesn't support latest Windows 10 SDK 10.0.15063,
we need a way to specify Windows SDK version when building module. We could leverage [TIMOB-23557](https://jira.appcelerator.org/browse/TIMOB-23557) to specify target sdk version in `timodule.xml`.

1.  Create module

```
appc new -t timodule -n testModule --id com.test.module
```

2. Building with VS 2015 + 10.0.14393.0

```
appc ti build -p windows --build-only -V 14.0 --run-cmake
```

**timodule.xml**

``` XML
<windows>
  <manifest>
    <uses-sdk targetSdkVersion="10.0.14393.0" />
  </manifest>
</windows>
```


ewanharris:
@infosia I think we need to move the run-cmake plugin into the module build process itself. Once a user creates a module we have no mechanism in place to update it, so in order for a customer to get this fix they must copy the files over from the SDK or this repo which isn't a great developer experience. 

That said, most existing modules are going to have this hook if we move it into the module build  then when a user uses `--run-cmake` the module build process **and** the hook will both with, with the possibility of the hook failing out. I guess we'd need a step to look for this hook, and if it exists just error out, telling the user to remove it and rebuild once they've done that.

What do you think?

infosia:
@ewanharris I think it's a good idea. We can use this opportunity to fix bad design. 👍

infosia:
@ewanharris Moved entire plugin logic into module build itself. Module build errors out when "old" plugin is detected for now, [please review the error message](https://github.com/appcelerator/titanium_mobile_windows/pull/1077/files#diff-0d6b2feefcc27b081ba23dd5b2464a6cR188)

cli/commands/_buildModule.js
```diff
@@ -442,6 +494,95 @@ WindowsModuleBuilder.prototype.packageZip = function packageZip(next) {
...
+    } else if (version == '14.0') {
+        return 'Visual Studio 14 2015';
+    } else if (/^Visual Studio \w+ 2017/.test(version)) {
+        return 'Visual Studio 15 2017';
+    } else {
+        return 'Visual Studio 14 2015';
+    }
+}
+
+function rmdir(dirPath, fs, path, logger, removeSelf) {
```
ewanharris: Now that this is inside the module builds we can remove this rmdir function as we have access to the SDKs node_modules and subsequently wrench.rmdirsyncrecursive, not an urgent issue that must be resolved before merge just some cleanup to do at a later date

cli/commands/_buildModule.js
```diff
@@ -442,6 +494,95 @@ WindowsModuleBuilder.prototype.packageZip = function packageZip(next) {
 
 };
 
+function selectVisualStudio(data) {
```
ewanharris: We maintain this function in a few different places if I'm remembering right, it would be nice to consolidate them into one place (there's probably some other bits that could be pulled into a common utils I'm guessing)

PR #1080, opened 2017-08-18
Ensure we select the SDK after installing it

ewanharris:
So I was checking out the failure for #1077 and was like huh it's using 6.2.0 `Error: ENOENT: no such file or directory, open 'C:\ProgramData\Titanium\mobilesdk\win32\6.2.0.v20170816173122\windows\lib\TitaniumKit\phone\x86\TitaniumKit.dll'` and was intrigued as to why, it's because if an SDK is already installed the `-d` flag in the sdk install command isn't respected 😱, that's a bug in titanium cli but for now we need to workaround it and this is how we do that


ewanharris:
Once this gets merged and approved I'll backport to 6_2_X

PR #1084, opened 2017-08-23
[TIMOB-25175] ListView.height for Ti.UI.SIZE uses row height

infosia:
[TIMOB-25175](https://jira.appcelerator.org/browse/TIMOB-25175)

I'm not quite actually sure why ListView height is set to row height with Ti.UI.SIZE, but I observed that the ListView sets right value for initial value and latter was wrong. I suppress this behavior by just using initial value only for height vaue.

```js

var myTemplate = {
    childTemplates: [
        {
            type: 'Ti.UI.View',
            bindId: 'itemContainer',
            properties: {
                width: Ti.UI.FILL,
                height: 50,
                backgroundColor: 'green'
            },
            childTemplates: [
                {
                    type: 'Ti.UI.Label',
                    bindId: 'info',
                    properties: {
                        width: Ti.UI.FILL,
                        height: Ti.UI.SIZE,
                        backgroundColor: 'pink'
                    }
                }
            ]
        },
    ]
};

var listView = Ti.UI.createListView({
    templates: { 'template': myTemplate },
    defaultItemTemplate: 'template',
    width:  Ti.UI.FILL,
    height: Ti.UI.SIZE,
});
var sections = [];

var fruitSection = Ti.UI.createListSection({ headerTitle: 'Fruits' });
var fruitDataSet = [
    { info: { text: 'Apple 1'  }},
    { info: { text: 'Banana 1' } },
];
fruitSection.setItems(fruitDataSet);
sections.push(fruitSection);
listView.setSections(sections);

win.add(listView);
win.open();
```

Expected: All row content should be shown.


infosia:
Based on the discussions in [TIMOB-25175](https://jira.appcelerator.org/browse/TIMOB-25175) it seems depending on screen size (and default Window size at launch on desktop) this still doesn't resize well. re-opening the original ticket.

PR #1088, opened 2017-08-30
Prevent building to 8.1 targets, and stop producing 8.1 artifacts

ewanharris:
Apologies in advance for the wall of text 😅, I greatly appreciate any input on this PR whether it be around the wording, behaviour or any other parts of it.

### Summary
This PR does the following

- Prevents building app projects to 8.1 targets, or when specifying 8.1 as `--win-sdk` SDK
    - `--win-sdk 8.1` will error with `[ERROR] Invalid "--win-sdk" value "8.1"`, prompting the user to select another version
    - With an 8.1 device plugged in `-T wp-device` will error with `[ERROR] The connected device is running 8.1, which is unsupported by Titanium SDK 7.0.0.v20170828180637`
    - When specifying TargetPlatformVersion or TargetPlatformMinVersion to 8.1 will error out [TIMOB-24704](https://jira.appcelerator.org/browse/TIMOB-24704), is that even a supported config? Setting to 10.0 is too vague imo
	- I still need to test building to `-T ws-local` on an Windows 8.1 install

- Prevents building 8.1 artifacts for module projects
    - When creating a module only Windows 10 projects are generated
    - When building with `--run-cmake` only Windows 10 projects are created
    - When specifying `<uses-sdk targetSdkVersion="8.1" />` in the timodule.xml the build process will error with `[ERROR] The specified version of the Windows and Windows Phone SDK "8.1" is not supported by Titanium 7.0.0.v20170828180637`

- When used with appcelerator/windowslib/pull/78 and running ti info -t windows
	- 8.1 SDK will be shown, but listed as not supported `8.1 **Not supported by Titanium SDK 7.0.0.v20170828180637**`
	- 8.1 emulators will not be shown (and subsequently Studio does not show them)
	- 8.1 devices will be shown (and subsequently Studio will show them)

-	The above is based on my understanding and observation of how we behave on other platforms, any input is greatly appreciated on error messages, behaviour etc.

- Stops producing 8.1 artifacts during the build process

This PR does not do the following

- Remove any code under Source\, cli\ or Tools\ relevant to Windows 8.1 or Windows Phone 8.1 (other then necessary for the above)
	- This was deemed a cleanup task that is not required for removal
- Make any edits to documentation in this repo
    - Again, will be a cleanup task
- Drop support for anything else i.e. Visual Studio 2013
    - Although it kind of does as that version only supports 8.1, there is no explicit removal of support. Although in my opinion we should do it in the same release

Tickets:

https://jira.appcelerator.org/browse/TIMOB-25207
https://jira.appcelerator.org/browse/TIMOB-25203

### DIFF
--- a/cli/commands/_build/config/deviceID.js
+++ b/cli/commands/_build/config/deviceID.js
@@ -27,8 +27,11 @@ module.exports = function configOptionDeviceID(order) {
 			 (cli.argv.target === 'wp-device' && value === 'de')) && devices[0]) {
 
 			// if win-sdk is not specified, use wpsdk for device
-			if (devices[0].wpsdk && !this.isWindowsSDKTargetSpecified()) {
+			if (devices[0].wpsdk && !this.isWindowsSDKTargetSpecified() && appc.version.satisfies(devices[0].wpsdk, this.packageJson.vendorDependencies['windows phone sdk'], false)) {
 				cli.argv['win-sdk'] = devices[0].wpsdk;
+			} else {
+				this.logger.error(__('The connected device is running %s, which is unsupported by Titanium SDK %s', devices[0].wpsdk, this.cli.sdk.name));
+				process.exit(0)
 			}
 			return callback(null, devices[0].udid);
 		}
@@ -47,8 +50,11 @@ module.exports = function configOptionDeviceID(order) {
 		}
 
 		// if win-sdk is not specified, use wpsdk specified for device
-		if (dev.wpsdk && !this.isWindowsSDKTargetSpecified()) {
+		if (dev.wpsdk && !this.isWindowsSDKTargetSpecified() && appc.version.satisfies(dev.wpsdk, this.packageJson.vendorDependencies['windows phone sdk'], false)) {
 			cli.argv['win-sdk'] = dev.wpsdk;
+		} else {
+			this.logger.error(__('The connected device is running %s, which is unsupported by Titanium SDK %s', dev.wpsdk, this.cli.sdk.name));
+			process.exit(10)
 		}
 
 		// check the device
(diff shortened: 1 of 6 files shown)


cli/commands/_build/config/deviceID.js
```diff
@@ -27,8 +27,11 @@ module.exports = function configOptionDeviceID(order) {
 			 (cli.argv.target === 'wp-device' && value === 'de')) && devices[0]) {
 
 			// if win-sdk is not specified, use wpsdk for device
-			if (devices[0].wpsdk && !this.isWindowsSDKTargetSpecified()) {
+			if (devices[0].wpsdk && !this.isWindowsSDKTargetSpecified() && appc.version.satisfies(devices[0].wpsdk, this.packageJson.vendorDependencies['windows phone sdk'], false)) {
 				cli.argv['win-sdk'] = devices[0].wpsdk;
+			} else {
+				this.logger.error(__('The connected device is running %s, which is unsupported by Titanium SDK %s', devices[0].wpsdk, this.cli.sdk.name));
+				process.exit(0)
```
ewanharris: We have to exit ourselves here and on line 57, calling the callback with an error will just keep us going round in an infinite loop. I kept the exit code 0 as it is in the ios build [here](https://github.com/appcelerator/titanium_mobile/blob/9e9a0b55ece14d963f3d3943098c93d26a1a4e03/iphone/cli/commands/_build.js#L2023-L2040)

infosia:
@sgtcoolguy Could you take a chance to review this? I'm wondering if we have more things to do to remove 8.1 support in backend, such as removing integration with Jenkins build artifact and our nightly build server?

sgtcoolguy:
@infosia The Jenkinsfile changes look OK to me, though it appears the build failed trying to install a 'master' branch SDK build (it uses a pre-built full SDK to use as a base and then copies the locally built Windows portion into it). Not sure why that is happening...

You basically just have to remove the sections that build and archive the 8.1 artifacts, which it looks like has been done here.

ewanharris:
Closing in favour of #1132

PR #1091, opened 2017-09-01
[TIMOB-25216] Docs do not show ti.cloud as supported

ewanharris:
- Add the skeleton of the ti.cloud module docs
- Archive both the Modules and WindowsOnly folders in the Jenkins jobs
- Update latest docs

Couple of questions that this has made me think of

1. Can we class `createX509CertificatePinningSecurityManager`, as supported on Windows given that we don't have a https module?
2. Our docs state that lifecycleContainer is supported on Windows, it looks to be implemented in https://github.com/appcelerator/titanium_mobile_windows/blob/3f1967e80db8a70f363f78661ce6665558709245/Source/TitaniumKit/src/Module.cpp#L276-L299, but it's an Android specific API? Does it provide any value on Windows? Should it be marked as unimplemented?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -74,8 +74,12 @@ timestamps {
 			echo 'copying generated docs to dist folder'
 			if (isUnix()) {
 				sh 'mv apidoc/Titanium dist/windows/doc/Titanium'
+				sh 'mv apidoc/Modules dist/windows/doc/Modules'
+				sh 'mv apidoc/WindowsOnly dist/windows/doc/Windowsonly'
 			} else {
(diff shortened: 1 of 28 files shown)


infosia:
@ewanharris 

- I think we don't support`createX509CertificatePinningSecurityManager` because I expect it doesn't work.
- We happened to implement `lifecycleContainer` on Windows but I believe it is not actually used from anywhere. I would leave it as it is for now unless it causes any harm.

ewanharris:
Marking as do not merge while the automated suite is updated so we can be sure of marking APIs as definitely supported

PR #1092, opened 2017-09-04
Use shared mocha suite

ewanharris:
This is the start of the (probably long) process of using the shared mocha suite like in titanium_mobile, I'm making a PR to help me test my steps and also to speed up the review process, same as #1088 drive by reviews more than welcome.

It is currently based off the drop_it branch from my repo (see #1088) as this change will _only_ land on master after that has been merged

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -182,6 +182,7 @@ timestamps {
 
 		parallel(
 			'ws-local unit tests': unitTests('ws-local', targetBranch, 'TIMOB-24816'),
+			'wp-emulator unit tests': unitTests('wp-emulator', targetBranch, 'TIMOB-24816'),
 			failFast: true
 		)
 	}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,43 +91,6 @@ timestamps {
 		}
 
 		parallel(
-			'Windows 8.1 Store x86': {
-				node('msbuild-12 && (vs2013 || vs2015) && hyper-v && windows-sdk-8.1 && npm && node && cmake && jsc') {
-					build('8.1', '12.0', 'WindowsStore-x86', gitCommit)
-
-					unstash 'NMocha' // for tests
-					dir('Tools/Scripts/build') {
-						timeout(testTimeout) {
-							echo 'Running Tests on Windows 8.1 Desktop'
-							bat "node test.js -s 8.1 -T ws-local -p Windows8_1.Store -b ${targetBranch}"
-						}
-						// Kill the desktop app, so workspace cleanup works...
-						bat 'taskkill /IM Mocha.exe /F'
-					}
-					junit 'dist/junit_report.xml'
-				}
-			},
-			'Windows 8.1 Phone x86': {
-				node('msbuild-12 && (vs2013 || vs2015) && hyper-v && windows-sdk-8.1 && npm && node && cmake && jsc') {
-					build('8.1', '12.0', 'WindowsPhone-x86', gitCommit)
-
-					unstash 'NMocha' // for tests
-					dir('Tools/Scripts/build') {
-						timeout(testTimeout) {
-							echo 'Running Tests on Windows 8.1 Phone Emulator'
-							bat "node test.js -s 8.1 -T wp-emulator -p Windows8_1.Phone -b ${targetBranch}"
-						}
-						// Kill the phone emulator, so workspace cleanup works...
-						bat 'taskkill /IM xde.exe'
-					}
-					junit 'dist/junit_report.xml'
-				}
-			},
-			'Windows 8.1 Phone ARM': {
-				node('msbuild-12 && (vs2013 || vs2015) && windows-sdk-8.1 && npm && node && cmake && jsc') {
-					build('8.1', '12.0', 'WindowsPhone-ARM', gitCommit)
-				}
-			},
 			'Windows 10 x86': {
 				node('msbuild-14 && vs2015 && hyper-v && windows-sdk-10 && npm && node && cmake && jsc') {
 					build('10.0', '14.0', 'WindowsStore-x86', gitCommit)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,6 @@
 
 def gitCommit = ''
 def testTimeout = 25
-def defaultEmulatorID = '10-0-3'
 
 def build(sdkVersion, msBuildVersion, architecture, gitCommit) {
 	unstash 'sources' // for build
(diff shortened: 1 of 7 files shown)


ewanharris:
Never been happy to see a build failed because of test failures on Jenkins before 😄

Jenkinsfile
```diff
@@ -91,76 +134,55 @@ timestamps {
...
-					dir('Tools/Scripts/build') {
-						timeout(testTimeout) {
-							echo 'Running Tests on Windows 10 Desktop'
-							bat "node test.js -s 10.0 -T ws-local -p Windows10.Store -b ${targetBranch}"
-						}
-						// Kill the desktop app, so workspace cleanup works...
-						bat 'taskkill /IM Mocha.exe /F'
-					}
-					junit 'dist/junit_report.xml'
+					// unstash 'NMocha' // for tests
```
sgtcoolguy: No need to stash the NMocha stuff anymore (line 100) if the common test suite is what we're going to use now. Remove all the commented out stuff here, remove the stashing of the Mocha suite in line 100.

Jenkinsfile
```diff
@@ -91,76 +134,55 @@ timestamps {
...
+			// 	node('msbuild-14 && vs2015 && hyper-v && windows-sdk-10 && npm && node && cmake && jsc') {
+			// 		build('10.0', '14.0', 'WindowsStore-ARM', gitCommit)
+			// 	}
+			// },
+			failFast: true
+		)
+	} // Stage build
+
+	stage('test') {
+		def targetBranch = env.CHANGE_TARGET // if it's a PR, use target merge branch as branch of SDK to install
```
sgtcoolguy: Don't need to re-define this stuff here. Juts move the definition of these variables up to a higher scope before the build stage (Basically just cut lines 128-134 and paste them above the `stage('Build') {` line).

Jenkinsfile
```diff
@@ -91,76 +134,55 @@ timestamps {
...
+		if (!env.BRANCH_NAME.startsWith('PR-')) {
+			targetBranch = env.BRANCH_NAME // if it isn't a PR, try to match the current branch
+		}
+		if (!targetBranch) { // if all else fails, use master as SDK branch to test with
+			targetBranch = 'master'
+		}
+
+		parallel(
```
sgtcoolguy: So the wrinkle here is that we can't guarantee that the same nodes that built the SDK will run the tests since you have them in separate parallel blocks. You can move running the unit tests to happen after the build on each node in the build stage above (so you know the build and test will always be on the same node).

Or you need to basically have the `unitTests` method do an `unarchive` call first to make sure it basically copies the built SDK onto the node. See my comments in that section...

Jenkinsfile
```diff
@@ -33,6 +34,48 @@ def build(sdkVersion, msBuildVersion, architecture, gitCommit) {
 	archiveArtifacts artifacts: 'dist/**/*'
 }
 
+def unitTests(target, branch, testSuiteBranch) {
+	node('msbuild-14 && vs2015 && hyper-v && windows-sdk-10 && npm && node && cmake && jsc') {
+
+		dir('Tools/Scripts/build') {
+			echo 'Setting up SDK'
+			bat 'dir'
+			bat "node setupSDK.js --branch ${branch}"
```
sgtcoolguy: There's no guarantee that there's a distilled dir with the built SDK here yet since the build and test stages are separate parallel blocks with potentially separate nodes (and maybe even separate workspace if they are the same node!)

You need to explicitly call `unarchive` to have it copy the "archived" SDK that was built in the build stage across multiple nodes.

So around line 39 here, add: `unarchive mapping: ['dist/' : '.']`

This will 'unarchive' the saved `dist` directory into the current repo root.

Jenkinsfile
```diff
@@ -33,6 +33,51 @@ def build(sdkVersion, msBuildVersion, architecture, gitCommit) {
...
+		}
+
+
+		// if our test suite already exists, delete it
+		bat 'if exist titanium-mobile-mocha-suite rmdir titanium-mobile-mocha-suite /Q /S'
+		// clone the tests suite fresh
+		// FIXME Clone once on initial node and use stash/unstash to ensure all OSes use exact same checkout revision
+		dir('titanium-mobile-mocha-suite') {
+			// TODO Do a shallow clone, using same credentials as from scm object
+			git changelog: false, poll: false, credentialsId: 'd05dad3c-d7f9-4c65-9cb6-19fef98fc440', url: 'https://github.com/ewanharris/titanium-mobile-mocha-suite.git', branch: testSuiteBranch
```
sgtcoolguy: So at this point, I think we should just set up a working branch that both of us can commit to on the common suite. I'll create a "windows" branch and we can iterate on changes there and just have this one PR point at it for now. Then when it's settled, we can merge that to master branch of the common suite and update this PR to point there.

sgtcoolguy:
```
git changelog: false, poll: false, credentialsId: 'd05dad3c-d7f9-4c65-9cb6-19fef98fc440', url: 'https://github.com/appcelerator/titanium-mobile-mocha-suite.git', branch: testSuiteBranch
```

Jenkinsfile
```diff
@@ -81,86 +125,34 @@ timestamps {
...
-					junit 'dist/junit_report.xml'
-				}
-			},
-			'Windows 10 ARM': {
-				node('msbuild-14 && vs2015 && hyper-v && windows-sdk-10 && npm && node && cmake && jsc') {
-					build('10.0', '14.0', 'WindowsStore-ARM', gitCommit)
-				}
-			},
+	stage('test') {
+		parallel(
```
sgtcoolguy: hardcode test suite branch to 'windows' now...

Jenkinsfile
```diff
@@ -81,86 +125,34 @@ timestamps {
...
-							echo 'Running Tests on Windows 10 Desktop'
-							bat "node test.js -s 10.0 -T ws-local -p Windows10.Store -b ${targetBranch}"
-						}
-						// Kill the desktop app, so workspace cleanup works...
-						bat 'taskkill /IM Mocha.exe /F'
-					}
-					junit 'dist/junit_report.xml'
-					// Delete the report from store, so if phone fails we don't pick this one up
-					bat 'del /f /q dist\\junit_report.xml'
+	// stage ('Build') {
```
sgtcoolguy: why is the build stage now commented out?

Jenkinsfile
```diff
@@ -33,6 +33,51 @@ def build(sdkVersion, msBuildVersion, architecture, gitCommit) {
 	archiveArtifacts artifacts: 'dist/**/*'
 }
 
+def unitTests(target, branch, testSuiteBranch) {
+	node('msbuild-14 && vs2015 && hyper-v && windows-sdk-10 && npm && node && cmake && jsc && Win-Gin10') {
+		def defaultEmulatorID = '10-0-1'
+		// unarchive mapping: ['dist/' : '.']
```
sgtcoolguy: why not unarchiving anymore? You need to ensure the built artifacts are on this node, since it may not be the same node that did the build stage.

ewanharris:
Will be worked on in https://github.com/appcelerator/titanium_mobile_windows/tree/TIMOB-25019

sgtcoolguy:
Closing since we've got this working now. Opening a new PR in its place...

PR #1112, opened 2017-09-18
[TIMOB-25291] Fix upgrade of GET with data to POST

ewanharris:
JIRA: https://jira.appcelerator.org/browse/TIMOB-25291

When performing a GET request and calling send with data the request used to be upgraded to a POST, this doesn't occur with other platforms.
<details>
<summary>Demo code</summary>

```js
    title: 'stuff',
    backgroundColor: '#fff'
});

    text: 'click me',
    color: "#333",
    font: {
        fontSize: 20
    }
});

label.addEventListener('click', function() {
    var url = 'https://httpbin.org/anything';
        onload : function(e) {
            alert("method was: " + JSON.parse(this.responseText).method);
            Ti.API.info(this.responseText);
        },
        onerror : function(e) {
            Ti.API.debug(e.error);
            alert('error');
        },
        timeout : 5000
    });
    client.open("GET", url);
    client.send({foo:'bar'});
})

win.add(label);
win.open()
```
</details>

### Verification

1. Add the code above to existing app.js and build
2. Tap the `click me` label
3. Method should be GET

### DIFF
--- a/Source/Network/src/HTTPClient.cpp
+++ b/Source/Network/src/HTTPClient.cpp
@@ -130,12 +130,6 @@ namespace TitaniumWindows
 
 		void HTTPClient::send(const std::map<std::string, JSValue>& postDataPairs, const bool& useMultipartForm) TITANIUM_NOEXCEPT
 		{
-			if (method__ == Titanium::Network::RequestMethod::Get) {
-				TITANIUM_MODULE_LOG_WARN("HTTPClient::send: Data found during a GET request. Method will be changed to POST.");
-				method__ = Titanium::Network::RequestMethod::Post;
-			}
-
-
 			Windows::Web::Http::IHttpContent^ postData;
 			if (useMultipartForm) {
 				postData = ref new Windows::Web::Http::HttpMultipartFormDataContent();


ewanharris:
@sgtcoolguy Looking at the history I believe this change was made by you. If you have time, could you take a check to make sure I'm not missing some reasoning for this. I ran through the Network tests and don't see any failures caused by this

The problem this is causing is that [ti.cloud sends data with GETs](https://github.com/appcelerator-modules/ti.cloud/blob/95ac378a2e2b72bcce222a34ddc8f19d1d15de36/commonjs/src/sdk/utils.js#L186-L204), so a few of the APIs there are failing, Android and iOS don't seem to change to POST so I figured it made sense to bring Windows in line with everyone else

garymathews:
The original changes look to be related to [TIMOB-19023](https://jira.appcelerator.org/browse/TIMOB-19023), but it doesn't go into details as to why that specific change was made?

infosia:
ok nice, I can see that the test case in [TIMOB-19023](https://jira.appcelerator.org/browse/TIMOB-19023) is working too. should be good for `in-qe-testing` now.

PR #1116, opened 2017-09-25
[TIMOB-25344] New setTimeout implementation

infosia:
JIRA ticket: [TIMOB-25344](https://jira.appcelerator.org/browse/TIMOB-25344)

According to the studies from [TIMOB-25019](https://github.com/appcelerator/titanium_mobile_windows/commits/TIMOB-25019), we found some pitfalls in current `setTimeout/setInterval` implementations. There may be rare chances to encounter but we want simpler/faster/safer implementation for it. 

- Make sure to clear callback right after callback is actually called from `setTimeout`. 
- Invoke callback immediately when interval equals zero. It had synchronization issue with `clearTimeout/clearInterval`.
- `Xaml::DispatcherTimer` functions should have been called from UI thread. Use `ThreadPoolTimer` when possible, because it doesn't have the issue. 
- With `ThreadPoolTimer`, make sure to invoke callback from UI thread.

I also observed that new implementation is much faster than old implementation. It had `30%` faster (14sec->10sec) on debug mode by simple code like below.

Note that this is about internal implementation, so the API functionalities setTimeout/setInterval/clearTimeout/clearInterval should not be changed.

```js
var label = Ti.UI.createLabel({ text: '0' });

var count = 0;
    var start = +new Date();
    var id = setInterval(function () {
        if (count > 100) {
            clearInterval(id);
            Ti.API.info((+new Date() - start));
            return;
        }
        count++;
        label.text = count;
    }, 100);
});

win.add(label);
win.open();
```

```js
var label = Ti.UI.createLabel({ text: '0' });

var count = 0;
function test() {
    setTimeout(function () {
        count++;
        label.text = count;
        if (count < 100) {
            test();
        }
    }, 100);
}

    test();
});

win.add(label);
win.open();
```

### DIFF
--- a/Source/Global/include/TitaniumWindows/GlobalObject.hpp
+++ b/Source/Global/include/TitaniumWindows/GlobalObject.hpp
@@ -37,6 +37,11 @@ namespace TitaniumWindows
 		GlobalObject& operator=(GlobalObject&&) = default;
 #endif
 
+		virtual unsigned setTimeout(JSObject& function, const std::chrono::milliseconds& delay) TITANIUM_NOEXCEPT override;
+		virtual unsigned setInterval(JSObject& function, const std::chrono::milliseconds& delay) TITANIUM_NOEXCEPT override;
+		virtual void clearTimeout(const unsigned& timerId) TITANIUM_NOEXCEPT override;
+		virtual void clearInterval(const unsigned& timerId) TITANIUM_NOEXCEPT override;
+
 		static void JSExportInitialize();
 		void setSeed(::Platform::String^ seed);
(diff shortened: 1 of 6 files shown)


sgtcoolguy:
Do we want to go forward with this? We ultimately found that we needed the extended execution API found in #1117 to fix the tests. We may want to hold off on this if it's not really necessary...

infosia:
That sounds fair. I would keep this open with "do not merge" label to get more community feedback.

infosia:
Closing this, as current Timer implementation doesn't actually cause issues.

PR #1117, opened 2017-09-26
[TIMOB-25345] Implement extended execution to prevent suspension

infosia:
[TIMOB-25345](https://jira.appcelerator.org/browse/TIMOB-25345)

```js

    Ti.UI.Windows.beginExtendedExecution();

    setTimeout(function () {
        Ti.UI.Windows.endExtendedExecution();
    }, 10000);
});

win.open();
```

Note: It seems this is working good on Jenkins server. But from what I observed, desktop app is not suspended even when you lock the screen...so you might not have a good way to test this. Just check if it does not throw exception when you test this.

### DIFF
--- a/Source/Ti/src/App/WindowsModule.cpp
+++ b/Source/Ti/src/App/WindowsModule.cpp
@@ -108,7 +108,7 @@ namespace TitaniumWindows
 #if defined(IS_WINDOWS_10)
 			using namespace Windows::ApplicationModel::ExtendedExecution;
 
-			ENSURE_OPTIONAL_INT_AT_INDEX(reason, 0, 0);
+			ENSURE_OPTIONAL_INT_AT_INDEX(reason, 0, static_cast<std::uint32_t>(Windows::ApplicationModel::ExtendedExecution::ExtendedExecutionReason::Unspecified));
 			ENSURE_OPTIONAL_OBJECT_AT_INDEX(result_callback, 1);
 			ENSURE_OPTIONAL_OBJECT_AT_INDEX(revoked_callback, 2);
 
--- a/Source/Ti/src/App/WindowsModule.cpp
+++ b/Source/Ti/src/App/WindowsModule.cpp
@@ -108,11 +108,21 @@ namespace TitaniumWindows
 #if defined(IS_WINDOWS_10)
 			using namespace Windows::ApplicationModel::ExtendedExecution;
 
-			ENSURE_OPTIONAL_INT_AT_INDEX(reason, 0, static_cast<std::uint32_t>(Windows::ApplicationModel::ExtendedExecution::ExtendedExecutionReason::Unspecified));
-			ENSURE_OPTIONAL_OBJECT_AT_INDEX(result_callback, 1);
-			ENSURE_OPTIONAL_OBJECT_AT_INDEX(revoked_callback, 2);
+			ENSURE_OPTIONAL_OBJECT_AT_INDEX(extendedExecutionArgs, 0);
+			auto reason = ExtendedExecutionReason::Unspecified;
+			if (extendedExecutionArgs.HasProperty("reason")) {
+				reason = static_cast<ExtendedExecutionReason>(static_cast<std::uint32_t>(extendedExecutionArgs.GetProperty("reason")));
+			}
+			auto result_callback = get_context().CreateObject();
+			if (extendedExecutionArgs.HasProperty("result")) {
+				result_callback = static_cast<JSObject>(extendedExecutionArgs.GetProperty("result"));
+			}
+			auto revoked_callback = get_context().CreateObject();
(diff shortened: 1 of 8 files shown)


Source/UI/src/WindowsModule.cpp
```diff
@@ -127,5 +134,65 @@ namespace TitaniumWindows
...
+#if defined(IS_WINDOWS_10)
+			TITANIUM_EXCEPTION_CATCH_START{
+				endExtendedExecution();
+			} TITANIUM_EXCEPTION_CATCH_END
+#endif
+			return get_context().CreateUndefined();
+		}
+
+#if defined(IS_WINDOWS_10)
+		void WindowsModule::beginExtendedExecution()
```
sgtcoolguy: So, this implementation is working fine for running the unit tests, BUT I would like to see if we can improve the API offered to the developer.

First, this should likely hang off Ti.App.Windows, not Ti.UI.Windows.

Second, I would think we'd want to give some mechanism to return the result of trying to request the extension, and to register a callback for if/when the extension is revoked. Additionally, right now this assumes the reason is "unspecified". We should probably expose the possible reasons.

Right now you simply request the extension and have no return value to indicate if it worked. You also have no way to handle when it gets revoked.

Can we take an optional object parameter that could contain the reason, the callback for the result, and the callback for revocation?

i.e.
```javascript
Ti.App.Windows.requestExtendedExecution({
  reason: Ti.App.Windows.EXTENDED_EXECUTION_REASON_SAVING_DATA,
  result: function (granted) { // granted is a boolean value, true or false
  },
  revoked: function(reason) { // reason is an enumerated constant from https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.extendedexecution.extendedexecutionrevokedreason
  }
});
```

Or maybe just avoid the object and do:
```javascript
Ti.App.Windows.requestExtendedExecution(Ti.App.Windows.EXTENDED_EXECUTION_REASON_SAVING_DATA, function (granted) { // granted is a boolean value, true or false
  }, function(reason) { // reason is an enumerated constant from https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.extendedexecution.extendedexecutionrevokedreason
  }
});
```
With the first arg optional, defaults to UNSPECIFIED, second arg is optional, callback for result, third is optional callback for revocation?

infosia:
@sgtcoolguy Updated, implemented like below.

```js

    Ti.App.Windows.requestExtendedExecution(
        Ti.App.Windows.EXTENDED_EXECUTION_REASON_UNSPECIFIED,
        function (granted) {
            alert('granted = ' + granted);
            Ti.App.Windows.closeExtendedExecution();
        },
        function (reason) {
            alert('revoked reason = ' + reason);
        }
    );
});

win.open();
```

Source/Ti/src/App/WindowsModule.cpp
```diff
@@ -43,12 +44,140 @@ namespace TitaniumWindows
...
+			return get_context().CreateNumber(1);
+#endif
+		}
+
+		TITANIUM_FUNCTION(WindowsModule, requestExtendedExecution)
+		{
+#if defined(IS_WINDOWS_10)
+			using namespace Windows::ApplicationModel::ExtendedExecution;
+
+			ENSURE_OPTIONAL_INT_AT_INDEX(reason, 0, 0);
```
sgtcoolguy: So I assume that 0 == Unspecified, but it leaves me a little concerned that we're not explicitly using the value of `Windows::ApplicationModel::ExtendedExecution::ExtendedExecutionReason::Unspecified` here.

infosia: Makes sense, updated 👍

infosia:
@ewanharris I think that makes sense. Making the parameter `Object` should make the developer's option more flexible and more unambiguous.

Fixed, now it's like below. Every params for `requestExtendedExecution` is optional now.

```js
var label = Ti.UI.createLabel();

var counter = 0;
var handle;
var swtch = Ti.UI.createSwitch({
    top: 100,
    title: 'Update in background?'
});

swtch.addEventListener('change', function (e) {
    if (e.value) {
        Ti.App.Windows.requestExtendedExecution({
            reason: Ti.App.Windows.EXTENDED_EXECUTION_REASON_UNSPECIFIED,
            result: function (granted) {
                Ti.API.info('granted = ' + granted);
                if (granted) {
                    handle = setInterval(function () {
                        counter++;
                        label.text = counter
                    }, 100);
                } else {
                    alert('Couldnt allow extended execution')
                }
            },
            revoked: function (reason) {
                Ti.API.info('revoked')
            }
        });
    } else {
        Ti.API.info('Stopping execution');
        Ti.App.Windows.closeExtendedExecution();
        clearInterval(handle);
    }
});

Ti.App.addEventListener('resume', function () {
    console.log('resumed');
})

win.add(label);
win.add(swtch);
win.open();
```

PR #1118, opened 2017-09-27
[TIMOB-25325] Click event of Button does not provide coordinates

infosia:
[TIMOB-25325](https://jira.appcelerator.org/browse/TIMOB-25325)

```js
    title: 'push',
    height: 100,
    width: 100
});

    // This should always returns "50 : 50" (center coordinate of the button) on Windows
    alert(e.x + ' : ' + e.y);
});

win.add(button);
win.open();
```

Note: On Windows, `x` and `y` always returns center position of the Button because Windows Button does not actually provide the pointer coordinate information along with the event.


infosia:
@garymathews  We can't make test case for this as we don't automate UI touch event in Jenkins?

garymathews:
@infosia Yep, you're right. No idea why I mentioned that. 😫

PR #1125, opened 2017-09-29
[6_3_X][TIMOB-25357] Add deprecation warning when building using 8.1

infosia:
[TIMOB-25357](https://jira.appcelerator.org/browse/TIMOB-25357)

```
appc run -p windows --win-sdk 8.1 --target ws-local -l trace
```

Expected: The build went through, but should see `Support for Windows 8.1 and Windows Phone 8.1 SDKs has been dropped as of Titanium SDK 6.3.0.GA, and will be removed in Titanium SDK 7.0.0.GA` warning message.

### DIFF
--- a/cli/commands/_build/validate.js
+++ b/cli/commands/_build/validate.js
@@ -42,6 +42,10 @@ function validate(logger, config, cli) {
 	this.deployType = !/^dist-$/.test(this.target) && cli.argv['deploy-type'] ? cli.argv['deploy-type'] : this.deployTypes[this.target];
 	this.buildType = cli.argv['build-type'] || '';
 
+	if (this.wpsdk == '8.1') {
+		logger.warn(__('Support for Windows 8.1 and Windows Phone 8.1 SDKs has been dropped as of Titanium SDK 6.3.0.GA, and will be removed in Titanium SDK 7.0.0.GA'));
+	}
+
 	// ti.deploytype is deprecated and so we force the real deploy type
 	if (cli.tiapp.properties['ti.deploytype']) {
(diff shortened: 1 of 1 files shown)


cli/commands/_build/validate.js
```diff
@@ -42,6 +42,10 @@ function validate(logger, config, cli) {
 	this.deployType = !/^dist-$/.test(this.target) && cli.argv['deploy-type'] ? cli.argv['deploy-type'] : this.deployTypes[this.target];
 	this.buildType = cli.argv['build-type'] || '';
 
+	if (this.wpsdk == '8.1') {
+		logger.warn(__('Support for Windows 8.1 and Windows Phone 8.1 SDKs has been dropped as of Titanium SDK 6.3.0.GA, and will be removed in Titanium SDK 7.0.0.GA'));
```
hansemannn: Can we include a note that this has been dropped because Microsoft's decision to drop support? Otherwise people might think we drop it because we are not able to support it.

infosia: What would you recommend?  `Support for Windows 8.1 and Windows Phone 8.1 SDKs has been dropped as Microsoft has ended support for Windows Phone 8.1` ? Too verbose? 

Note that things may be little complicated here, we are going to drop `8.1` build target but it actually includes both `Windows Store 8.1 app` and `Windows Phone 8.1` app target. Microsoft has only ended support for Windows Phone 8.1.

cc @ewanharris @sgtcoolguy

ewanharris: @infosia  Sorry I meant to update here/the ticket, we discussed the message after standup and the general consensus was to link it back to Microsoft. However I had overlooked that it was only 8.1 being dropped, I think considering that it might be better to leave this as it is and then we can explain better in the blog post?

infosia: @ewanharris Sounds good to me 👍

PR #1134, opened 2017-10-05
[TIMOB-25233] Implement TextField.setSelection

infosia:
[TIMOB-25233](https://jira.appcelerator.org/browse/TIMOB-25233)

```js
var _window = Ti.UI.createWindow({ backgroundColor: 'green' });
var textfield0 = Titanium.UI.createTextField({
    width: Ti.UI.FILL,
    top: 100,
    value: 'textfield0'
});

_window.addEventListener('open', function () {
    textfield0.setSelection(1, 4);
});

_window.add(textfield0);
_window.open();
```

Expected: texts between 1 to 3 character (`'ext'`) is selected.

### DIFF
--- a/Source/UI/src/TextField.cpp
+++ b/Source/UI/src/TextField.cpp
@@ -295,6 +295,19 @@ namespace TitaniumWindows
 			return false;
 		}
 
+		void TextField::setSelection(const std::uint32_t& start, const std::uint32_t& end) TITANIUM_NOEXCEPT
+		{
+			if (text_box__) {
+				text_box__->Select(start, (end - start));
+			} else if (password_box__) {
+				//
+				// Windows PasswordBox doesn't support selection with range. 
+				//
+				TITANIUM_LOG_WARN("TextField.setSelection doesn't work with passwordMask on Windows");
+			}
+		}
+
+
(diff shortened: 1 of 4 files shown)


Source/UI/src/TextField.cpp
```diff
@@ -295,6 +295,19 @@ namespace TitaniumWindows
 			return false;
 		}
 
+		void TextField::setSelection(const std::uint32_t& start, const std::uint32_t& end) TITANIUM_NOEXCEPT
+		{
+			if (text_box__) {
+				text_box__->Select(start, (end - start));
```
garymathews: Should probably validate `start < length` and `end < length`

infosia: Yes, updated 👍

ssjsamir:
@infosia 'ext' is not highlighted on a mobile device running windows 10  (10.0.14393.0) but when running on ws-local 'ext' is highlighted as expected.

infosia:
@ssjsamir Hmm I think it is platform component issue. Microsoft doesn't say it does not work on mobile but it does not actually  work :(  I will add note to our API doc.

PR #1137, opened 2017-10-10
[TIMOB-25250] Implement ActivityIndicator.indicatorDiameter

infosia:
[TIMOB-25250](https://jira.appcelerator.org/browse/TIMOB-25250)

Implement `ActivityIndicator.indicatorDiameter` property. Without specifying `indicatorDiameter`, the progress ring goes to have a default size which tends to be small. You can resize size of the ring by changing `indicatorDiameter` property.

```js
﻿﻿var win = Ti.UI.createWindow({
    backgroundColor: 'white',
});

var activityIndicator = Ti.UI.createActivityIndicator({
    message: 'Loading...',
    top: 10,
    left: 10,
    height: 120,
    width: 120,
    indicatorDiameter: 80,
});

win.add(activityIndicator);

win.addEventListener('open', function (e) {
    activityIndicator.show();
});

win.open();
```

### DIFF
--- a/Source/UI/src/ActivityIndicator.cpp
+++ b/Source/UI/src/ActivityIndicator.cpp
@@ -61,10 +61,9 @@ namespace TitaniumWindows
 
 			label__->FontSize = TitaniumWindows::UI::Label::DefaultFontSize;
 
-			// relative position from container should start from top & left
-			panel__->VerticalAlignment   = VerticalAlignment::Top;
-			panel__->HorizontalAlignment = HorizontalAlignment::Left;
-			label__->HorizontalAlignment = HorizontalAlignment::Center;
+			// alignment of the components should be centered
+			panel__->VerticalAlignment   = VerticalAlignment::Center;
+			panel__->HorizontalAlignment = HorizontalAlignment::Center;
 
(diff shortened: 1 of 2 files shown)


infosia:
@garymathews Oh wait...I was wrong. It is never used 💦 Deleted it from sample code 👍

Source/UI/src/ActivityIndicator.cpp
```diff
@@ -89,6 +88,14 @@ namespace TitaniumWindows
 			ring__->Foreground = ref new Windows::UI::Xaml::Media::SolidColorBrush(color_obj);
 		}
 
+		void ActivityIndicator::set_indicatorDiameter(const std::string& diameter) TITANIUM_NOEXCEPT
+		{
+			Titanium::UI::ActivityIndicator::set_indicatorDiameter(diameter);
+			const auto value = std::stod(diameter);
```
garymathews: Since this accepts `String`, shouldn't we allow for density units? (`dp``px` etc..)

infosia: Makes sense, updated.

PR #1142, opened 2017-10-12
[TIMOB-25385] TableViewRow with Label breaks layout

infosia:
[TIMOB-25385](https://jira.appcelerator.org/browse/TIMOB-25385)

The first row in `TableView` is not displayed when using with `Label`. I'm pushing this to `6_3_X` too because it's critical.

```js
    backgroundColor: 'green',
    layout: "vertical",
});

var row1 = Ti.UI.createTableViewRow();
row1.add(Ti.UI.createLabel({
    text: "THIS IS THE FIRST ROW",
}));

var row2 = Ti.UI.createTableViewRow();
row2.add(Ti.UI.createLabel({
    text: "THIS IS THE SECOND ROW",
}));

    data: [row1, row2],
});

    text: "This label breaks first row",
});

win.add(label);
win.add(table);

win.open();
```

Note: This is only a workaround because I'm still not sure why table view row returns incorrect row size only for the first row.

### DIFF
--- a/Source/UI/src/Label.cpp
+++ b/Source/UI/src/Label.cpp
@@ -79,7 +79,9 @@ namespace TitaniumWindows
 							const auto width  = panel->ActualWidth;
 							const auto height = panel->ActualHeight;
 							if (width > 0) {
-								if (!skipResizing) label__->MaxWidth = width;
+								if (!skipResizing) {
+									label__->MaxWidth = width;
+								}
 								if (!layout->get_right().empty()) {
 									const auto ppi = WindowsViewLayoutDelegate::ComputePPI(Titanium::LayoutEngine::ValueName::Width);
 									const auto rightPadding = Titanium::LayoutEngine::parseUnitValue(layout->get_right(), Titanium::LayoutEngine::ValueType::Fixed, ppi, "px");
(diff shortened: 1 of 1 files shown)


Source/UI/src/Label.cpp
```diff
@@ -66,11 +66,20 @@ namespace TitaniumWindows
...
+							const auto parent = get_parent();
+							if (parent && parent->get_apiName() == "Ti.UI.TableViewRow") {
+								skipResizing = true;
+							}
 							const auto layout = getViewLayoutDelegate<WindowsViewLayoutDelegate>();
 							const auto width  = panel->ActualWidth;
 							const auto height = panel->ActualHeight;
 							if (width > 0) {
-								label__->MaxWidth = width;
+								if (!skipResizing) label__->MaxWidth = width;
```
garymathews: I know its only a workaround, but I think we always encapsulate `if` statements?
```C++
if (!skipResizing) {
    label__->MaxWidth = width;
}
```

infosia: Yes. I was lazy.

ewanharris:
@garymathews I assume other than the if statement style the CR was good? Could you merge this and #1143 if so. Thanks!

infosia:
Should be ok to merge now.  Merging.

PR #1145, opened 2017-10-16
[TIMOB-24746] Upgrade JavaScriptCore, use DLL instead of Lib

infosia:
[TIMOB-24746](https://jira.appcelerator.org/browse/TIMOB-24746)

We need to catch up on the latest JavaScriptCore. We are also going to use **JavaScriptCore.DLL** instead of **JavaScriptCore.LIB**, that is much smaller regarding file size (**2GB -> 100MB**), and thus it should make our build faster. I expect this could make our CI build more stable because this should use less internal network resources.


sgtcoolguy:
Manually merged

sgtcoolguy:
Manually merged

sgtcoolguy:
Manually merged

PR #1154, opened 2017-10-26
[TIMOB-25433] Add "run-cmake" on missing cmake project

infosia:
[TIMOB-25433](https://jira.appcelerator.org/browse/TIMOB-25433)

Add `--run-cmake` automatically when cmake directory does not exist.

How to test:

1. Create module project
2. cd `windows/`
3. Remove `Windows10.ARM/` and `Windows10.Win32/` directory
4. `appc ti build -p windows -l trace`
5. cmake projects should be re-created and module is built


cli/commands/_buildModule.js
```diff
@@ -542,6 +541,14 @@ function runCmake(data, platform, arch, sdkVersion, next) {
         targetSdkVersion = data.targetPlatformSdkVersion;
     }
 
+    logger.debug('cmake ' +
```
garymathews: I think we should put the args into an array to use in `logger` and `spawn`. So they both use the same array.

infosia: Updated 👍

ssjsamir:
@infosia Ready to merge once Jenkins passes

mukherjee2:
@infosia can you please look at the unit test failures?

infosia:
Merged.

PR #1165, opened 2017-12-07
[TIMOB-25587] Stop deleting build directory on rebuild

infosia:
[TIMOB-25587](https://jira.appcelerator.org/browse/TIMOB-25587)

Windows build wipes out the directory and rebuilds each time. We should be able to re-use existing project directory and stop wiping it out on each build, which should improve build time.

I observed that this improves build time, for me it was `48 sec` to `35 sec`.

**Expected**: `C:\Users\USER_NAME\.titanium\vsbuild\PROJECT_NAME` should not be erased on each re-build.

### DIFF
--- a/cli/commands/_build/checks.js
+++ b/cli/commands/_build/checks.js
@@ -207,10 +207,6 @@ function checkIfNeedToRecompile(next) {
 		wrench.rmdirSyncRecursive(this.buildDir);
 	}
 
-	// now that we've read the build manifest, delete it so if this build
-	// becomes incomplete, the next build will be a full rebuild
-	fs.existsSync(this.buildManifestFile) && fs.unlinkSync(this.buildManifestFile);
-
 	next();
 };
 
--- a/cli/commands/_build/run.js
+++ b/cli/commands/_build/run.js
@@ -68,7 +68,6 @@ function run(logger, config, cli, finished) {
 		},
 
 		'writeBuildManifest',
-		'copyResultsToProject',
 
 		function (next) {
 			// now that the app is built, if we're going to do some logging, then we print how long the app took so far
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -21,29 +21,7 @@ exports.mixin = mixin;
  Implementation.
  */
 function mixin(WindowsBuilder) {
-	WindowsBuilder.prototype.copyResultsToProject = copyResultsToProject;
 	WindowsBuilder.prototype.copyResources = copyResources;
-}
-
-// Copy to original location!
-/**
- * Copies the build directory back to the project's build directory if we compiled in temp.
- *
- * @param {Function} next - A function to call after the build manifest has been written.
- */
-function copyResultsToProject(next) {
-	if (this.originalBuildDir) {
-		this.logger.info(__('Copying results back to project build directory'));
-		// if already exists, wipe it
-		fs.existsSync(this.originalBuildDir) && wrench.rmdirSyncRecursive(this.originalBuildDir);
-		// make sure destination exists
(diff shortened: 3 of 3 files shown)


cli/commands/_build/checks.js
```diff
@@ -207,10 +207,6 @@ function checkIfNeedToRecompile(next) {
 		wrench.rmdirSyncRecursive(this.buildDir);
 	}
 
-	// now that we've read the build manifest, delete it so if this build
-	// becomes incomplete, the next build will be a full rebuild
-	fs.existsSync(this.buildManifestFile) && fs.unlinkSync(this.buildManifestFile);
```
infosia: I don't remember why we did this. I know build manifest has some important configuration but I don't think we need to re-create it each time on build. @sgtcoolguy @garymathews Do you guys remember why?

cli/commands/_build/run.js
```diff
@@ -57,7 +57,6 @@ function run(logger, config, cli, finished) {
 		'copyModuleOverride',
 		'compileApp',
 		'writeBuildManifest',
-		'copyResultsToProject',
```
infosia: I don't remember why we need this either. I would guess we thought we expect our project file placed under `TITANIUM_PROJECT_DIR/build/windows/` to align what we have seen on other platforms? I don't think we need this because Windows app build actually worked without this. @sgtcoolguy @garymathews Do you guys remember why we did this?

garymathews: Isn't this because we ran into the MAX_PATH issue? So we build in a temp directory and copy it back as a means to avoid that issue.

infosia: @garymathews Yeah, so we are doing everything under temp directory for now, and I'm wondering if we really need to copy it back to the project build folder. Windows app build actually worked without copying it back.

Buder:
Don't know if I can comment on.this, but yes at the time I was building the
windows part of a app project Inhad problems with build while.it was not
moved "back" , I believe it was so that win could build at all. Sorry if I
only.confirm this problem but it was so at the time, over a year ago.

maybe now it is different.

I used Windows 8.1 phone and host and I was early.on. As you may remember.

Let me know if I.can help. Upgrading my win system as we speak but holds
off the win OS and VS until Monday.

By the way, such great work you are doing with appcelerator and windows.
MUCH appreciated.

Thanks

-------------------------------------------------
Jörgen Buder, CTO & Partner
Mobile Solutions Architect
Purple Scout AB
Stortorget 29
211 34 Malmö

An Appcelerator Premium Partner, Training partner and Certified Instructor.

mobile: + 46 708 737879
www.purplescout.se

Den 7 dec. 2017 11:40 em skrev "Kota Iguchi" <notifications@github.com>:

cli/commands/_build/checks.js
ewanharris: I think this is standard across all platforms build processes [android](https://github.com/appcelerator/titanium_mobile/blob/da62debb1eb8e74f664aa8e89301457823802fc5/android/cli/commands/_build.js#L2231-L2233), [iOS](https://github.com/appcelerator/titanium_mobile/blob/da62debb1eb8e74f664aa8e89301457823802fc5/iphone/cli/commands/_build.js#L2661-L2663). My understanding is that if the build process fails, we will always want to start from a clean state next time, we write a new buildmanifest.json as one of the last build steps so on successful builds it will always exist.

cli/commands/_build/run.js
ewanharris: It was because of MAX_PATH issues, but I think we need to copy the resources back because the initial build checks to check if a full rebuild is needed use the <project-dir>/build directory. Rather than copying using the blocking `wrench.copyDirSyncRecursive` could we maybe copy non-blocking and allow the build to continue on while we copy?

infosia:
Closing, turns out removing manifest is needed it helps detecting incomplete build.

PR #1168, opened 2017-12-19
[TIMOB-25616] Don't uninstall exisitng app unless user explicitly specifies it

infosia:
This requires https://github.com/appcelerator/windowslib/pull/83 to be merged.

[TIMOB-25616](https://jira.appcelerator.org/browse/TIMOB-25616)

Titanium have been uninstalling existing app in order to complete the installation when same version of the app is already installed, but we should not uninstall existing app unless user explicitly specifies `--forceUnInstall`option. It should ends up showing error message when there's same version of the app installed, which should be the default behavior.

Expected:

1. Installation should succeed when there's no same app installed.
2. Installation should fail when there's same app installed. 
3. Installation should succeed with `--forceUnInstall` option when there's same app installed. 

```
appc ti build -p windows --wp-sdk 10.0 --target wp-device --skipInstallDependencies -l trace --forceUnInstall
```

### DIFF
--- a/Tools/Scripts/build/test.js
+++ b/Tools/Scripts/build/test.js
@@ -307,7 +307,7 @@ function runBuild(target, deviceId, sdkVersion, count, next) {
 			'--target', target,
 			'--wp-sdk', sdkVersion,
 			'--win-publisher-id', '13AFB724-65F2-4F30-8994-C79399EDBD80',
-			'--no-prompt', '--no-colors'
+			'--no-prompt', '--no-colors', '--forceUnInstall'
 		];
 	if (deviceId) {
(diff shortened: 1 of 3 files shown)


sgtcoolguy:
appcelerator/windowslib#83 and appcelerator/titanium_mobile#9760 have been merged (to master). This should be ready for QE to test (not sure if @ewanharris is the right now to tag for it).

infosia:
@ewanharris Updated, suppress uninstallation unless `--forceUnInstall` is specified for `ws-local` too 👍

infosia:
Merged these changes into #1177. Closing.

PR #1170, opened 2018-01-05
[TIMOB-25641] View.add should be ignored when it's ready added

infosia:
[TIMOB-25641](https://jira.appcelerator.org/browse/TIMOB-25641)

TIMOB-25465 and TIMOB-25639 show that rendering issue happens when you try to add a view that is already added. According to the current code, we are not actually checking if the View is already added but just leaves the platform behavior how to render it in that case. We should just ignore the add operation when it is already added.

For test cases, just make sure nothing wrong happens when you add same view more than twice.

```js

var view = Ti.UI.createView({
    backgroundColor: 'red',
    width: 100, height: 100
});

win.add(view);
win.add(view);
win.open();
```

### DIFF
--- a/Source/TitaniumKit/src/UI/View.cpp
+++ b/Source/TitaniumKit/src/UI/View.cpp
@@ -113,13 +113,20 @@ namespace Titanium
 				return;
 			}
 			auto view_ptr = view.GetPrivate<View>();
+			const auto parent_view_ptr = this->get_object().GetPrivate<View>();
+
+			// Compare parent, this indicates you're trying to add it before it's removed
+			if (view_ptr->get_parent() == parent_view_ptr) {
+				get_context().JSEvaluateScript("Ti.API.warn('View.add: This indicates this component is already added');");
+				return;
+			}
 
 			// For mixing Ti.Ui.View and native view.
 			if (view_ptr == nullptr) {
 				view_ptr = layoutDelegate__->rescueGetView(view);
 			}
 
-			view_ptr->set_parent(this->get_object().GetPrivate<View>());
+			view_ptr->set_parent(parent_view_ptr);
 
 			layoutDelegate__->add(view_ptr);
 		}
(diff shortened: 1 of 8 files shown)


Source/TitaniumKit/src/UI/View.cpp
```diff
@@ -113,13 +113,20 @@ namespace Titanium
 				return;
 			}
 			auto view_ptr = view.GetPrivate<View>();
+			const auto parent_view_ptr = this->get_object().GetPrivate<View>();
+
+			// Compare parent, this indicates you're trying to add it before it's removed
+			if (view_ptr->get_parent() == parent_view_ptr) {
+				TITANIUM_LOG_WARN("Ti.API.warn('View.add: This indicates this component is already added');");
```
garymathews: Shouldn't this just be:
```
View.add: This indicates this component is already added.
```

Source/TitaniumKit/src/UI/View.cpp
```diff
@@ -680,7 +689,20 @@ namespace Titanium
...
 
-			layoutDelegate__->insertAt(js_to_ViewInsertOrReplaceParams(params));
+			const auto viewInsertParams = js_to_ViewInsertOrReplaceParams(params);
+			const auto view_ptr = viewInsertParams.view;
+
+			const auto parent_view_ptr = this->get_object().GetPrivate<View>();
+
+			// Compare parent, this indicates you're trying to add it before it's removed
+			if (view_ptr->get_parent() == parent_view_ptr) {
+				TITANIUM_LOG_WARN("Ti.API.warn('View.add: This indicates this component is already added');");
```
garymathews: Same here

Source/TitaniumKit/src/UI/View.cpp
infosia: oh... right. why did I do this 😓

PR #1175, opened 2018-01-28
[TIMOB-24610] Implement babel transpile

garymathews:
##### NOTE: requires https://github.com/appcelerator/node-titanium-sdk/pull/20

- Implement ability to transpile javascript via babel by specifying:
###### tiapp.xml
```XML
...
<transpile>true</transpile>
...
``` 

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-24610)


cli/commands/_build/copy.js
```diff
@@ -603,25 +603,33 @@ function copyResources(next) {
...
 							// parse the AST
-							var r = jsanalyze.analyzeJsFile(from, {minify: this.minifyJS});
+							const transpile = this.cli.tiapp.transpile || false;
+							const originalContents = fs.readFileSync(from).toString();
+							const r = jsanalyze.analyzeJs(originalContents, {
+								filename: from,
+								minify: this.minifyJS,
+								transpile: transpile,
+								targets: {
+									safari: '10' // matches the version of jscore we use
```
sgtcoolguy: So I'd prefer we keep this out of the code. I stuck hard-coded values in myself to get started, but then tried to move the JS engine info to the package.json of the platform (and ideally hook that into our actual build process is possible, like with android - so that it stays up to date).

cli/commands/_build/copy.js
```diff
@@ -603,25 +603,33 @@ function copyResources(next) {
 					this.cli.createHook('build.windows.analyzeJsFile', this, function (from, to, ast, traverse, types, cb) {
 						this.cli.createHook('build.windows.copyResource', this, function (from, to, cb) {
 							// parse the AST
-							var r = jsanalyze.analyzeJsFile(from, {minify: this.minifyJS});
+							const transpile = this.cli.tiapp.transpile || false;
```
sgtcoolguy: I'm not super fond of using a new tiapp.xml tag to determine this, but I suppose as an override it makes sense. I think the default value should be to transpile, though. We want to have ES6+ support out of the box with no changes. I think technically because Windows fork of JSC is newer they could use ES6 without transpiling, but I know for example that at least one of our ES6 tests crashed Windows before.

In fact, a good test for this PR is to see if the es6 rest args test now works (i.e. override the app.js file and don't skip requiring it now):
https://github.com/appcelerator/titanium-mobile-mocha-suite/blob/master/Resources/app.js#L41

We may find that the fork isn't quite at the Safari 10 level yet...

sgtcoolguy:
@garymathews The build is failing with:
```
[ERROR] :  symlinkFiles is not defined
```

I don't think symlinkFiles is ever set up in the build command, and I assume it's never true on Windows anyways?

cli/commands/_build/copy.js
```diff
@@ -603,25 +603,32 @@ function copyResources(next) {
...
 
-								this.cli.createHook('build.windows.compileJsFile', this, function (r, from, to, cb2) {
-									fs.writeFile(to, r.contents, cb2);
-								})(r, from, to, cb);
-							} else {
-								// we've already read in the file, so just write the original contents
-								this.logger.debug(__('Copying %s => %s', from.cyan, to.cyan));
-								fs.writeFile(to, r.contents, cb);
-							}
+								if (symlinkFiles && newContents === originalContents) {
```
sgtcoolguy: symlinkFiles is not defined. I assume it'd always be false on Windows anyhow?

I think here the best we can do is compare the new contents versus the destination file. If the destination file exists and has the same contents, then skip writing. Otherwise write the file.

garymathews:
@sgtcoolguy Updated PR and tested against https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/36

Works great! 👍

PR #1177, opened 2018-02-02
[TIMOB-25743] Auto-increment build number

infosia:
[TIMOB-25743](https://jira.appcelerator.org/browse/TIMOB-25743)

Simplest implementation for automatic version increment.

- If you omit the build number, we just generate it behind scenes, available only on `development` and `test` build.

For instance when you have `1.0` version in `tiapp.xml`, we generate the 4th segment `1.0.0.x` version number and increment this build number on each build. (numbering spec: **Major.Minor.Build.Revision**>
 
```xml
<version>1.0<version>
```

If you set build number explicitly in `tiapp.xml`, we don't generate the build number. In this case you need to bump up the build number manually in order to prevent deployment issue described in [TIMOB-25017](https://jira.appcelerator.org/browse/TIMOB-25017).

```xml
<version>1.0.1.1<version>
```

This feature can be disabled by setting false on `use-auto-versioning` in `windows` section.

```xml
  <windows>
    <use-auto-versioning>false</use-auto-versioning>
  </windows>
```

## Note

I'm pushing this simplest implementation hoping to get some more feedback from developers. Here's some starting point to discuss about the auto-increment spec:

- Do we accept the difference between version numbering in tiapp.xml and actual app? Some developers may want to fix specific version number for the release build. Do we want to update the tiapp.xml based on generated version number?
- We could introduce special placeholder like `<version>1.0.*</version>` but this may cause breaking change and parity issue between platforms. Do we wan to do like this? Or we could just say "If you omit the build number, we just automatically increment it behind scenes, only on Windows platform" like this PR.
- Do we accept this breaking change regarding new version numbering spec? 
- Do we want to enable it by default? Otherwise do we want to introduce new property in tiapp.xml like `<enable-auto-build-numbering/>` or so?
- Do we want to introduce this new version numbering spec for other platforms? Do we want to keep parity for it?
- Do we want to make it configurable from Appc Studio? For instance ["Automatically increment" checkbox in UWP packager](https://docs.microsoft.com/en-us/windows/uwp/packaging/images/packaging-screen5.jpg)

## Update (02/06/18)

Having discussed about the spec: we're going to implement as follows:

- We don't expose this auto-increament as a new feature
- We don't enable it on distribution build. We enable it only when you deploy it for testing installation (like `appc run`) to workaround the particular issue (TIMOB-25017)
- We enable it by default. It can be disabled by setting tiapp property
- We are going to target `7.1.0`


cli/commands/_build/generate.js
```diff
@@ -32,6 +33,31 @@ function mixin(WindowsBuilder) {
 	WindowsBuilder.prototype.fixCSharpConfiguration = fixCSharpConfiguration;
 	WindowsBuilder.prototype.copyModuleOverride = copyModuleOverride;
 }
+/**
+ * Generates new build number based on tiapp.version
+ *
+ * @param {Function} next - A function to call after new version number have been generated.
+ */
+function generateBuildVersion(next) {
+    var version = appc.version.format(this.tiapp.version, 2, 4, true);
```
ewanharris: I think we should only be doing this if we know we'll be installing the app. So we shouldn't do it on dist-phonestore, dist-winstore. I'm on the fence as to whether it should be done when using the `-build-only` flag, I guess seeing as we don't support remote deployment we should do it for that too

cli/commands/_build/generate.js
```diff
@@ -32,6 +33,31 @@ function mixin(WindowsBuilder) {
...
+ * Generates new build number based on tiapp.version
+ *
+ * @param {Function} next - A function to call after new version number have been generated.
+ */
+function generateBuildVersion(next) {
+    var version = appc.version.format(this.tiapp.version, 2, 4, true);
+    // If build number is omitted, we generate it
+    if (version.split('.').length < 3) {
+        var buildNumber = 0;
+        var buildNumberCache = path.join(this.projectDir, 'tiapp_build_number.txt');
```
ewanharris: Could we store the number we increment as a property on this.buildManifest? This should probably be as transparent as possible to a user so I'm wary of maintaining a file in their project, we would lose the buildmanifest file on a `ti clean` but that seems fine to me

infosia:
@ewanharris Thanks, Updated. I agree that we want to leverage buildManifest in this case. I'd think that we should bump up the build number on each build because of what "build number" indicates though. I'm also dithering whether if we should be doing this on "dist-*" build too...I would want this feature on dist build too because I found this useful to have this new auto-increament feature.

infosia:
Pushed fixes based on discussions today.

ewanharris:
@infosia I'm wondering whether it's worth picking the changes in #1168 into this to make testing easier, at the moment this PR will still do the force uninstall

cli/commands/_build/generate.js
```diff
@@ -33,6 +34,37 @@ function mixin(WindowsBuilder) {
...
+ * @param {Function} next - A function to call after new version number have been generated.
+ */
+function generateBuildVersion(next) {
+	// Generates new build number based on tiapp.version
+	var version = appc.version.format(this.tiapp.version, 2, 4, true);
+	// If build number is omitted, we generate it. Only available on development/test build.
+	// Can be disabled when 'use-auto-versioning' set false in <windows> section.
+	var disabled = this.tiapp.windows && this.tiapp.windows['use-auto-versioning'] === false;
+
+	if (!disabled && !/^dist-/.test(this.target) && version.split('.').length < 3) {
```
ewanharris: Should this be `&& version.split('.').length <= 3`? as we're changing the 4th segment I think we should only not do this if a user is specifying their version as `1.0.1.0`, currently if my version is `1.0.1` the auto-increment wont happen

infosia: 👍 Right, updated. Also merged #1168 so we can test `forceUninstall` too.

infosia:
@ewanharris  Nice! Fixed 👍

infosia:
Pushed backport for 7_1_X: #1199

PR #1178, opened 2018-02-06
[TIMOB-23247] CLI: Implement Native Module clean

sgtcoolguy:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-23247

**Description:**
```
 ti clean -p <platform>
```

This hooks up ti clean as a command that can be used for modules as we do with normal apps.
It's implemented in a single-platform at a time way, and isn't meant to yet support multi-platform modules like say hyperloop.

I took a best guess at the files we should be wiping out, but certainly open to suggestions/tweaks here.

Should be merged in conjunction with appcelerator/titanium_mobile#9807

### DIFF
--- a/cli/commands/_cleanModule.js
+++ b/cli/commands/_cleanModule.js
@@ -1,0 +1,44 @@
+/**
+* Android module clean command.
+*
+* @module cli/_cleanModule
+*
+* @copyright
+* Copyright (c) 2018 by Appcelerator, Inc. All Rights Reserved.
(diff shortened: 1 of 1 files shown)


cli/commands/_cleanModule.js
```diff
@@ -0,0 +1,44 @@
+/**
+* Android module clean command.
```
infosia: Windows

PR #1179, opened 2018-02-06
DO NOT MERGE Add linting for JS

ewanharris:
DO NOT MERGE

I was cleaning up some repos/branches and found this that I did a few months back and forgot to PR so figured why not do it now 🤷‍♂️ 

I'm marking as do not merge because:

- Is this wanted?
- No point landing until PRs touching .js files are merged (at which point I'll merge this branch with master)
- I assumed we wanted to keep the same line-endings rule (unix) as all other repos, but now I'm wondering if that's the right thing to do, it's super simple to switch back to Windows line-endings if that's what's wanted, it's super easy to switch back.
- (oh and looks like I broke the scripts too 😬 )

I also included the clang format bits for C++ files but haven't enforced it, that should probably be done in a separate PR

@sgtcoolguy @infosia @garymathews think it's worth it?


infosia:
Nice work! 👍 

I would think I'd like to avoid touching lots of .js files at once while we have pending PRs...
So how about this: 

Can we do like this, can we force to do `eslint` with the PR, just against specific JS files that is related to the PR? Hopefully I want see these changes by `eslint` as independent commit in the PR, that way we don't have to worry about touching JS files all at once, and we could revert the changes easily by just reverting the commit.

ewanharris:
@infosia I'm happy to wait on the merge for this until after all those PRs are in, I've had to rebase/merge master in a few times now.

I think we'd have to come up with a script to do that, I think the incremental updating was initially something that was desired for the Android formatting as I believe as they wished to update the formatting as they touched a file.

One concern I have with that is that a fair few of the js files in this repo are rarely touched, I believe we'd be best off bringing all files in line at once so that it's done with, any bugs that arise from most of the linting changes here are valid in my opinion things like using variables out of scope, strict equality checks not passing e.g 8.1 !== '8.1' and I think it's better to clear all of these out at once.

With Jenkins and a run through the functionalities as I'm making the changes I think we'd be able to tease out any lurking issues too

infosia:
@ewanharris I saw cwilliams cleaned up scripts in #1319 recently. Would it be plausible enough to close this PR?

infosia:
@ewanharris ok, sounds good to me 👍

PR #1182, opened 2018-02-12
[TIMOB-25760] Handle HttpClient redirect to get correct location

infosia:
[TIMOB-25760](https://jira.appcelerator.org/browse/TIMOB-25760)

```js
var _tmpURL = "https://goo.gl/TckK14"
var _tmpClient = Ti.Network.createHTTPClient({
    onload: function (e) {
        serverAddressRedirected = _tmpClient.location;
        //this should return redirected url
        alert("Location:" + serverAddressRedirected);
    },
    onerror: function (e) {
        alert("Error:" + e.error);
    },
    timeout: 30000
});
_tmpClient.open("GET", _tmpURL);
_tmpClient.send();
```

### DIFF
--- a/Source/Network/src/HTTPClient.cpp
+++ b/Source/Network/src/HTTPClient.cpp
@@ -106,7 +106,7 @@ namespace TitaniumWindows
 			filter__ = ref new Windows::Web::Http::Filters::HttpBaseProtocolFilter();
 			httpClient__ = ref new Windows::Web::Http::HttpClient(filter__);
 			cancellationTokenSource__ = concurrency::cancellation_token_source();
-			filter__->AllowAutoRedirect = true;
+			filter__->AllowAutoRedirect = false;
 			filter__->CacheControl->ReadBehavior = Windows::Web::Http::Filters::HttpCacheReadBehavior::MostRecent;
 
 			if (!username__.empty()) {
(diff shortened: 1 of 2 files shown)


Source/Network/src/HTTPClient.cpp
```diff
@@ -216,24 +216,50 @@ namespace TitaniumWindows
...
+			if (response) {
 				interruption_point();
-				readyState__ = Titanium::Network::RequestState::Opened;
-				onreadystatechange(readyState__);
 
-				SerializeHeaders(response);
+				// Handle redirects because we need to update location
+				const auto status = static_cast<std::uint32_t>(response->StatusCode);
+				const auto location = response->Headers->Location;
+				if (location && status >= 300 && status <= 399) {
```
infosia: Memo: Tried to handle all possible redirect request - https://long2know.com/2016/07/capturing-redirects-with-httpclient/ . As far as I saw `response->Headers->Location` goes `null` when you don't have redirects.

infosia: Pushed https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/41

infosia:
Pushed https://github.com/appcelerator/titanium-mobile-mocha-suite/pull/41

tests/Resources/ti.network.httpclient.test.js
```diff
@@ -0,0 +1,539 @@
...
+	});
+
+	// HttpClient.location should point to redirected location
+	it('TIMOB-25760', function (finish) {
+		var xhr = Ti.Network.createHTTPClient(),
+			attempts = 3;
+		xhr.setTimeout(6e4);
+
+		xhr.onload = function () {
+			should(xhr.location).be.eql('https://www.appcelerator.com/');
```
ewanharris: I think this should be `https://www.appcelerator.com/`. I also forgot to note that tests needs to be like `ti.network.httpclient.addontest.js` so this suite didn't run, I'll merge this and then file a new PR to fix the syntax and assertion

ewanharris: Doh, my bad I forgot I have an extension that removes the `https://www.` bit of links 😳 so ignore the comment about `https://www.appcelerator.com/.`

PR #1184, opened 2018-02-13
Add test overrides

ewanharris:


ewanharris:
Job https://jenkins.appcelerator.org/blue/organizations/jenkins/titanium-sdk%2Ftitanium_mobile_windows/detail/PR-1184/2/pipeline shows that this works, reverting ba0e1ca3749502ddfe65aa827acdf4cd368849bd

ewanharris:
@infosia Looks like I forgot to add this step when adding migrating to the mocha-suite, this'll allow you to check-in tests in a PR that will be ran in addition to the existing tests over at mocha-suite. Then every now and then we can migrate them from here to there in batches

infosia:
Nice! 👍 I'll try this with #1182.

PR #1187, opened 2018-02-17
[TIMOB-25787] Improve validation

garymathews:
- Improve some methods/properties validation

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-25787)


garymathews:
@infosia The tests technically would not be valid. For example I should not be able to do:
```JS
Ti.Media.AudiPlayer.time; // this still calls AudioPlayer::get_time(), it should not
```

infosia:
So yeah this is a basic design issue in HAL...It basically doesn't distinguish between static property and instance property and there's no way to define "instance only" property 😞  - [[TIMOB-18681](https://jira.appcelerator.org/browse/TIMOB-18681)]

At least I think we want to make sure that accessing to non-existent static property doesn't crash. I'll take over that.

PR #1173, opened 2018-01-23
[TIMOB-25696] Runtime error in TCP callback causes crash

infosia:
[TIMOB-25696](https://jira.appcelerator.org/browse/TIMOB-25696)

Application crashes when there's runtime error during TCP callback. For example, following app crashes without any error logs.

### app.js
```javascript
var win = Ti.UI.createWindow({ backgroundColor: 'white' });

win.addEventListener('open', function () {
    console.log('Connecting to server');
    var socket = Ti.Network.Socket.createTCP({
        host: '127.0.0.1', port: 6262,
        connected: function (e) {
            Ti.API.info('Connected to server at port ' + port);
            win.backgroundColor = 'green';
        },
        error: function (e) {
            alert(e.error);
            Ti.API.info('Error connecting to server (' + e.errorCode + '): ' + e.error);
            win.backgroundColor = 'red';
        },
    });
    socket.connect();
});

win.open();
```
**Expected**: This Titanium app should not crash. This should alert `Can't find variable: port` because there's runtime error in `connected` callback (the variable `port` is not defined).

Here's a sample server code to debug the issue. Launch this using `node server.js` in Node.js in order to test this.

### server.js
```javascript
var net = require('net'),
	port = 6262;

net.createServer(function (socket) {
	console.log('client ' + socket.remoteAddress + ' connected');

	socket.on('end', function () {
		console.log('client ' + socket.remoteAddress + ' disconnected');
	});

}).listen(port, function() {
	console.log('server listening on port ' + port);
});
```

PR #1192, opened 2018-02-20
Revert #1182

infosia:
Because it broke liveview.


infosia:
Hmm no, #1173 broke it. Closing this.

PR #1209, opened 2018-03-07
[TIMOB-13379] Implement console.time/timeEnd

ewanharris:
**JIRA:** https://jira.appcelerator.org/browse/TIMOB-13379

Implement console.time/timeEnd according to https://console.spec.whatwg.org/#time, with the optional log on `console.time('duplicate label')` implemented, and a slight deviation to log a warning when calling `console.timeEnd('does not exist')`.

#### Test code

```js
console.time('set label'); // Set custom labels
console.time('set label'); // Should log "Label set label already exists" at a warn lebel
setTimeout(() => {
    console.timeEnd('set label'); // Should log "set label: 100ms" (not exactly 100ms)
}, 100);

console.time(); // Defaults to "default" if no value passed in
setTimeout(() => {
    console.timeEnd(); // Should log "set label: 5000ms" (not exactly 5000ms)
}, 5000);
console.timeEnd('No exist') // Should log "Label No exist does not exist" at warn level
console.time('foo');
console.timeEnd('foo');
console.time('foo'); // should not log a warning as we allow reuse if timeEnd(label) has been called
```

#### Output

```txt
[WARN] Label "set label" already exists
[WARN] Label "No exist" does not exist
[INFO] set label: 141ms
[INFO] default: 5047ms
```

### DIFF
--- a/Source/TitaniumKit/src/TiModule.cpp
+++ b/Source/TitaniumKit/src/TiModule.cpp
@@ -488,38 +488,63 @@ namespace Titanium
 
 		JSString builtin_functions_script = R"js(
 			try {
-			  console = {};
-			  console.log   = Ti.API.info;
-			  console.info  = Ti.API.info;
-			  console.warn  = Ti.API.warn;
-			  console.error = Ti.API.error;
-			  console.debug = Ti.API.debug;
+				console = {};
+				console._times = {};
+				console.log   = Ti.API.info;
+				console.info  = Ti.API.info;
+				console.warn  = Ti.API.warn;
+				console.error = Ti.API.error;
+				console.debug = Ti.API.debug;
+				console.time = function (label) {
+					if (console._times[label]) {
+						console.warn('Label "' + label + '" already exists');
+						return;
+					}
+					if (!label) {
+						label = 'default';
+					}
+					console._times[label] = Date.now();
+				};
+				console.timeEnd = function (label) {
+					if (!label) {
+						label = 'default';
+					}
+					var startTime = console._times[label];
+					if (!startTime) {
+						console.warn('Label "' + label + '" does not exist');
+						return;
+					}
+					var duration = Date.now() - startTime;
+					console.log(label + ': ' + duration + 'ms');
+					delete console._times[label];
+				};
+			
 
-			  // Create the global alert function
-			  alert = function (_msg) {
-				  Ti.UI.createAlertDialog({
-					  title: '',
-					  message: _msg + ''
-				  }).show();
-			  };
+				// Create the global alert function
+				alert = function (_msg) {
+					Ti.UI.createAlertDialog({
+						title: '',
+						message: _msg + ''
+					}).show();
+				};
 
-			  // Load _app_info_.json
-			  Ti.App._loadAppInfo();
-			  // Load _app_props_.json
-			  Ti.App.Properties._loadAppProperties();
+				// Load _app_info_.json
+				Ti.App._loadAppInfo();
+				// Load _app_props_.json
+				Ti.App.Properties._loadAppProperties();
 
-			  // Let's set up our user agent in JS, way easier this way...
-			  Ti.userAgent = 'Appcelerator Titanium/' + Ti.version + ' (' + Ti.Platform.model + '/' + Ti.Platform.version + '; ' + Ti.Platform.osname + '; ' + Ti.Platform.locale + ';)';
+				// Let's set up our user agent in JS, way easier this way...
+				Ti.userAgent = 'Appcelerator Titanium/' + Ti.version + ' (' + Ti.Platform.model + '/' + Ti.Platform.version + '; ' + Ti.Platform.osname + '; ' + Ti.Platform.locale + ';)';
 
-			  Ti.Network.encodeURIComponent = encodeURIComponent;
-			  Ti.Network.decodeURIComponent = decodeURIComponent;
+				Ti.Network.encodeURIComponent = encodeURIComponent;
+				Ti.Network.decodeURIComponent = decodeURIComponent;
 
-			  // Start analytics
-			  Ti.Analytics._start();
+				// Start analytics
+				Ti.Analytics._start();
 
-			  L = Ti.Locale.getString;
+				L = Ti.Locale.getString;
 			} catch (E) {
-			  Ti.API.error("Failed to initialize Titanium: " + E.toString());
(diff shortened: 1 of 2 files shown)


Source/TitaniumKit/src/TiModule.cpp
```diff
@@ -488,38 +488,63 @@ namespace Titanium
...
+				console.warn  = Ti.API.warn;
+				console.error = Ti.API.error;
+				console.debug = Ti.API.debug;
+				console.time = function (label) {
+					if (console._times[label]) {
+						console.warn('Label "' + label + '" already exists');
+						return;
+					}
+					if (!label) {
+						label = 'default';
```
infosia: default label is `undefined`,

```js
console.time();
setTimeout(() => {
    console.timeEnd();
}, 5000);
```

actually returns "`undefined: 5000ms`" (not exactly 5000ms) on Node.js.

Source/TitaniumKit/src/TiModule.cpp
```diff
@@ -488,38 +488,63 @@ namespace Titanium
...
+					}
+					var startTime = console._times[label];
+					if (!startTime) {
+						console.warn('Label "' + label + '" does not exist');
+						return;
+					}
+					var duration = Date.now() - startTime;
+					console.log(label + ': ' + duration + 'ms');
+					delete console._times[label];
```
infosia: Deleting label resets time here, right? This should have been available even after `timeEnd` is called multiple times like below (behavior in Node.js).

```js
> console.time('set label');
undefined
> console.timeEnd('set label');
set label: 9435ms
undefined
> console.timeEnd('set label');
set label: 11010ms
undefined
> console.timeEnd('set label');
set label: 13125ms
undefined
```

Source/TitaniumKit/src/TiModule.cpp
```diff
@@ -488,38 +488,63 @@ namespace Titanium
...
+					}
+					console._times[label] = Date.now();
+				};
+				console.timeEnd = function (label) {
+					if (!label) {
+						label = 'default';
+					}
+					var startTime = console._times[label];
+					if (!startTime) {
+						console.warn('Label "' + label + '" does not exist');
```
infosia: Calling invalid label on `timeEnd` should throw JS Error according to Node.js.

```js
> console.timeEnd('No exist');
Error: No such label: No exist
    at Console.timeEnd (console.js:69:11)
    at repl:1:9
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:441:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:212:10)
    at REPLServer.Interface._line (readline.js:551:8)
> try {console.timeEnd('No exist');} catch (E) {}
undefined
> try {console.timeEnd('No exist');} catch (E) {E}
[Error: No such label: No exist]
```

Source/TitaniumKit/src/TiModule.cpp
```diff
@@ -488,38 +488,63 @@ namespace Titanium
...
+				console = {};
+				console._times = {};
+				console.log   = Ti.API.info;
+				console.info  = Ti.API.info;
+				console.warn  = Ti.API.warn;
+				console.error = Ti.API.error;
+				console.debug = Ti.API.debug;
+				console.time = function (label) {
+					if (console._times[label]) {
+						console.warn('Label "' + label + '" already exists');
```
infosia: This should just reset the time according to behavior on Node.js.

```js
> console.time('set label');
undefined
> console.timeEnd('set label');
set label: 6882ms
undefined
> console.time('set label');
undefined
> console.timeEnd('set label');
set label: 2370ms
undefined
> console.time('set label');
undefined
> console.timeEnd('set label');
set label: 1380ms
undefined
```

Source/TitaniumKit/src/TiModule.cpp
ewanharris: @infosia Looks to be a difference between Node versions, 9.0.0 and above default to `default`, previous versions default to `undefined`. I prefer the newer way of defaulting to `default`but happy to update my PRs if consensus is on the latter

Source/TitaniumKit/src/TiModule.cpp
ewanharris: Again looks to be a difference between Node versions. They look to have begun deleting labels when timeEnd is called [since Node 6](https://github.com/nodejs/node/pull/3562) to align with Chromes behaviour.

Source/TitaniumKit/src/TiModule.cpp
ewanharris: There's no definition in the spec around what the behaviour should be here. Node throws, Chrome logs `No exist: 0ms`, I didn't want to throw as it seemed like an unnecessary use of an error (plus throwing the RSOD seemed unnecessary too) so I settled for logging at warning. Happy to change though

Source/TitaniumKit/src/TiModule.cpp
ewanharris: This is an optional from the [spec](https://console.spec.whatwg.org/#time), where if a label exists it does not overwrite, and can optionally log a warning

```
If the associated timer table contains an entry with key label, return, optionally reporting a warning to the console indicating that a timer with label label has already been started.
```

garymathews: @ewanharris I think it would be better to reset the time instead of logging a message. It seems more functional than having to call `timeEnd` to remove the label and re-add it.

ewanharris: Yeah I agree on it being more functional, I don't think we should differ from other environments though. There looks to be some discussion around logging values between calling `console.time`/`console.timeEnd` [in the spec repo](https://github.com/whatwg/console/issues/84) so I'll track that and if a conclusion is reached I'll implement on Ti

sgtcoolguy: Looks like they've been converging towards adding a timeLog() method that prints out the elapsed time for a label (with an optional extra decription/label) without removing the timer from the map - to handle the ambiguity/usefulness around trying to print out multiple points from the same beginning.

Heres the timeLog addition: https://github.com/whatwg/console/pull/138

So I think what's specified in @ewanharris 's PRs looks Ok for now, but we should consider adding `console.timeLog` now as well.

PR #1211, opened 2018-03-19
[TIMOB-25881] Implement WebView.blacklistedURLs

infosia:
[TIMOB-25881](https://jira.appcelerator.org/browse/TIMOB-25881) and [TIMOB-25868](https://jira.appcelerator.org/browse/TIMOB-25868)

### [TIMOB-25881](https://jira.appcelerator.org/browse/TIMOB-25881) Implement `WebView.blacklistedURLs`.

```js
﻿var win = Titanium.UI.createWindow();

var webview = Titanium.UI.createWebView({
    blacklistedURLs: [
        'http://www.appcelerator.com/'
    ]
});

    webview.url = 'http://www.appcelerator.com/';
});
webview.addEventListener('blacklisturl', function (e) {
    alert('WebView blocked loading ' + e.url);
});

win.add(webview);
win.open();
```

Expected: 

- WebView should not load the blocked URL
- `blacklisturl` event should be fired 

### [TIMOB-25868](https://jira.appcelerator.org/browse/TIMOB-25868) Implement `WebView.onlink callback`

```js
﻿var window = Ti.UI.createWindow(),
    webView = Ti.UI.createWebView({
        url: 'https://en.wikipedia.org/wiki/Portable_Document_Format',
        onlink: function (e) {
            if (e.url.endsWith('.pdf')) {
                alert('PDF: ' + e.url);
                return false;
            }
            return true;
        }
    });

window.add(webView);
window.open();
```

Expected: 

- Clicking PDF link should alert PDF URL
- WebView should not load the PDF URL because `onLink` callback returns false.

Known issue: Navigating web link that opens new Window (`_blank` etc) doesn't fire `onLink` callback. That is because native WebView component opens new browser window that's outside of control of the app and we can't handle the instance.


ssjsamir:
@infosia For the second test case the wiki page for .PDF loads straight away on emulator?  
I see the following message:
"Redirected from "Portable document format" "

PR #1213, opened 2018-03-23
[TIMOB-25895] Rename HAL to AXWAYHAL

infosia:
[TIMOB-25895](https://jira.appcelerator.org/browse/TIMOB-25895)

Apparently Microsoft start rejecting our `HAL.DLL` in WACK (Windows Apps Certification Kit) just because Windows system has `HAL.DLL` in system library. We rename `HAL.DLL` to `AXWAYHAL.DLL` in order to make it pass the cert.

Expected: The built package should not contain `HAL.DLL`.


ewanharris:
Merging as [per comment in backport](https://github.com/appcelerator/titanium_mobile_windows/pull/1214#issuecomment-377398039)

PR #1214, opened 2018-03-26
[7_1_X][TIMOB-25895] Rename HAL to AXWAYHAL 

infosia:
Cherry-pick #1213 for 7_1_X


ewanharris:
After chatting with Kota, we need to merge this in order to prep the module builds so merging for now and testing can be performed with the modules

PR #1227, opened 2018-04-11
Try to update

ewanharris:


Jenkinsfile
```diff
@@ -16,7 +17,7 @@ def build(sdkVersion, msBuildVersion, architecture, gitCommit, nodeVersion) {
 	bat 'mkdir dist\\windows'
 
 	// nodejs(nodeJSInstallationName: "node ${nodeVersion}") {
-	// 	bat 'npm install -g npm@5.4.1' // Install NPM 5.4.1
+	bat "npm install -g npm@${npmVersion}"
 	// 	def nodeHome = tool(name: "node ${nodeVersion}", type: 'nodejs')
 	// 	echo nodeHome
 	// 	bat "netsh advfirewall firewall add rule name=\"Node ${nodeVersion}\ TCP" program=\"${nodeHome}\\node.exe\" dir=in action=allow protocol=TCP"
```
sgtcoolguy: I'm not sure why this didn't work for me. Looking at what chef is doing, I am missing `description ="Firewall rule" profile=any interfacetype=any` as additional arguments I think it passes.

Jenkinsfile
```diff
@@ -16,7 +17,7 @@ def build(sdkVersion, msBuildVersion, architecture, gitCommit, nodeVersion) {
 	bat 'mkdir dist\\windows'
 
 	// nodejs(nodeJSInstallationName: "node ${nodeVersion}") {
```
sgtcoolguy: We should uncomment this and get the Jenkins tool-based node usage working.

Right now Windows SDK relies on the system installed Node, which means it's a chef change/manual change to use a different node version. The reason I backed off here is because while using it worked fine for the build itself, it broke our unit testing - because we need to punch "holes" in the firewall rules so Node (the unit test CLI/logging) can talk to the phone emulator. Otherwise the phone logs never get communicated to the CLI. (it should pop a firewall GUI asking the user to allow or not, but obviously we're not running in an interactive way here)

Jenkinsfile
sgtcoolguy: Also note that this is probably the wrong spot to run this command. I think we only need it on the unit testing portion (so around line 54)

Jenkinsfile
```diff
@@ -6,51 +6,54 @@ def gitCommit = ''
...
-	unarchive mapping: ['dist/' : '.'] // copy in built SDK from dist/ folder (from Build stage)
+	// unarchive mapping: ['dist/' : '.'] // copy in built SDK from dist/ folder (from Build stage)
 	unstash 'sources'
-	// nodejs(nodeJSInstallationName: "node ${nodeVersion}") {
-		// bat 'npm install -g npm@5.4.1' // Install NPM 5.4.1
+	nodejs(nodeJSInstallationName: "node ${nodeVersion}") {
+		// bat "npm install -g npm@${npmVersion}"
+		def nodeHome = tool(name: "node ${nodeVersion}", type: 'nodejs')
+		echo nodeHome
+		bat "netsh advfirewall firewall add rule name=\"Node ${nodeVersion} All\" program=\"${nodeHome}\\node.exe\" dir=in action=allow"
```
sgtcoolguy: Looking at one of the build machines: This rule exists twice and has no program attached to it, so I don't think it will work.

There's valid TCP rules attached to the node exe on the node, but no valid UDP rule.

I think the log relay uses UDP, so we need a UDP rule specifically? And you may need to delete any existing rules under the given name as you may have created bad rules for the same name and the new/updated rule doesn't get set.

sgtcoolguy:
@ewanharris I spent quite a lot of time on this today. Basically it boils down to this:
I had to generate rules for tcp and ump separately under the same name before Windows handled it/displayed it as it would have if I manually added the application to the firewall listing. If I generated a TCP or UDP rule separately it seemed to just fail.

So I narrowed it down to manually executing these commands in an admin cmd.exe prompt on a node and saw that the unit tests worked:
```
netsh advfirewall firewall add rule name="Node 8.9.1" program="C:\Jenkins\tools\jenkins.plugins.nodejs.tools.NodeJSInstallation\node_8.9.1\node.exe" dir=in action=allow protocol=udp description="Firewall rule"
netsh advfirewall firewall add rule name="Node 8.9.1" program="C:\Jenkins\tools\jenkins.plugins.nodejs.tools.NodeJSInstallation\node_8.9.1\node.exe" dir=in action=allow protocol=tcp description="Firewall rule"
```

PR #1237, opened 2018-05-16
[7_2_X] Sync with 7_1_X codebase

infosia:
The branch `7_2_X` is not quite in sync with `7_1_X` codebase. Pushing the deltas.


garymathews:
@infosia Are any of these missing from `master`?

PR #1238, opened 2018-05-17
[TIMOB-26005] Support opt-out of Analytics at runtime

infosia:
[TIMOB-26005](https://jira.appcelerator.org/browse/TIMOB-26005)

```js
Ti.API.info('1. Checking Ti.Analytics.optedOut=' + Ti.Analytics.optedOut); // This should preserve last state, by default it's false
Ti.API.info('2. Checking Ti.Analytics.getOptedOut()=' + Ti.Analytics.getOptedOut()); // This should preserve last state, by default it's false
Ti.Analytics.setOptedOut(true); // This should print "Saving Ti.Analytics.getOptedOut=true" in debug log
Ti.API.info('3. Checking Ti.Analytics.optedOut=' + Ti.Analytics.optedOut); // should return true
Ti.API.info('4. Checking Ti.Analytics.getOptedOut()=' + Ti.Analytics.getOptedOut()); // should return true
Ti.Analytics.featureEvent('app.feature'); // This should print 'Skipping Ti.Analytics.postEvents...' in debug log
Ti.Analytics.optedOut = false; // This should print "Saving Ti.Analytics.optedOut=false" in debug log
Ti.API.info('5. Checking Ti.Analytics.optedOut=' + Ti.Analytics.optedOut);  // should return false
Ti.API.info('6. Checking Ti.Analytics.getOptedOut()=' + Ti.Analytics.getOptedOut()); // should return false
Ti.Analytics.featureEvent('app.feature'); // This should print nothing
Ti.Analytics.optedOut = true;  // This should print "Saving Ti.Analytics.optedOut=true" in debug log
Ti.API.info('7. Checking Ti.Analytics.optedOut=' + Ti.Analytics.optedOut); // should return true
Ti.API.info('8. Checking Ti.Analytics.getOptedOut()=' + Ti.Analytics.getOptedOut()); // should return true
Ti.Analytics.featureEvent('app.feature'); // This should print 'Skipping Ti.Analytics.postEvents...' in debug log
Ti.Analytics.optedOut = false;  // This should print "Saving Ti.Analytics.optedOut=false" in debug log

// After all of above, then you should see "Start posting initial Ti.Analytics events" in debug log.
```


garymathews:
Will this send events before `app.js` has ran? (before the user can optOut)

infosia:
@garymathews  Right now Titanium sends first event when `Ti.Analytics` object is created, before developer can opt out. I'll delay the post (maybe after initial `Window.open`?) so developer can get a chance to create/display opt-out prompt.

garymathews:
I think we need to agree on the API specification for Ti.Analytics optOut because this is varying per implementation:
```
Ti.Analytics.optOut // instead of Ti.Analytics.isOptOut
Ti.Analytics.setOptOut(boolean)
Ti.Analytics.getOptOut()
```

For consistency like our other Titanium APIs

For example, the Android implementation uses `optedOut`:
https://github.com/appcelerator/titanium_mobile/pull/10058

infosia:
Yeah good point, Hans recently suggested following and we'll go with it.

```
Ti.Analytics.optedOut = true|false;
Ti.Analytics.setOptedOut(true|false);
Ti.Analytics.isOptedOut();
```

infosia:
@garymathews Updated.

infosia:
Oh wait, I forgot to "delay" the initial enrollment...working on it :|

infosia:
Implemented, should be good to review now.

hansemannn:
@infosia Sorry for the confusion here! Going with `optedOut` + getter `getOptedOut()` and setter `setOptedOut(true|false)`. The `isOptedOut()` did not seem to get much approval, so we removed that one.

infosia:
@hansemannn ok sounds good, updated.

apidoc/Titanium/Analytics.yml
```diff
@@ -3,11 +3,16 @@ platforms: [windowsphone]
...
     platforms: [windowsphone]
   - name: navEvent
     platforms: [windowsphone]
   - name: getLastEvent
     platforms: [windowsphone]
-
+  - name: getOptedOut
+    platforms: [windowsphone]
+  - name: setOptedOut
+    platforms: [windowsphone]
```
hansemannn: On iOS and Android, these are automatically generated by the API docs generator (unless we specify `accessors: false`. Do you know if that also applies for Windows?

infosia: For Windows, my understanding is that we are still using [Windows-specific custom docs generator](https://github.com/appcelerator/titanium_mobile_windows/blob/master/apidoc/ti_win_yaml.js). But I'm not quite sure actually, is this still effective, @garymathews ?

infosia:
@hansemannn I am under impression that we should not throw JS exception from Analytics when end-user has opted-out. It is because Analytics events tend to happen occasionally along with application events such as `resume` and `suspend`, and developers don't have full control of Analytics events actually. I am thinking that just ignoring them will be most convenient for developers.

infosia:
For the `aps_sdk`, we are not using the aps code on Titanium-Windows because the [code for Windows](https://github.com/appcelerator/aps_sdk/tree/master/windows) is written in C#, which is not the main language we are using in `titanium_moble_windows` that is written in C++/CX.  I actually didn't know much about it, seems like it is originally written for Windows 8.1 and is not maintained for nearly 3 years...Axway doesn't officially support APS SDK for Windows [from what I can see in the docs](https://docs.axway.com/bundle/AMPLIFY_Appcelerator_Services_allOS_en/page/amplify_appcelerator_services_native_sdks.html).

hansemannn:
Do they? For iOS at least, they do not I think. If so, that is most likely undocumented. But in that case, I am okay with not throwing and only ignoring them. cc @ypbnv

infosia:
I can see that Windows posts `ti.foreground` and `ti.background` events when application is resumed/suspended. [analytics.js#L280](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/TitaniumKit/src/analytics.js#L280). I thought they are "official" events for Analytics but seems like it's not documented.

apidoc/Titanium/Analytics.yml
garymathews: It gets exactly what's defined in the source, so if there is a setter/getter it will document it

PR #1242, opened 2018-05-21
Attempt to clean-up test build directory

infosia:
Mocha test needs to cleanup `C:\Users\build\.titanium\vsbuild\mocha` before build because previous build files are there.


infosia:
Closing, turns out this has to be done in [titanium-mobile-mocha-suite](https://github.com/appcelerator/titanium-mobile-mocha-suite).

PR #1244, opened 2018-05-22
Attempt to fix build issue on master

infosia:
This seems to me that sdk installation has been skipped. Why is this skpped?


infosia:
@ewanharris I noticed Jenkins has been skipping to copy built-sdk which causes build errors. I believe I have fixed it by re-enabling some functions, let me know if I did it wrong, just in case.

ewanharris:
@infosia Yes that looks to be correct, it was commented out to speed the feedback loop up when testing the `nodejsinstallation` changes and must have been forgotten about. Apologies for that!

PR #1246, opened 2018-05-23
Attempt to update test scripts for 7_2_X

infosia:
The Jenkins build for 7_2_X has been failing. Trying to update Jenkins-related scripts.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,5 @@
-#!groovy​
+#!groovy
+library 'pipeline-library'
 // Keep history of up to 30 builds, but only keep artifacts for the last 5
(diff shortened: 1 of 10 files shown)


ewanharris:
@infosia So the problem was that [this commit](https://github.com/appcelerator/titanium-mobile-mocha-suite/commit/0896e8086a54bd962385dfd2ba201c8f49f5883f) was missing from the mocha-suite so we were running into [TIMOB-25827](https://jira.appcelerator.org/browse/TIMOB-25827). I cherry-picked it across and after a few tries it looks to have passed 🎆

infosia:
@ewanharris  Awesome thanks for following it up ❗️ 👍

PR #1248, opened 2018-05-30
[TIMOB-26048] Refactor RSOD UI

infosia:
[TIMOB-26048](https://jira.appcelerator.org/browse/TIMOB-26048)

- Refactor the red-screen of death (crash dialog) with stack traces to match the iOS/Android look and feel.

Note: On Windows we does not provide file name and file number where exception happens because Windows JavaScriptCore does not actually returns detailed information about where it happens. We are just providing the error message and stack traces.

```js
win.addEventListener('open', function (e) {
    win.add(null);
});
win.open();
```

Expected:

timob-26048

- `add@[native code]` strings are the `stack` value that JavaScriptCore returns.
- `1  JSExportClass<class TitaniumWindows::UI::Window>::add` strings are the `nativeStack` values that represents Titanium module stack traces.

### DIFF
--- a/Source/TitaniumKit/src/Application.cpp
+++ b/Source/TitaniumKit/src/Application.cpp
@@ -32,44 +32,51 @@ function Titanium_RedScreenOfDeath(e) {
 
     try {
         var win = Ti.UI.createWindow({
-                backgroundColor: "#f00",
+                backgroundColor: "#f4f4f4",
                 layout: "vertical"
             }),
-            view,
             button;
 
         function makeMessage(e) {
             return (e.fileName ? "In " + e.fileName + " " : "") + (e.message || e.toString()).trim() + (e.lineNumber ? " (line " + e.lineNumber + " column " + e.columnNumber + ")" : "");
-        } 
-        function makeLabel(text, height, color, fontSize) {
-            var label = Ti.UI.createView({
-                height: height,
-                width: Ti.UI.FILL
-            });
-            label.add(Ti.UI.createLabel({
-                color: color,
-                font: { fontSize: fontSize, fontWeight: "bold" },
-                width:  Ti.UI.FILL,
-                height: Ti.UI.FILL,
-                textAlign: Ti.UI.TEXT_ALIGNMENT_CENTER,
-                text: text
-            }));
-            win.add(label);
         }
 
-        Ti.API.error("----- Titanium Javascript Runtime Error -----");
         Ti.API.error("Message: Uncaught Error: " + makeMessage(e));
+        if (e.nativeStack) {
+            Ti.API.error(e.nativeStack);
+        }
 
-        win.add(view = Ti.UI.createView({ height: "12%" }));
-        makeLabel("Application Error", "15%", "#0f0", "34");
-        makeLabel(makeMessage(e), "45%", "#fff", "26");
-        win.add(view = Ti.UI.createView({ height: "12%" }));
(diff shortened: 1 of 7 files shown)


infosia:
Related to appcelerator/titanium_mobile#10022 and appcelerator/titanium_mobile#10003

sgtcoolguy:
@garymathews @ewanharris Ideally this should have been scheduled for 7.3.0 to have parity with iOS/Android. Any chance we could get this CR'd/FR'd and then we can cherry-pick to 7_3_X?

Source/TitaniumKit/src/Application.cpp
```diff
@@ -32,45 +32,55 @@ function Titanium_RedScreenOfDeath(e) {
 
     try {
         var win = Ti.UI.createWindow({
-                backgroundColor: "#f00",
+                backgroundColor: "#f4f4f4",
```
garymathews: Could we set the background to `#F5F5F5`

Source/TitaniumKit/src/Application.cpp
```diff
@@ -32,45 +32,55 @@ function Titanium_RedScreenOfDeath(e) {
...
-                height: height,
-                width: Ti.UI.FILL
-            });
+            return "Runtime Exception: " + (e.fileName ? "In " + e.fileName + " " : "") + (e.message || e.toString()).trim() + (e.lineNumber ? " (line " + e.lineNumber + " column " + e.columnNumber + ")" : "");
+        }
+        function makeLabel(label, str) {
             label.add(Ti.UI.createLabel({
-                color: color,
-                font: { fontSize: fontSize, fontWeight: "bold" },
+                color: 'red',
```
garymathews: Text should be `#E53935`

Source/TitaniumKit/src/Application.cpp
```diff
@@ -32,45 +32,55 @@ function Titanium_RedScreenOfDeath(e) {
...
+            makeLabel(label, e.nativeStack);
+        }
+        win.add(label);
 
-        win.add(view = Ti.UI.createView({ height: "12%" }));
-        makeLabel("Application Error", "15%", "#0f0", "34");
-        makeLabel(makeMessage(e), "45%", "#fff", "26");
-        win.add(view = Ti.UI.createView({ height: "12%" }));
-        view.add(button = Ti.UI.createButton({ title: "Dismiss" }));
+        win.add(button = Ti.UI.createButton({ title: "CONTINUE", backgroundColor: 'red', width: Ti.UI.FILL, height: '10%', bottom: 0 }));
```
garymathews: Button should be `#E53935`

garymathews:
Also, we should use `Lucida Console` as the font since it's similar to the font on other platforms and is monospaced.

garymathews:
I'm sure we can obtain the Javascript stacktrace? Any exceptions we get from an evaluation should contain what we need?

infosia:
Updated the colors.

@garymathews 
> I'm sure we can obtain the Javascript stacktrace; any exceptions we get from an evaluation should contain what we need?

Do you mean `stack` property in the exception? Interestingly JavaScriptCore returns following values on Windows for `stack` property and this is what we shows in the dialog. I don't see any other properties we can use for JSCore stacktrace.

```
open@[native code]
global code
require@[native code]
global code
```

infosia:
FR passed, still waiting for final approval from @garymathews before 7_3_X code freeze.

PR #1252, opened 2018-06-06
[TIMOB-25986] Add physical pixels support 

infosia:
EXPERIMENTAL FEATURE

[TIMOB-25986](https://jira.appcelerator.org/browse/TIMOB-25986)

Introduce new unit specifier `ppx` to give ability to layout with physical pixels on Windows.

## Background

On Windows, the default unit specifier `px` is used as _effective pixels_, not actual physical pixels. Effective pixels are a virtual unit of measurement, and they're used to express layout dimensions and spacing, independent of screen density. (See [Introduction to UWP app design](https://docs.microsoft.com/en-us/windows/uwp/design/basics/design-and-ui-intro) for details). So when Microsoft uses the term "pixels" it doesn't mean actual physical pixels and that's why Titanium is using this effective pixels when you use "px" unit specifier. 

In order to support layout with actual physical pixels, we are introducing new unit specifier `ppx` (physical pixels) that is specific to Windows.

## Use Cases

Easiest way to deal with physical pixels is setting `ppx` for `i.ui.defaultunit` property in `tiapp.xml`. In this case layout will be done with physical pixels.

```xml
  <property name="ti.ui.defaultunit" type="string">ppx</property>
```

```js
    width: 1920, height: 1400, backgroundColor: 'red'
});

var view = Ti.UI.createView({
    width: 1916, height: 1396, backgroundColor: 'white'
});

win.add(view);
win.open();
```

You can also specify unit specifier explicitly.

```js
    width: `1920ppx`, height: `1400ppx`, backgroundColor: 'red'
});

var view = Ti.UI.createView({
    width: `1916ppx`, height: `1396ppx`, backgroundColor: 'white'
});

win.add(view);
win.open();
```

### DIFF
--- a/Source/LayoutEngine/include/LayoutEngine/LayoutEngine.hpp
+++ b/Source/LayoutEngine/include/LayoutEngine/LayoutEngine.hpp
@@ -21,6 +21,8 @@ namespace Titanium
 {
 	namespace LayoutEngine
 	{
+		extern double PhysicalPixelsFactor;
+
 		struct Rect
 		{
 			double width = 0;
--- a/Source/LayoutEngine/src/ParseProperty.cpp
+++ b/Source/LayoutEngine/src/ParseProperty.cpp
@@ -16,6 +16,8 @@ namespace Titanium
 {
 	namespace LayoutEngine
 	{
+		double PhysicalPixelsFactor = 1.0;
+
 		enum ValueType _getValueType(const std::string& value)
 		{
 			if (value == "UI.SIZE") {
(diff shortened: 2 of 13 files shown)


sgtcoolguy:
Here's a very relevant PR from titanium_mobile for testing Ti.UI.convertUnits: appcelerator/titanium_mobile#10080

Seems like we're using `px` on Windows more like `dp`/`dip` was intended to be used.

infosia:
Right, originally our layout engine is alculating `dp` based on Android-like measurement  ([LayoutEngine/src/ParseProperty.cpp#L72](https://github.com/appcelerator/titanium_mobile_windows/blob/master/Source/LayoutEngine/src/ParseProperty.cpp#L72)). We might be able to change the meaning of it, but I assume we can't because it should introduce huge breaking change on existing projects (this relates to [ti.ui.defaultunit support](#1163) too)....So easier way to deal with this is to introduce new unit specifier for Windows like this PR.

PR #1257, opened 2018-06-13
[TIMOB-26117] Include resources path

garymathews:
- Required for https://github.com/appcelerator/node-titanium-sdk/pull/32

[JIRA Ticket](https://jira.appcelerator.org/browse/TIMOB-26117)

### DIFF
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -662,7 +662,8 @@ function copyResources(next) {
 								transpile: this.transpile,
 								targets: {
 									safari: '10' // matches the version of jscore we use
-								}
+								},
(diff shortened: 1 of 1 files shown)


infosia:
How do I test this?

cli/commands/_build/copy.js
```diff
@@ -662,7 +662,8 @@ function copyResources(next) {
 								transpile: this.transpile,
 								targets: {
 									safari: '10' // matches the version of jscore we use
-								}
+								},
+								resourcesDir: path.join(this.buildDir, 'bin', 'assets', 'Resources')
```
ewanharris: I think this should be `path.join(this.buildDir, 'Assets')`

PR #1284, opened 2018-09-12
[TIMOB-25983] Ti.UI.TextArea is not resized when its size is not specified

infosia:
[TIMOB-25983](https://jira.appcelerator.org/browse/TIMOB-25983)

`Ti.UI.TextArea` should resize its size when it is not specified. Revised #1230 because it did not consider `value` property.

```js
    backgroundColor: 'green',
});

var parent = Ti.UI.createView({
    width: '60%',
    height: '60%',
    backgroundColor: 'gray'
});
var textArea = Ti.UI.createTextArea({
    width: Ti.UI.FILL,
    backgroundColor: 'white',
    color: 'black',
    left: 40,
    right: 40,
    hintText: 'this is a long hint that is not wrapping in windows version. Not sure what is the problem. All text show in single line instead of multi line. please check',
    value: 'this is a long text that is not wrapping in windows version. Not sure what is the problem. All text show in single line instead of multi line. please check',
});

parent.add(textArea);
win.add(parent);
win.open();
```

Expected: TextArea should not be shown in single line. It should have multiple lines and texts should be wrapped.

timob-25983

### DIFF
--- a/Source/UI/include/TitaniumWindows/UI/TextArea.hpp
+++ b/Source/UI/include/TitaniumWindows/UI/TextArea.hpp
@@ -80,7 +80,7 @@ namespace TitaniumWindows
 #pragma warning(push)
 #pragma warning(disable : 4251)
 			std::unordered_map<std::string, std::string> custom_fonts__;
-			bool dirty__ { false };
+			bool isLoaded__ { false };
 #pragma warning(pop)
 			Windows::UI::Xaml::Controls::Border^ border__;
 			Windows::UI::Xaml::Controls::TextBox^ text_box__;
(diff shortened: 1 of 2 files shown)


keerthi1032:
FR Passed.Texts displayed in multiple lines .waiting for CR and Merge
Test Environment:
Operating System
  Name                        = Microsoft Windows 10 Pro
  Version                     = 10.0.17134
  Architecture                = 64bit
Node.js
  Node.js Version             = 8.9.1
  npm Version                 = 5.5.1
Titanium CLI
  CLI Version                 = 5.1.1
Titanium SDK
  SDK Version                 = local sdk
  device		      =Lumia 650 dual sim windows 10

PR #1285, opened 2018-09-20
[TIMOB-26400] Add JavaScript app bootstrapping support

infosia:
[TIMOB-26400](https://jira.appcelerator.org/browse/TIMOB-26400)

Added bootstrap script support.
  * All files named `bootstrap.js` or end with name `*.bootstrap.js` under the "Resources" directory will be automatically required-in before `app.js` gets loaded. This is a recursive search.

**ui.bootstrap.js**
```javascript
exports.execute = function(finished) {
	var dialog = Ti.UI.createAlertDialog({
		title: 'Alert Dialog',
		message: 'This dialog was launched by a bootstrap script.',
		buttonNames: ['OK'],
		cancel: 0,
	});
	dialog.addEventListener('click', function(e) {
		finished();
	});
	dialog.show();
};
```


sgtcoolguy:
Windows equivalent of appcelerator/titanium_mobile#10112 (which is now merged)

sgtcoolguy:
@jquick-axway May want to take a quick once-over look to confirm that Kota's implementation looks/operates how yours does (and should work).

jquick-axway:
Kota, your changes look good. I just have a few questions/concerns...

1. For you "Example\NG" project, I'm a little worried that we'll break it in the future. Especially when we add new JS files to the "ti.internal/extensions" directory. Are we able to copy the "titanium_mobile/common/Resources" file tree as a pre-built step instead like we do in Xcode [here](https://github.com/appcelerator/titanium_mobile/pull/10112/files#diff-4595477f31d142c0d5a3e3b1a5385b82)?
2. I've noticed [here](https://github.com/appcelerator/titanium_mobile_windows/pull/402/files) that you had to add special handling to support Alloy globals declared in the "app.js". Do you think this will still work?
3. Unfortunately, I discovered a **regression** with my merged bootstrap changes. See: [TIMOB-26400](https://jira.appcelerator.org/browse/TIMOB-26400). Local `var` declared variables in the "app.js" are supposed to be globally available to all other JS files. That's broken now (I wasn't aware that this was a JS "feature"). This is related to `#2` above. I don't have an answer for this one yet.

infosia:
@jquick-axway 

1. Yeah, I understand the downside but `titanium_mobile_windows` is completely separated from `titanium_mobile` unfortunately. The `NG` project is only the local project that I use for debugging and I think it's ok for now because we integrate whole tests in `titanium-mobile-mocha-suite` and we can find bugs if anything wrong happens.

2. I think #402 is happening again because of the regression described in 3. I'm now trying to find a workaround.

jquick-axway:
Regarding the regression, you can reproduce it in a simple Classic app as well. Just add the following to the "app.js". With what's merged now on master, `global.test` will be undefined.
```javascript
var test = "Hello World";
alert("global.test: " + global.test);
```

infosia:
Pushed a fix for the regression.

`app.js`

``` javascript
var Alloy = { Globals:{} };
Alloy.Globals.winTop = 6;

var m = require('testmodule')
m.test();

var test = "Hello World";
alert("global.test: " + global.test);
```

`testmodule.js`

``` javascript
exports.test = function() {
    Ti.API.info("Test Module");
    Ti.API.info("Module Wintop " + Alloy.Globals.winTop);
};
```

Expected: app should run without errors and it should show `Hello, World`.

jquick-axway:
@infosia, we can implement a similar solution in our "ti.main.js" file.
https://github.com/appcelerator/titanium_mobile/blob/master/common/Resources/ti.main.js

Instead of doing this in "ti.main.js"...
```javascript
require('./app');
```

We can do an `eval()` of the "app.js" using the global scope. This effectively does the same thing you are doing in C++ on Windows. I've tested this on Android and iOS... and it works. _(Is this naughty? Should I feel dirty about this?)_
```javascript
eval.call(global, Ti.Filesystem.getFile('app.js').read().text + '\n//# sourceURL=app.js');
```

Note that I need to append the `'\n//# sourceURL=app.js'` to the end of the "app.js" source string. This is metadata for the JavaScript engine which is used in stack-traces and for the JS debugger/breakpoints. Based on my testing, it definitely works for stack-traces, but I haven't confirmed that it works with the debugger yet. @infosia, if breakpoints are not working within your "app.js" on Windows (I haven't tried), then this might be worth looking into. Also, I've confirmed that this metadata does not change the `__filename` variable, but we've already hardcoded this natively to "app.js" on all 3 platforms.

So, if I use the above `eval()` call with the following "app.js" code below...
```javascript
var test = "Hello World";
Ti.API.info("global.test: " + global.test);
Ti.API.info("'app.js' __dirname: " + __dirname);
Ti.API.info("'app.js' __filename: " + __filename);
Ti.API.info("'app.js' error: " + JSON.stringify(new Error('uh-oh')));
```

On Android, I get the following output. Note that the stack-trace reveals that the "app.js" has gone through an `eval()` function.
```
global.test: Hello World
'app.js' __dirname: /
'app.js' __filename: /app.js
'app.js' error: {
	"stack":"Error: uh-oh
	    at eval (app.js:5:53)
	    at eval (<anonymous>)
	    at /ti.main.js:31:7
	    at /ti.internal/bootstrap.loader.js:152:3
	    at doLoad (/ti.internal/bootstrap.loader.js:137:4)
	    at Titanium.<anonymous> (/ti.internal/bootstrap.loader.js:144:5)",
	"message":"uh-oh"
}
```

Unfortunately, on iOS, the stack-trace does not show "app.js". It reads `eval@[native code]` instead. Kind of bummed by that. Might need to play around with it a bit more.
```
error: {
	"message":"uh-oh",
	"line":6,
	"column":62,
	"stack":"eval code
		eval@[native code]
		file:///Users/jquick2/Library/Developer/CoreSimulator/Devices/AD6CBD12-585F-4136-A967-088D2D841D6E/data/Containers/Bundle/Application/301D4B20-AD63-47BA-A3FE-738333E86B16/ClassicAppTest.app/ti.main.js:31:11
		file:///Users/jquick2/Library/Developer/CoreSimulator/Devices/AD6CBD12-585F-4136-A967-088D2D841D6E/data/Containers/Bundle/Application/301D4B20-AD63-47BA-A3FE-738333E86B16/ClassicAppTest.app/ti.internal/bootstrap.loader.js:152:11
		doLoad@file:///Users/jquick2/Library/Developer/CoreSimulator/Devices/AD6CBD12-585F-4136-A967-088D2D841D6E/data/Containers/Bundle/Application/301D4B20-AD63-47BA-A3FE-738333E86B16/ClassicAppTest.app/ti.internal/bootstrap.loader.js:137:12
		file:///Users/jquick2/Library/Developer/CoreSimulator/Devices/AD6CBD12-585F-4136-A967-088D2D841D6E/data/Containers/Bundle/Application/301D4B20-AD63-47BA-A3FE-738333E86B16/ClassicAppTest.app/ti.internal/bootstrap.loader.js:144:11"
}
```

I'm not sure what else we can do to get the "app.js" local `var` variables into the global namespace. What LiveView does to solve this issue seems even naughtier by stripping out all of the `var` keywords from the "app.js" (yikes).
https://github.com/appcelerator/liveview/blob/master/lib/fserver.js#L298-L317

I'm also somewhat conflicted about this JS "feature" since I wouldn't want local variables polluting the global namespace. But it is a breaking-change that we need to avoid. At least for now.

@sgtcoolguy, do you have any feedback?
_(Sorry about the long post.)_

infosia:
On Windows, `__dirname` and `__filename` is hardcoded in `require` so in order to emulate it using `eval` I ended up defining them explicitly like following;

```js
eval.call(global, 'var __dirname="/", __filename="app.js";' + Ti.Filesystem.getFile('app.js').read().text + '\n//# sourceURL=app.js');
```

Then I got following on Windows with @jquick-axway 's sample code, which looks good to me.

```
[INFO] global.test: Hello World
[INFO] 'app.js' __dirname: /
[INFO] 'app.js' __filename: app.js
[INFO] 'app.js' error: {"message":"uh-oh","line":6,"column":58,"stack":"eval code\neval@[native code]\n/ti.main.js:28:14\nloadAsync@/ti.internal/bootstrap.loader.js:106:11\n/ti.main.js:24:52\nglobal code@/ti.main.js:42:8\nrequire@[native code]\nglobal code"}
```

jquick-axway:
I know we have to implement the `require()` function ourselves on every platform. I'd rather leverage that code than go through an `eval()` if possible. So, I want to see if I can leverage that solution instead. Perhaps by adding an extra argument to provide the scope (in this case `global`).
https://github.com/appcelerator/titanium_mobile/blob/master/android/runtime/common/src/js/module.js
https://github.com/appcelerator/titanium_mobile/blob/master/iphone/Classes/KrollContext.m

I'll sleep on it for now. I don't want to make any rash decisions. Thanks for your help.

Source/TitaniumKit/src/GlobalObject.cpp
```diff
@@ -331,7 +331,7 @@ namespace Titanium
 				//
 				// app entry point should not be treated as "CommonJS module". It should expose every variables to children.
 				//
-				if (moduleId == "/app") {
+				if (module_path == "/app.js") {
```
sgtcoolguy: We're trying to fix this in the CLI and in Alloy, so that app.js is treated specially for now until SDK 9.0.0 and a babel plugin will hang top-level vars/functions off global for the user under the hood. We're deprecating that implicit behavior in SDK 8.0.0.

So I think going forward, we should eliminate this branch of code that does a straight eval. And then when we officially remove support for this, we'll just drop the babel plugin from the CLI build.

infosia:
mocha test app on Jenkins keeps crashing. I'm looking into it.

PR #1288, opened 2018-10-03
[TIMOB-26403] Pass logger to analyzeJs call

ewanharris:
JIRA: jira.appcelerator.org/browse/TIMOB-26403

- Pass logger into to analyzeJs so we can log deprecation warning
To test:

Build kitchensink-v2 to device or simulator
You should see a warning log about "implicit global scope being deprecated
It should launch successfully


sgtcoolguy:
Relates to #1285 - once adding bootstrap support, we lost the effect of top-level app.js variables being implicitly assigned to global properties/scope. We've updated node-titanium-sdk so that we run a babel plugin to explicitly assign top-level vars/functions to global properties now, and this PR passes along our Cli logger object so that when we do so, the plugin can spit out a warning about implicit globals being deprecated.

PR #1297, opened 2018-10-29
[TIMOB-26464] Add Ti.UI.Window.safeAreaPadding and extendSafeArea

infosia:
[TIMOB-26464](https://jira.appcelerator.org/browse/TIMOB-26464)

Add `safeAreaPadding` and `extendSafeArea` property for `Ti.UI.Window`.

On Windows, we just hard-code the padding to all zeros because safe are is just not supported on Windows. I wouldn't update the API doc to state these properties are "supported" because these properties only intend to be a placeholder on Windows just to make sure use of these properties doesn't cause runtime errors.

```js
Ti.API.info(JSON.stringify(win.safeAreaPadding));
Ti.API.info(win.extendSafeArea);
```

Expected: `safeAreaPadding` returns empty padding. `extendSafeArea` defaults to `false`.

```
[INFO] {"bottom":0,"left":0,"right":0,"top":0}
[INFO] false
```

### DIFF
--- a/Source/TitaniumKit/CMakeLists.txt
+++ b/Source/TitaniumKit/CMakeLists.txt
@@ -318,6 +318,8 @@ set(SOURCE_UI
   src/UIModule.cpp
   include/Titanium/UI/View.hpp
   src/UI/View.cpp
+  include/Titanium/UI/ViewPadding.hpp
+  src/UI/ViewPadding.cpp
   include/Titanium/UI/ViewLayoutDelegate.hpp
   src/UI/ViewLayoutDelegate.cpp
   include/Titanium/UI/WebView.hpp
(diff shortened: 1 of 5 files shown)


jquick-axway:
Returning an empty padding is perfect.
Also, the "extendSafeArea" property is a creation-only property.

infosia:
Unfortunately on Windows TitaniumKit framwork, creation-only property never work actually. It is forced to be normal read-write property.

jquick-axway:
@infosia, how are we supposed to update the API docs to indicate Windows supports this in 8.0.0? Do we update the yml file in this repo?

infosia:
@jquick-axway Yes, we need to update [Titanium/UI/Window.yml](https://github.com/appcelerator/titanium_mobile_windows/blob/master/apidoc/Titanium/UI/Window.yml) in that case.

ssjsamir:
@infosia Cam be merged once Jenkins is ready.

PR #1300, opened 2018-10-31
[TIMOB-25582] Add support for dp in ti.ui.defaultunit

infosia:
[TIMOB-25582](https://jira.appcelerator.org/browse/TIMOB-25582)

Add support for `dp` in `ti.ui.defaultunit`. This will introduce breaking change on view sizes because our `tiapp.xml` default unit is set to `dp` usually but we've been using `px` for Windows default unit. In order to keep consistency between old versions, you need to set ` <property name="ti.ui.defaultunit" type="string">px</property>` explicitly in `tiapp.xml`.

```js

var view1 = Ti.UI.createView({
    width: "100px", height: "100px",
    top: 100, left: 100,
    backgroundColor: 'blue',
});

var view2 = Ti.UI.createView({
    width: "100dp", height: "100dp",
    top: 200, left: 100,
    backgroundColor: 'red',
});

var view3 = Ti.UI.createView({
    width: "100", height: "100",
    top: 300, left: 100,
    backgroundColor: 'yellow',
});

win.add(view1);
win.add(view2);
win.add(view3);
win.open();
```

Expected: When you set `dp` for `ti.ui.defaultunit` in `tiapp.xml`, the size of red view and yellow view should be same. When you set `px` for `ti.ui.defaultunit` in `tiapp.xml`, the size of blue view and yellow view should be same.


ssjsamir:
@infosia Can be merged once Jenkins is ready

infosia:
Update the doc to state this breaking change: https://wiki.appcelerator.org/display/guides2/UI+Composite+Layout+Behavior+Spec#UICompositeLayoutBehaviorSpec-tiapp.xmlproperties

PR #1303, opened 2018-11-02
Revert "Fix: Label.minimumFontSize does not work (#1295)"

infosia:
DO NOT MERGE.

See if latest Label changes causes [TIMOB-26510](https://jira.appcelerator.org/browse/TIMOB-26510)

This reverts commit 87750d0b6f4d820bb284aef3b438d1b32e4ac9e4.


infosia:
Closing, it turns out the "layout cycle" crash is caused by enclosing `ScrollView`. It is reported that Windows UWP `ScrollViewer` tends to crash with it when vertical scrollbar is not visible.

PR #1304, opened 2018-11-02
Add propertiesSet guard before requesting Label layout

infosia:
DO NOT MERGE.

See if guarding Label layout fixes [TIMOB-26510](https://jira.appcelerator.org/browse/TIMOB-26510)


infosia:
After several tries I was able to reproduce the issue and no of my guard worked for it. Closing.

PR #1305, opened 2018-11-02
Add measureDesiredSize guard to make sure it doesn't cause layout cycle

infosia:
DO NOT MERGE.

Attempt to suppress layout cycle in different way.


infosia:
After several tries I was able to reproduce the issue and no of my guard worked for it. Closing.

PR #1306, opened 2018-11-02
[TIMOB-26510] Workaround "layout cycle" issue

infosia:
Seems like the [Layout cycle detected](https://jira.appcelerator.org/browse/TIMOB-26510) is something related to `ScrollView`. It is reported that Windows UWP `ScrollViewer` tends to crash with this error when its scroll bar is not visible for no reason.


infosia:
So for me this doesn't cause layout cycle error. I believe we safely merge this because use of`ScrollBarVisibility::Visible` instead of `Auto`  doesn't make any visual differences actually.

PR #1307, opened 2018-11-02
[TIMOB-26510] ScrollBar crash with "Layout cycle detected"

infosia:
See if setting minimum height fixes the layout cycle issue.


infosia:
Closing, this doesn't work.

PR #1322, opened 2018-11-15
fix(build): Remove babel requires and unneeded babel traversal step

ewanharris:
Fixes https://jira.appcelerator.org/browse/TIMOB-26564. 

This code is required for hyperloop 2.x which is not supported by SDK 8

This should fix the current build issues on Jenkins fyi @ssjsamir

### DIFF
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -7,9 +7,6 @@ var appc = require('node-appc'),
 	path = require('path'),
 	ti = require('node-titanium-sdk'),
 	wrench = require('wrench'),
-	babylon = require('babylon'),
-	types = require('babel-types'),
-	traverse = require('babel-traverse').default,
 	__ = appc.i18n(__dirname).__;
 
 /*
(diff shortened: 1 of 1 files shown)


cli/commands/_build/copy.js
```diff
@@ -717,7 +649,7 @@ function copyResources(next) {
 								}
 							})(r, from, to, cb);
 						})(from, to, cb);
-					})(from, to, ast, traverse, types, done);
+					})(from, to, done);
```
ewanharris: I think these have to stay to be compatible with 3.x versions of hyperloop. Need to investigate

ewanharris: But if they stay that means we need to bring back babel-*, erugh

sgtcoolguy:
I assume this then requires appcelerator/hyperloop.next#310 to be merged too (and a new hyper loop to be released with 8.0.0)?

ewanharris:
@sgtcoolguy yup, I'm guessing we'll maybe do hyperloop 3.2.0 or hyperloop 4.0.0 with SDK 8.0.0

cli/commands/_build/copy.js
infosia: I think we can remove 3.x compatibility for upcoming 8.0.0 sdk. we should be able to release new version.

PR #1326, opened 2018-11-22
chore(build): Archive WindowsOnly docs

ewanharris:
Fixes https://jira.appcelerator.org/browse/TIMOB-26592.

This allows the Windows specific API namepsaces to be included in the api.jsca files (and similar).

Will test the process locally once the artifacts from this build are produced


ewanharris:
Verified locally that Windows namespaces are in the jsca file

PR #1327, opened 2018-11-28
await fs.readFile returns undefined for some reason.

infosia:
Relates to #1319

Jenkins master build is failing because of `TypeError: Cannot read property 'replace' of undefined` and turns out `await fs.readFile` returns undefined for some reason. Pushing a workaround.


infosia:
@sgtcoolguy Merging this for a workaround because it seems `await fs.readFile` always returns `undefined` on Jenkins.

infosia:
Similar failures happening on `setupSDK.js`. Seems like `fs-extra` functions doesn't really support Promises?

PR #1331, opened 2018-12-05
[TIMOB-24401] View.visible="false" should hide the view

infosia:
[TIMOB-24401](https://jira.appcelerator.org/browse/TIMOB-24401)

Unlike JavaScript standard, setting `"false"` string value should be converted to `false` JS boolean value according to the behavior on Titanium iOS and Android.

```js

var view = Titanium.UI.createView({
   backgroundColor:'red',
   width:100,
   height:100,
});
win.add(view);
var button = Titanium.UI.createButton({
   title: 'Push here to change visible status',
   top: 100,
});
button.addEventListener('click',function(e) {
    view.visible = view.visible ? "false" : "true";
});
win.add(button);
win.open();
```

Expected: Clicking button should change the view visible appearance.

### DIFF
--- a/tests/Resources/ti.ui.view.test.js
+++ b/tests/Resources/ti.ui.view.test.js
@@ -399,6 +399,28 @@ describe('Titanium.UI.View', function () {
 		win.open();
 	});
 
+	// TIMOB-24401: visible="false" should hide the view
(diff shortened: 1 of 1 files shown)


infosia:
Closing, this is merged along with #1334

PR #1345, opened 2019-01-10
[TIMOB-26700] Optimize UI rendering calls

infosia:
[TIMOB-26700](https://jira.appcelerator.org/browse/TIMOB-26700)

- Optimize UI rendering calls

```js
Ti.Analytics.optedOut = true; // Just to make sure networking won't affect the test

    var start = +new Date();
    for (var i = 0; i < 200; i++) {
        win.add(Ti.UI.createLabel({
            width: 200, height: 50, left: i * 10, top: i * 10, backgroundColor: '#0000FF', text: 'Hello, World'
        }));
    }

    // setTimeout() effectively push tasks on UI thread.
    // This can be useful when you need to see actual UI response time.
    setTimeout(function () {
        alert((+new Date() - start) + ' m seconds elapsed');
    }, 0)
});

win.open(); 
```

Locally, I've got average 20% improvement when compared to 7.5.0.GA. I also saw that Jenkins runs unit tests faster than before by 4 mins (10 % faster) which is a good sign too.

```
## 7.5.0.GA
1641 ms
1625 ms
1015 ms
1203 ms
1610 ms
1609 ms
1000 ms

## TIMOB-26700
1328 ms
860 ms
1344 ms
1125 ms
1406 ms
1031 ms
890 ms
```

Also, I can see good results on other practical app from customer (aka PerformanceTestApp). I can see up to **37%** less application startup time and **45%** less UI rendering time.

```
7.5.0.GA

[INFO]  Application startup in 1750 msec
[WARN] [summary/contentView]  Total time to render UI:  2718

8.1.x (with TIMOB-26700)
[INFO]  Application startup in 1110 msec
[WARN] [summary/contentView]  Total time to render UI:  1500
```

Expected: 
- Jenkins should pass all unit tests
- Test code above gets better performance by average when compared to 7.5.0.GA.


Source/Ti/src/App/WindowsModule.cpp
```diff
@@ -22,8 +25,9 @@
   JSObject App_WIN = static_cast<JSObject>(App_WIN_property); \
   JSValue NAME##_property = App_WIN.GetProperty(#NAME); \
   TITANIUM_ASSERT(NAME##_property.IsObject()); \
-  JSObject NAME = static_cast<JSObject>(NAME##_property); \
-  auto NAME##_obj = NAME.CallAsConstructor(parameters); \
+  NAME##_ctor = static_cast<JSObject>(NAME##_property); \
+  });\
+  auto NAME##_obj = NAME##_ctor.CallAsConstructor(); \
```
infosia: Developer note: This caches module constructor so we can avoid calling property getter redundantly. 

Also note that the definition `static auto NAME##_ctor = this_object;` here is just assigning temporary `this_object` only once just because `HAL::JSObject` doesn't allow declaring variable using default constructor such as `static JSObject NAME##_ctor; // HAL doesn't allow this`

Source/Ti/src/DisplayCaps.cpp
```diff
@@ -26,24 +26,32 @@ namespace TitaniumWindows
...
 	{
-		auto display = Windows::Graphics::Display::DisplayInformation::GetForCurrentView();
-		if (display) {
-			return display->LogicalDpi;
-		}
-		return Titanium::Platform::DisplayCaps::dpi();
+		static double value;
+		static std::once_flag of;
+		std::call_once(of, [=] {
+			const auto display = Windows::Graphics::Display::DisplayInformation::GetForCurrentView();
```
infosia: Developer note: We saw accessing to `DisplayInformation` is very slow.

Source/TitaniumKit/src/GlobalObject.cpp
```diff
@@ -327,7 +327,7 @@ namespace Titanium
 			JSValue result = js_context.CreateUndefined();
 			if (boost::ends_with(module_path, ".json")){
 				result = js_context.CreateValueFromJSON(module_js);
-			} else if (js_context.JSCheckScriptSyntax(module_js, moduleId)) {
+			} else {
```
infosia: Developer note: We don't actually need to check script syntax. `JSEvaluateScript` does check it and throws exception after all.

Source/UI/src/Label.cpp
```diff
@@ -166,7 +164,7 @@ namespace TitaniumWindows
 
 			if (get_attributedString()) {
 				set_attributedString(get_attributedString());
-			} else if (!get_autoLink().empty()) {
+			} else if (get_autoLink_enabled()) {
```
infosia: This change disables rich text format when autolink is not enabled. By doing this, UWP TextBlock enables "fast path" which decreases overall memory use and greatly reduces the CPU time to do text measuring and arranging. See also [TextBlock Performance Considerations](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/text-block#performance-considerations).

Source/Filesystem/src/Filesystem.cpp
```diff
@@ -45,34 +45,47 @@ namespace TitaniumWindows
...
-#else
-		TITANIUM_MODULE_LOG_WARN("Filesystem.applicationCacheDirectory is not supported on Windows Store.");
-		return "";
-#endif
+		static std::string value;
+		static std::once_flag of;
+		std::call_once(of, [=] {
+			value = TitaniumWindows::Utility::ConvertString(Windows::Storage::ApplicationData::Current->LocalCacheFolder->Path);
+		});
+		return value + separator(); // FIXME Only append separator if not already there!
```
garymathews: Should we fix this here?
```C++
if (!value.ends_with("/")) {
  value += separator();
}
```

Source/Ti/src/DisplayCaps.cpp
garymathews: Should we cache `display` in the DisplayCaps class so it's not per method?

infosia: Make sense, fixed 👍

Source/Filesystem/src/Filesystem.cpp
infosia: Fixed 👍

garymathews:
@infosia Needs, backporting to `8_0_X`

infosia:
I forgot to mention but you need to turn off `transpile` flag in tiapp.xml when you compare fair performance. `<transpile>false</transpile>` We saw enabling transpile slows down the app but it's a different story.

PR #1346, opened 2019-01-14
Update cmake and usage, speed up SDK builds

sgtcoolguy:
This is a series of updates to our SDK build process. I've tried to separate out the major steps I performed to make review slightly easier.

**Update our cmake usage/organization**
This adds a top-level CMakeLists.txt that drives the sub-directories under `Source`. It also attempts to use more modern cmake practices in defining the projects and dependencies between them.

**Update to latest CMake**
This updates cmake to 3.14.0-rc1. The latest GA 3.13.4 has a bug with WinRT project references which broke app builds. This update is to enable use to further modernize our project definitions in the future, as well as hopefully improve performance (which we may need to do further tweaks to our build process to have cmake drive msbuild as I did here: https://github.com/appcelerator/HAL/pull/90/commits/e8741b3b62ab81e0ff0135f06ce51638782b046b#diff-58231b16fdee45a03a4ee3cf94a9f2c3R19

**Make use of pre-compiled headers**
This speeds up the compilation during clean builds. It's possible it may lead to longer incremental dev builds. I tried to balance how much of the library headers were referenced. If we reference none, then we don't need to rebuild a whole project when one changes (during development) - but at the cost of repeatedly re-importing chunks of the header tree over and over.

**trim down unnecessary includes and use more forward declarations**
I tried to move away from referencing any of the `Titanium/detail` headers as they are intended to be private (and once we do remove all external references to them we can edit the build to make them so).
I also tried to aggressively use forward declarations when possible. This does conflict a little with the notion that including the header of a class should bring everything you need to implement (if the API only references another class/struct via pointers I did remove the include from the app and move it to the cpp, but this can be annoying to then also re-add the include to the cpp of subclasses).

**Jenkins changes**
- tweaked some of the file transferring that happens to speed up builds.
- fixed gathering app crash/hang reports.
- turned off deleting the build directory when tests fail - so that I can manually log in to diagnose

⚠️ ☠️ **Please note that I am seeing the test suite app hang on windows store intermittently right now.** I would like to try and diagnose on the CI servers to see what's going on, but perhaps @infosia  can take a look in his dev environment with this PR? (I don't have a Windows SDK dev setup on my new laptop)


infosia:
This PR is reverted by #1362 because we had application performance x5 times slower based on the result from nightly build v8.1.0.v20190301155716. Created [TIMOB-26880](https://jira.appcelerator.org/browse/TIMOB-26880) to track this.

PR #1361, opened 2019-02-26
[TIMOB-26702] Make HAL more lightweight

infosia:
[TIMOB-26702](https://jira.appcelerator.org/browse/TIMOB-26702)

DO NOT MERGE THIS

This PR is created upon #1345 

This PR rewrites our `HAL` entirely from scratch to see how lightweight implementation of HAL affects performance. 

### Results

- No gain on application startup/rendering time (mostly no difference compared to #1345)
- Lightweight HAL uses 16MB (18%) less application memory
- Lightweight HAL uses 15% less CPU

**BEFORE**
before

**AFTER**
after

However, because this is quite big internal change in our basic framework, I'm not sure if we should apply these changes to our master branch. 

- We don't have no difference on our startup/rendering time.
- This may work with less CPU/memory, but I think most users don't care about it because latest PC machines often work with rich CPU/memory compared to mobile devices. Because our performance gain on #1345 is pretty significant (up to 45% improvement), I'm not sure lightweight HAL is really worth it now.


infosia:
Closing, #1365 will take over this.

PR #1362, opened 2019-03-04
Revert "Update cmake and usage, speed up SDK builds (#1346)"

infosia:
This reverts commit fe5e74062216d652dcd5d5e09a62294d071e1b89.

I'm observing this commits makes application performance x5 times slower based on the result from nightly build v8.1.0.v20190301155716. I'm trying to bring back previous state.


infosia:
@sgtcoolguy I merged this to master now because reverting your commits fixes the performance issue. I am under impression that this has something to do with linking with `HAL` because I saw the issue in #1345 but most interestingly I don't see the issue on my PR #1361 which basically re-wrote HAL framework entirely from scratch.

PR #1367, opened 2019-03-15
fix(require): require of directory with package.json pointing at dir

sgtcoolguy:
Fixes https://jira.appcelerator.org/browse/TIMOB-26788

equivalent of appcelerator/titanium_mobile#10672


sgtcoolguy:
merged manually

PR #1368, opened 2019-03-18
[TIMOB-26880] Modernize CMake usage

infosia:
DO NOT MERGE

Trying to see why updating CMake usage slows down the app.


infosia:
So I'm pretty much could't to see why this changes slow down the app. Here's some notes;

- Eliminating entire precompiled headers didn't help. So this has nothing to do with precompiled headers.
- Reverting to "old" CMake didn't help. So this has nothing todo with CMake version.
- This changes tend to trigger crash on Jenkins unit tests. I would guess that something is crashing in JavaScriptCore but I don't see how to fix it.

PR #1369, opened 2019-03-21
[TIMOB-26855] Add "baseURL" support to WebView.setHtml()

infosia:
[TIMOB-26855](https://jira.appcelerator.org/browse/TIMOB-26855)

Add "baseURL" support to WebView.setHtml().

Note that this implementation requires a little dirty "hack" because Windows UWP WebView doesn't have official API to set base URL when you use `Navigate*` methods. So what I did actually was inspired by [this article](https://codingisartblog.wordpress.com/2016/04/02/wp-hacks-implementing-loaddatawithbaseurl-on-windows-phone-8-1/) where you replace `html` content after WebView loads the base URL. It's dirty but this logic only triggered when you use `setHTML` with `baseURL` option so we could say it's nice to have and it's better than nothing.

```js
var htmlText =
    '<!DOCTYPE html>' +
    '<html>' +
    '	<body>' +
    '		<iframe width="560" height="315" src="https://www.youtube.com/embed/dQw4w9WgXcQ" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen/>' +
    '	</body>' +
    '</html>';

var window = Ti.UI.createWindow();
var webView = Ti.UI.createWebView();
//webView.html = htmlText;
webView.setHtml(htmlText, { baseURL: "https://www.youtube.com" });
window.add(webView);
window.open();
```

Expected: Tapping on the video image should plays the video fine.


jquick-axway:
@infosia, when setting the "html" property or calling the `setHtml()` method, paths are supposed to be relative to the app's "Resources" by default... unless you change it via the "baseURL" argument. Is this true on Windows?

For example, if you create a Titanium Classic "Default Project" app (it sets up a `TabGroup`) and run the below code, it should show the "tab1.png" within the `WebView`. The below works on both Android and iOS.
```javascript
var htmlText =
		'<!DOCTYPE html>' +
		'<html>' +
		'	<body>' +
		'		<p>Relative Path Test</p>' +
		'		<img src="assets/images/tab1.png"/>' +
		'	</body>' +
		'</html>';

var window = Ti.UI.createWindow();
var webView = Ti.UI.createWebView({
	html: htmlText,
});
window.add(webView);
window.open();
```

If it doesn't, then I suppose we can make a separate ticket for it. But note that "baseURL" can also be set to a "file://" URL to reference a local directory instead of a web location. (For "baseURL", I think this is a lower priority feature.)

infosia:
@jquick-axway As for the relative resources from resources directory, Windows WebView requires them to place into subfolder of `local` or `temp` folder by using `ms-appdata://` scheme such as`ms-appdata:///local/folder/file.html` and `ms-appdata:///temp/folder/file.html` (See also [Windows.UI.Xaml.Controls.WebViewAPI Reference](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.WebView)). So it's a parity issue due to a limitation of Windows WebView.

jquick-axway:
Does a UWP `WebView` support a "file://" scheme for the URL?
Can the HTML contents such as `` support a file:// URL as well?

I ask because I'm looking for something that'll work between all platforms, including Android and iOS.

infosia:
@jquick-axway I gave it a try and it seems `file://` doesn't work. The [API doc](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.WebView) does not state anything on `file://` but it says we can use `ms-appx-web://` scheme in that case.

ssjsamir:
@infosia can be merged once Jenkins is ready

PR #1376, opened 2019-04-04
Fix hang?

sgtcoolguy:
This moves the requiring of test files into a function. Then we wait until the window open callback and schedule setup of mocha, loading the test files and then running mocha - all on a setTimeout callback.

So maybe the initial window will load/open quickly and then fire off the suite async and avoid getting killed off as "hanging" by Windows Desktop?


infosia:
Awesome! I will merge this onto [titanium-mobile-mocha-suite](https://github.com/appcelerator/titanium-mobile-mocha-suite) later today.

PR #1394, opened 2019-05-22
[TIMOB-25121] View.top should be undefined by default

infosia:
DO NOT MERGE

[TIMOB-25121](https://jira.appcelerator.org/browse/TIMOB-25121)

Some View parameters should be `undefined` by default when values are not set. 

```js
var view1 = Ti.UI.createView({
		height: 50,
		center: {
			y: 200
		}
	});
win.addEventListener('postlayout', function () {
	try {
		Ti.API.info(view1.top);
		Ti.API.info(view1.left);
		Ti.API.info(view1.right);
		Ti.API.info(view1.bottom);
	} catch (e) {
		Ti.API.info(e);
	}
});
win.add(view1);
win.open();
```

Expected:
```
[INFO] undefined
[INFO] undefined
[INFO] undefined
[INFO] undefined
```

### DIFF
--- a/tests/Resources/ti.ui.layout.test.js
+++ b/tests/Resources/ti.ui.layout.test.js
@@ -1,0 +1,1967 @@
+/*
+ * Appcelerator Titanium Mobile
+ * Copyright (c) 2011-Present by Appcelerator, Inc. All Rights Reserved.
+ * Licensed under the terms of the Apache Public License
+ * Please see the LICENSE included with this distribution for details.
+ */
+/* eslint-env mocha */
+/* global Ti */
+/* eslint no-unused-expressions: "off" */
+'use strict';
+var should = require('./utilities/assertions'),
+	utilities = require('./utilities/utilities');
+
+function createWindow(_args) {
+	_args = _args || {};
+	_args.backgroundColor = _args.backgroundColor || 'red';
+	return Ti.UI.createWindow(_args);
+}
+
+describe('Titanium.UI.Layout', function () {
+	var win,
+		didPostlayout = false;
+	this.timeout(5000);
+
+	beforeEach(function () {
+		didPostlayout = false;
+	});
+
+	afterEach(function (done) {
+		if (win) {
+			win.addEventListener('close', function () {
+				done();
+			});
+			win.close();
+		} else {
+			done();
+		}
+		win = null;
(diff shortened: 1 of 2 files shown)


Source/LayoutEngine/src/Composite.cpp
```diff
@@ -278,7 +279,8 @@ namespace Titanium
 					x2 -= topValue;
 				}
 				if (centerYType == Fixed) {
-					x2 = 2 * (bottomValue - centerYValue);
+					x1 = 2.0;
+					x2 = -2 * (bottomValue + centerYValue);
```
infosia: The reason we do this is because:

```
View height = 2x(window_height - bottom - center)
```

and `doCompositeLayout` does calculate the view's height based on this

```
measuredHeight = heightLayoutCoefficients.x1 * height + heightLayoutCoefficients.x2;
```

Source/LayoutEngine/src/ParseProperty.cpp
```diff
@@ -24,7 +24,7 @@ namespace Titanium
 				return Size;
 			} else if (value == "UI.FILL") {
 				return Fill;
-			} else if (value == "NONE" || value == "undefined") {
+			} else if (value.empty() || value == "NONE" || value == "undefined") {
 				return None;
```
infosia: This fix is needed to treat empty string as "None" value type. Empty string has been treated as "Fixed" type with value "0" but it was wrong.

tests/Resources/ti.ui.layout.test.js
```diff
@@ -0,0 +1,1967 @@
+/*
```
sgtcoolguy: again, if possible, making an "addontest" file with the additional tests would be much preferable, since overriding the existing suite makes it impossible to tell what the actual difference is in this PR.

But, I'm assuming that this fix actually requires modifying existing tests to re-enable them on Windows now that they're fixed?

Can you point out what exactly was changed here vs the existing suite test?

infosia: Right, this PR is a fix for `undefinedTop` test failure for `ti.ui.layout`. ([TIMOB-25121](https://jira.appcelerator.org/browse/TIMOB-25121))

PR #1430, opened 2019-10-09
fix: only copy windows ti.main.js file

ewanharris:
https://jira.appcelerator.org/browse/TIMOB-27452
Fixes TIMOB-27452

Requires appcelerator/titanium_mobile/pull/11269


infosia:
I'll re-run https://github.com/appcelerator/titanium_mobile/pull/11269 tests and see how it goes.

PR #1434, opened 2019-11-13
refactor(cli/build): move to using common processjstask

ewanharris:
This aligns Windows with the same js processing code as Android and iOS, which handles all the knowledge like dont minify the internal common files, incremental builds etc

Fixes [TIMOB-27452](https://jira.appcelerator.org/browse/TIMOB-27452)

Pinging @janvennemann to just make sure I'm using the processjs stuff right

### DIFF
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -589,9 +589,11 @@ function copyResources(next) {
 			.then(() => {
 
 				const tasks = copyUnmodified.map(relPath => {
-					const from = jsFiles[relPath];
-					const to = path.join(this.buildTargetAssetsDir, relPath);
-					copyFile.call(this, from, to, next);
+					return next => {
+						const from = jsFiles[relPath];
+						const to = path.join(this.buildTargetAssetsDir, relPath);
+						copyFile.call(this, from, to, next);
+					}
 				});
 
 				appc.async.parallel(this, tasks, () => {
--- a/cli/commands/_build/copy.js
+++ b/cli/commands/_build/copy.js
@@ -9,6 +9,7 @@ const jsanalyze = require('node-titanium-sdk/lib/jsanalyze');
 const path = require('path');
 const ti = require('node-titanium-sdk');
 const __ = appc.i18n(__dirname).__;
+const ProcessJsTask = require('../../../../cli/lib/tasks/process-js-task');
 
 /*
  Public API.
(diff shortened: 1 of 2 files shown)


cli/commands/_build/copy.js
```diff
@@ -543,133 +544,117 @@ function copyResources(next) {
...
-				// ignore appxmanifest
-				if (prop !== 'manifest') {
-					props[prop] = this.tiapp.windows[prop];
-				}
-			}, this);
-			fs.writeFileSync(
-				appPropsFile,
-				JSON.stringify(props)
-			);
-			this.encryptJS && jsFilesToEncrypt.push('_app_props_.json');
```
sgtcoolguy: My understanding is that each platform is still doing the app props/info files separately and it is not actually part of the process-js-task code right now.

cli/commands/_build/copy.js
```diff
@@ -543,133 +544,117 @@ function copyResources(next) {
...
 			}
-
-			this.processEncryption(next);
 		});
+
+		task.run()
+			.then(() => {
+
+				const tasks = copyUnmodified.map(relPath => {
+					const from = jsFiles[relPath];
```
ewanharris: This code makes no sense I think, will fix

sgtcoolguy:
I mistakenly thought the app_props and info json file stuff was removed, but it was just re-arranged. But that does raise the obvious question to @janvennemann (and the rest of us) - can we move more of this common code into the process-js-task (or some common build code location) - because it looks like each platform is doing it's own _app_props_.json stuff.

The _app_info_.json seems specific to Windows.

cli/commands/_build/copy.js
ewanharris: Confirmed non-sensical code, fix incoming

ewanharris: Fix in 1cb9de77fe5088d9e8f5ea4a4853af4825980780

cli/commands/_build/copy.js
```diff
@@ -543,133 +544,119 @@ function copyResources(next) {
...
-				// if we're encrypting the JavaScript, copy the files to the assets dir
-				// for processing later
-				if (this.encryptJS) {
-					to = path.join(this.buildTargetAssetsDir, id);
-					jsFilesToEncrypt.push(id);
-				}
-
-				try {
-					this.cli.createHook('build.windows.analyzeJsFile', this, function (from, to, cb) {
-						this.cli.createHook('build.windows.copyResource', this, function (from, to, cb) {
```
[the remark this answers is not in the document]
ewanharris: I think we'll have to update the ProcessJsTask to do that, will put a PR on titanium_mobile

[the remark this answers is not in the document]
infosia: I think that makes sense. I'll merge this PR.

infosia:
Merged, still needs to update the ProcessJsTask to do that, @ewanharris will put a PR on titanium_mobile

PR #1435, opened 2019-11-13
refactor(cli/build): move to using common processjstask

ewanharris:
This aligns Windows with the same js processing code as Android and iOS, which handles all the knowledge like dont minify the internal common files, incremental builds etc

Fixes [TIMOB-27452](https://jira.appcelerator.org/browse/TIMOB-27452)


infosia:
Merged, still needs to update the ProcessJsTask to do that, @ewanharris will put a PR on titanium_mobile

